PR #41, opened 2021-09-07
feat: whitelist packages via custom rollup plugin

ndom91:

## ⚙️ Affected Components
* [ ] Commands
* [ ] Modules
* [x] Services
* [ ] SDK

### 📝 Notes for the Reviewer

Basic rollup plugin which fetches our runtime (from a hardcoded endpoint atm) and does a  basic comparison of the package names against our allowed list of package names. If the user is trying to import a package we don't have on the whitelist, it'll bail on the bundling and print "Package XYZ not allowed". Currently it is using the dependencies of the newest runtime as thats what we use to run the cli checks by default, but that should probably be dynamic as well.

You can test it with a browser check that defines a `path: '../browse-check-script.js'` instead of the `script: | ..` inline script, for example. Use the standard template browser check script to start. Then try to import a package we don't support, like `react`, and you will see the following output:

image

**EDIT**: Had to temporarily switch the bundler output rollup settings from `cjs` to `es` because with cjs it didnt support top-level await so our template browser check script wouldn't bundle. Obviously we can change this back if thats the more appropriate option, idk whats best to send to our backend yet.


ndom91:
@ianaya89 what do you think about this? Does this fit to your preexisting rollup setup / goals?

ianaya89:
@ndom91 it's a smart approach, I think we can move forward with this.

ianaya89 (approved):

ndom91:
@ianaya89 I made some minor changes:

1. used the prod runtimes URL (its unauthenticated so we shouldn't have to switch to local during dev, etc. can probably just leave it like this for local dev + prod)
2. made the runtime dynamic. Defaults to the latest one (`2021.06`) but if a check defines a custom runtime in their yml like:
```yml
runtime: "2020.01"
```
it will use that instead to check dependencies.

ianaya89 (approved):
:shipit:

PR #61, opened 2021-09-23
refactor: remove conf command and re organize configuration properties

ianaya89:
I hereby confirm that I followed the code guidelines found at [engineering guidelines](https://www.notion.so/checkly/Engineering-Guidelines-a3a165a203a04dc1a112f0e26b2f2d3f)

## ⚙️ Affected Components
* [ ] Commands
* [ ] Modules
* [x] Services
* [ ] SDK

### 📝 Notes for the Reviewer
- Split config settings in two files: `auth.json` and `config.json`.
- Add warning messages to both files.
- Handle environment (only) with `NODE_ENV` .
- Remove `conf` command.
- Add helpers to access `apiKey` and `env`. We should always use them.
- Improve config service structure and code.

### 🏗️ New Dependency Submission
```diff
- conf-cli
```

> Resolves #50

PR #68, opened 2021-09-28
feat: add account switch command

### LINKED ISSUE #48: Account Switching
# Rationale

A single, authenticated user can have multiple accounts on Checkly. We want to give users the option to explicitly switch between accounts from the CLI.

This switching makes sense for many interactive, read-only commands like `checkly status` where a user might want to quickly see what's up in their accounts.

The switching might cause friction when 
1. A user is on an initialized project that is deployed.
2. Then switches to a different account.
3. Then runs deploy again. This would create a completely new project and set of checks on the API/cloud side.

At this moment, this is acceptable and actually expected behaviour.

# Acceptance criteria

We do not have to reinvent the wheel here. The functionality should be similar to for instance the `vercel teams` command.

- [x] `checkly accounts` should list the accounts you have access to. This is an alias for `checkly accounts ls`
- [x] `checkly accounts switch`should start an interactive prompt that allows the user to pick an account.
- [ ] If a user has only one account, the user cannot switch.

Example for account switcher from the `vercel` cli.

```
vercel teams switch
> Switch to: (Use arrow keys)
  ── Personal Account ──────────
❯ Tim Nolet (timnolet) (current) 
  ── Teams ─────────────────────
  Checkly (checkly) 
  ──────────────────────────────
  Abort 
```

ianaya89:

## ⚙️ Affected Components
* [x] Commands
* [ ] Modules
* [ ] Services
* [ ] SDK

### 📝 Notes for the Reviewer

> ⚠️ Depends on #61 

> 🚨 This command does not make sense until we move to user-based API Keys.

Add new account switch command.
- Interactive mode: pick your new account from the list
- Manual mode: set your new account using a flag (`-a --accountId`)

image

> Resolves #48


src/services/api.js
```diff
@@ -11,27 +11,45 @@ const spinner = ora({
   color: 'cyan',
 })
 
-function getApiDefaults() {
+function getDefatuls() {
+  const environmments = {
```
ndom91, line 15: Nit: Small typo in environment and getDefaults, as well as in their usage below then obviously

ndom91 (dismissed):
Small typo but otherwise LGTM 🎉

PR #72, opened 2021-09-29
feat: add use case for single account users (switch)

ianaya89:

## ⚙️ Affected Components
* [x] Commands
* [ ] Modules
* [ ] Services
* [ ] SDK

### 📝 Notes for the Reviewer
Consider the scenario where a user with a single account tries to switch

### DIFF
--- a/src/commands/switch.js
+++ b/src/commands/switch.js
@@ -28,6 +28,14 @@ class SwitchCommand extends Command {
 
     const { data: accounts } = await api.accounts.find()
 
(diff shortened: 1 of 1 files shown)


ndom91 (approved):
LGTM!

Nit: maybe word it something like `'`Your user is only a member of one account`? Since they dont really _have_ accounts they're invited to them / are members of them?

ndom91 (approved):
LGTM 👍

PR #87, opened 2021-11-04
User key login

### LINKED ISSUE #47: CI Login
# Rationale

A user should be able to authenticate the CLI in a non-interactive mode. This is specifically useful for throw-away environments like in a typical CI provider, or when used in bash scripts. In all of these cases, an interactive web-based login is not an option.

Instead, the Checkly CLI should have a `--token` flag that overrides any current, local auth tokens. The user gets this token by creating it explicitly in the Checkly web UI.

# Acceptance criteria

- [x] The CLI has a `--token` flag at the global level. This flag has a per-invocation scope. It is not stored anywhere locally.
- [x] The Checkly web UI has an option to create and delete this token.
- [x] The CLI, when invoked with `--token` overrides any locally available auth methods.

Example:

```sh
# checkly status --token ABCD123XYZ
```

ianaya89:

## ⚙️ Affected Components

- [x] Commands
- [x] Modules
- [x] Services
- [x] SDK
- [x] Tests

### 📝 Notes for the Reviewer

Adapt login to user keys auth and allow checkly account selection/switch

https://user-images.githubusercontent.com/3258966/140398051-727753d4-99ba-408d-98f7-42c7ff22833b.mp4

> Resolves #47 

```diff
+ open
```


src/services/login-util.js
```diff
@@ -17,6 +18,7 @@ const generateAuthenticationUrl = (codeChallenge, scope, state) => {
     redirect_uri: 'http://localhost:4242',
     scope: scope,
     state: state,
+    audience: ['api-dev.checkly.run'],
```
umutuzgur, line 21: This is different for dev/test and prod

ianaya89: You are right, will move this to config file.

PR #114, opened 2021-12-09
feat: github actions smoke-test

ndom91:

## ⚙️ Affected Components

- [ ] Commands
- [ ] Modules
- [ ] Services
- [ ] SDK
- [x] Tests

### 📝 Notes for the Reviewer

Add GH Action smoke-test which executes the following:
1. project init
2. deploying project (create project, create check)
3. update check
4. delete check
5. delete project


src/commands/init.js
```diff
@@ -130,6 +140,12 @@ InitCommand.description = 'Initialise a new Checkly Project'
 
 InitCommand.flags = {
   force,
+  skipSetup: flags.boolean({
```
ianaya89, line 143: The `force` flag and `skipSetup` flag have the same purpose. I would use `force` since is more standard in UNIX commands

src/services/config.js
```diff
@@ -49,6 +52,20 @@ const config = {
...
+    try {
+      const parsedSettings = YAML.parse(
+        fs.readFileSync(
+          path.join(process.cwd(), '.checkly/settings.yml'),
+          'utf8'
+        )
+      )
+      return parsedSettings.projectId
+    } catch (e) {
+      return undefined
```
ianaya89, line 65: I think `null` is more accurate

src/parser/file-parser.js
```diff
@@ -76,7 +76,8 @@ function parseChecklyDirectory() {
     )
     return parseChecklySettings(parsedFiles)
   } catch (err) {
-    throw new Error(err.message)
+    console.error(err)
```
ianaya89, line 79: Can we use `consola.error` instead?

ianaya89 (changes requested):
Looks good but I have a few comments

src/commands/init.js
ndom91: Ahh didnt see there was a force. But it doesn't seem to skip the init inquirer.js prompts, does it?

src/services/config.js
ndom91: Does it matter in this case? The only place its being used atm is in the new `projects.delete()` services/api method being added in this PR as well, i.e. at `src/modules/projects/delete.js:10`

src/parser/file-parser.js
ndom91: Yeah I was thinking about that actually, but consola wasn't imported here yet and I didnt want to do it necessarily just for this. But you're right - its much nicer and probably worth it :+1:

src/commands/init.js
ianaya89: It was there but was not implemented yet. The behaviour you add is fine but we can just replace `skipSetup` with `force`

ndom91: Okay great, ill adjust it :+1:

src/services/config.js
ianaya89: It's more a philosophical concept where I believe that `null` is more intended to represent the intentional absence. (IMO)

umutuzgur: Intentional absence is my new favorite phrase :D

.github/workflows/test.yml
```diff
@@ -0,0 +1,49 @@
...
+      GITHUB_ACCESS_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v2
+        with:
+          node-version: '14'
+          cache: 'npm'
+      - run: npm install
+      - name: Login
+        run: ./bin/run login --api-key ${{ secrets.USER_API_KEY_TEST }} --account-id ${{ secrets.USER_ACCOUNT_TEST }}
```
umutuzgur, line 22: This is a nit pick but is there way we can link ./bin/run to checkly-cli so it looks prettier :D

umutuzgur: Something like example below but we need to rename run or give it an alias

echo "::add-path::$RUNNER_WORKSPACE/$(basename $GITHUB_REPOSITORY)/bin" # Make it accessible from runner

src/services/config.js
ndom91: Yeah I was thinking about this a bit actually, I wanted the projectId to be there, but not _be_ anything, so I went with undefined. But it really doesn't matter to me, I'll change it to `null` :+1:

.github/workflows/test.yml
ndom91: We could just add a symlink to `/bin/run` as `/bin/checkly-cli` to the bin directory. That way we can use it here with your `add-path` workaround and in dev as `checkly-cli` more easily as well. What do you think?

umutuzgur (approved):
LGTM!

PR #118, opened 2021-12-11
fix(init): create project on BE before writing .checkly dir [gh-00]

ndom91:

## ⚙️ Affected Components

- [x] Commands
- [ ] Modules
- [ ] Services
- [ ] SDK
- [ ] Tests

### 📝 Notes for the Reviewer

This PR changes the order of some actions in the `init` command. Moving the creation of the project, and communication with the backend, to earlier in the process. 

This is so that in case of network or any other BE errors, the process exits earlier, and the users are not left with a dangling invalid `.checkly` directory.

The current order of operations is such that - first the `.checkly` directory gets created, as well as an example check. After that, we talk with the backend and create the project in the DB, get back an ID for it, etc. and write that to the `.checkly/settings.yml` file. If it were to fail at that point, we do not get a valid `.checkly/settings.yml` file which contains our `projectId`, and are simply left with an invalid `.checkly` directory. 

This way, if there is an error, the user still has a clean working directory, no invalid .checkly dir, and can cleanly just try again. I've also added a try/catch to be able to graceful tell the user whats wrong.

### DIFF
--- a/src/commands/init.js
+++ b/src/commands/init.js
@@ -70,6 +70,15 @@ class InitCommand extends Command {
       return process.exit(1)
     }
 
+    const { data: project } = await projects.create({
+      accountId: config.getAccountId(),
+      name: args.projectName,
+      repoUrl: await getRepoUrl(cwd),
+      activated: true,
+      muted: false,
+      state: {},
+    })
+
     if (!force) {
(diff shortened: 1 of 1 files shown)


umutuzgur (approved):
LGTM

ianaya89 (approved):

PR #119, opened 2021-12-11
feat: test CRUD group actions [gh-00]

ndom91:

## ⚙️ Affected Components

- [ ] Commands
- [ ] Modules
- [ ] Services
- [ ] SDK
- [x] Tests

### 📝 Notes for the Reviewer

Add actions for groups, in addition to individual checks, to smoke-test GH Action.

1. Create Group + Check
2. Update Group
3. Delete Group

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -37,10 +37,8 @@ jobs:
       - name: Create test group files
         run: |
           mkdir -p .checkly/checks/group-1
-          pwd
-          ls -lah
           cp test/fixtures/yml/group-valid.yml .checkly/checks/group-1/settings.yml
-          cp test/fixtures/yml/browser-check-valid.yml .checkly/checks/group-1/browser-check.yml
+          cp test/fixtures/yml/api-check-valid.yml .checkly/checks/group-1/api-check.yml
       - name: Create group
         run: checkly-cli deploy
       - name: Change group name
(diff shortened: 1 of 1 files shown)


ianaya89 (approved):

ndom91:
So it is currently failing on the last deploy step, its receiving a 500 response where it should be syncing the deletion of the `group-1` and `group-1/api-check.yml` resources.

I've run through these exact steps as I've defined them in the GH Action locally as well, and it seems to have worked there. 

Heres the log output from here, looks like its having trouble deleting the check inside the group possibly?

image

I've rerun it a few times, and it doesn't seem like a fluke, it always fails at this step.

ianaya89 (approved):

ndom91:
@ianaya89 I dont think I wanna merge it in this state, what do you think? Smoke-test gh action would be broken. 

I could disable this one failing new step - `DELETE` group - so that the smoke-test passes again. This PR will then at least add the testing of `CREATE` and `UPDATE` check-groups to the smoke-test.

Or maybe you or @umutuzgur have some time to look a bit deeper into why the `DELETE` groups step is failing on prod. As mentioned, it worked multiple times locally. I can't reproduce this failure myself :/. The 500 error from the BE seems to be pointing toward an issue with the SQL delete check statement, see papertrail logs linked above ^^

PR #128, opened 2021-12-17
feat: add testing for assertions

### LINKED ISSUE #59: Support for Assertions

danielpaulus:

## ⚙️ Affected Components

- [x] Commands
- [ ] Modules
- [x] Services
- [ ] SDK
- [x] Tests

### 📝 Notes for the Reviewer
I moved out deploy code into a separate file so it becomes testable without having to use oclif 
`function getDefaults() {` I think we should remove, would do that in a small follow up PR, we can and should use
env vars or config for these things

dotenv - so we can use env vars instead of hardcoding stuff into the code
mock-require - so we can mock npm modules in mocha tests
memfs - so we can use an in memory filesystem for tests, that is a ton easier to handle and makes the test faster to run

### 🎟 Affected Issue
Closes #59


src/services/config.js
```diff
@@ -20,11 +20,17 @@ const authSchema = {
...
+    configName: 'auth',
+    projectSuffix,
+    schema: authSchema,
+    projectName: 'checkly-cli',
+  }),
   data: new Conf({
     configName: 'config',
     projectSuffix,
     schema: dataSchema,
+    projectName: 'checkly-cli',
```
ianaya89, line 33: Why do we need this?

src/services/api.js
```diff
@@ -77,6 +77,6 @@ function init() {
 }
 
 module.exports = {
-  getDefatuls,
+  getDefatuls: getDefaults,
```
umutuzgur, line 80: dont forget to fix this

ianaya89 (approved):
I think looks good. I would rename `deployservice` to `deploy` (because is already in the `services` directory)

Regarding `getDefaults`, the idea was to avoid having a lot of configuration variables. Actually, those things used to live in the config file, so we didn't need ENV variables for everything. We decided to remove them to make it easier and avoid over-engineering, that is the reason those values now depend on the value of `NODE_ENV`

danielpaulus:
> 

Thanks! Good point changing the name of deployservice! I need to still get used to NodeJS/JS good practices so I appreciate the help :-D

We can leave the defaults in place for now. I wanted to use dotenv only to enable a simple CI setup. I need to add some Checkly API key for the CI to run its tests against. The idea is that anyone can just run the tests against a checkly account of their choice.

src/services/config.js
danielpaulus: removed it again

ianaya89:
@danielpaulus sounds good!

Something else regarding the `deploy` service. For most of the commands, I splited them into `modules` (checks, groups, etc), so we can have all the "business logic" outside the `oclif` command and facilitate testing. Maybe `deploy` should be a module as well (instead of a service).

I think `services` as a kind of helper to share code across modules/commands and `modules` as a sort of controller.

umutuzgur (dismissed):
LGTM

danielpaulus:
> 

ok

umutuzgur (approved):

PR #151, opened 2021-12-23
feat: allow users to pass logical id to run checks

umutuzgur:

## ⚙️ Affected Components

- [ ] Commands
- [ ] Modules
- [ ] Services
- [ ] SDK
- [ ] Tests

### 📝 Notes for the Reviewer

This is an idea on how `checkly-cli run` would work with logical ids we have in the repo

`checkly-cli run example-api.yml`. This is not the path but the logical ids what the parser returns. We can also change the command to something like `checkly-cli run .checkly/checks/example-api.yml`


ianaya89 (approved):
Nice 💯

PR #383, opened 2023-01-09
chore: report cli version in http headers

clample:

### Notes for the reviewer
Knowing which agents are in-use could be helpful for making product decisions. This PR adds a `x-checkly-cli-version` header which will be incldued on requests to the Checkly backend.

### DIFF
--- a/package/src/hooks/configure-api-client.ts
+++ b/package/src/hooks/configure-api-client.ts
@@ -1,0 +1,14 @@
+import { Hook } from '@oclif/core'
+import { api } from '../rest/api'
+
+// eslint-disable-next-line require-await
+const hook: Hook.Prerun = async function ({ config: oclifConfig }) {
+  api.interceptors.request.use((config) => {
+    if (config.headers) {
+      config.headers['x-checkly-cli-version'] = oclifConfig.version
+    }
+    return config
+  })
+}
+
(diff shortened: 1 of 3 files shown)


package/src/hooks/configure-api-client.ts
```diff
@@ -0,0 +1,14 @@
+import { Hook } from '@oclif/core'
+import { api } from '../rest/api'
+
+// eslint-disable-next-line require-await
+const hook: Hook.Prerun = async function ({ config: oclifConfig }) {
+  api.interceptors.request.use((config) => {
+    if (config.headers) {
+      config.headers['x-checkly-cli-version'] = oclifConfig.version
+    }
```
umutuzgur, line 9: Shouldn't we create headers if it is missing?

clample: Good point. I was kind of just following what we have in the rest API https://github.com/checkly/checkly-cli/blob/5f4f2dd5c3435bc5fee3400ce9fd5e734ab65012/package/src/rest/api.ts#L43-L57

I just switched this to use ["Axios custom instance defaults"](https://github.com/axios/axios#custom-instance-defaults) instead.

umutuzgur (approved):

PR #393, opened 2023-01-12
feat: add relative path support to check and check groups [sc-0]

umutuzgur:

### Notes for the reviewer
Add relative path support to checks and check group pattern

### DIFF
--- a/package/src/constructs/browser-check.ts
+++ b/package/src/constructs/browser-check.ts
@@ -42,7 +42,7 @@ export class BrowserCheck extends Check {
         if (!this.__checkFilePath) {
           throw new Error('You cant use relative paths without the __checkFilePath')
         }
-        absoluteEntrypoint = path.join(this.__checkFilePath, entrypoint)
+        absoluteEntrypoint = path.join(path.dirname(this.__checkFilePath), entrypoint)
       }
       // runtimeId will always be set by check or browser check defaults so it is safe to use ! operator
       const bundle = BrowserCheck.bundle(absoluteEntrypoint, this.runtimeId!)
--- a/package/src/constructs/browser-check.ts
+++ b/package/src/constructs/browser-check.ts
@@ -35,8 +35,17 @@ export class BrowserCheck extends Check {
       this.script = script
     } else if ('entrypoint' in props.code) {
       const entrypoint = props.code.entrypoint
+      let absoluteEntrypoint = null
+      if (path.isAbsolute(entrypoint)) {
+        absoluteEntrypoint = entrypoint
+      } else {
+        if (!this.__checkFilePath) {
+          throw new Error('You cant use relative paths without the __checkFilePath')
+        }
+        absoluteEntrypoint = path.join(this.__checkFilePath, entrypoint)
+      }
       // runtimeId will always be set by check or browser check defaults so it is safe to use ! operator
-      const bundle = BrowserCheck.bundle(entrypoint, this.runtimeId!)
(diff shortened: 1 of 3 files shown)


package/src/constructs/browser-check.ts
```diff
@@ -35,8 +35,11 @@ export class BrowserCheck extends Check {
       this.script = script
     } else if ('entrypoint' in props.code) {
       const entrypoint = props.code.entrypoint
+      const absoluteEntrypoint = path.isAbsolute(entrypoint)
+        ? entrypoint
+        : path.join(this.__checkFilePath, entrypoint)
```
clample, line 40: `path.resolve` should achieve the same behavior ([docs](https://nodejs.org/api/path.html#pathrelativefrom-to))

```suggestion
      const absoluteEntrypoint = path.resolve(this.__checkFilePath, entrypoint)
```

package/src/constructs/check-group.ts
```diff
@@ -74,7 +77,7 @@ export class CheckGroup extends Construct {
   }
 
   addChecks (pattern: string) {
-    const matched = glob.sync(pattern, { nodir: true })
+    const matched = glob.sync(pattern, { nodir: true, cwd: this.__checkFilePath })
```
clample, line 80: Nice, this seems like better behavior

package/src/constructs/check.ts
```diff
@@ -59,7 +59,7 @@ export abstract class Check extends Construct {
     this.groupId = props.groupId
     // alertSettings, useGlobalAlertSettings, groupId, groupOrder
 
-    this.__checkFilePath = Session.checkFilePath
+    this.__checkFilePath = Session.checkFilePath!
```
clample, line 62: I think that sometimes the `Session.checkFilePath` will be undefined. For checks that come from `*.check.js` files, we should always have `Session.checkFilePath` but for browser checks that are created automatically from `*.spec.js` files, it isn't set.

https://github.com/checkly/checkly-cli/blob/3b9a21bffda3999c8bdaec1c2c04c17b8a0d2ad9/package/src/services/project-parser.ts#L60-L108

umutuzgur: Hmm. We gotta pass that each time then. I will take a look at it

clample: Sounds good 👍

tnolet:
@clample please also update the example in README.md

umutuzgur:
@tnolet Gotcha and this my PR 😛

clample (approved):

package/src/constructs/browser-check.ts
```diff
@@ -35,8 +35,17 @@ export class BrowserCheck extends Check {
...
     } else if ('entrypoint' in props.code) {
       const entrypoint = props.code.entrypoint
+      let absoluteEntrypoint = null
+      if (path.isAbsolute(entrypoint)) {
+        absoluteEntrypoint = entrypoint
+      } else {
+        if (!this.__checkFilePath) {
+          throw new Error('You cant use relative paths without the __checkFilePath')
+        }
+        absoluteEntrypoint = path.join(this.__checkFilePath, entrypoint)
```
clample, line 45: I think that the test failure might be coming from here? Since __checkFilePath is a file, it might make sense to do:

`path.join(path.dirname(this.__checkFilePath), entrypoing)`

package/src/constructs/check-group.ts
clample, line 80: `__checkFIlePath` is a path rather than a directory. Maybe this should be `path.dirname(this.__checkFilePath)`.

clample (approved):

Issue #313, opened 2022-12-23
Run E2E tests in PR check


clample:
An E2E test harness is introduced in https://github.com/checkly/checkly-cli/pull/312. We should run the E2E tests with each PR. This will require setting up a test account in Checkly, adding config variables/secrets for GH Actions, and adding `npm run test:e2e` to the GH Action.

PR #394, opened 2023-01-12
chore: add e2e tests in PR check

clample:

### Notes for the reviewer
Closes https://github.com/checkly/checkly-cli/issues/313

* Add the e2e test to the PR check GitHub Action
* Make sure that resources are cleaned up for `deploy.spec.ts`
* Fix the project structure for `check-parse-error.spec.ts`

### DIFF
--- a/package/e2e/__tests__/deploy.spec.ts
+++ b/package/e2e/__tests__/deploy.spec.ts
@@ -1,10 +1,12 @@
 import * as path from 'path'
 import * as config from 'config'
-import { runChecklyCli } from '../run-checkly'
+import { v4 as uuidv4 } from 'uuid'
 import axios from 'axios'
+import { runChecklyCli } from '../run-checkly'
 import Projects from '../../src/rest/projects'
+import { DateTime, Duration } from 'luxon'
 
-async function deleteProject(projectLogicalId: string) {
+async function cleanupProjects(projectLogicalId?: string) {
   const baseURL: string = config.get('baseURL')
   const accountId: string = config.get('accountId')
   const apiKey: string = config.get('apiKey')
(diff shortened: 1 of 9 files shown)


package/e2e/__tests__/deploy.spec.ts
```diff
@@ -1,8 +1,36 @@
...
+  const { data: projects } = await projectsApi.getAll()
+  const project = projects.find(({ logicalId }) => logicalId === projectLogicalId)
+  if (project) {
+    await projectsApi.deleteProject(project.id)
+  }
+}
 
 describe('deploy', () => {
+  // Attempt to delete the project in case it wasn't removed during the previous run
+  beforeAll(() => deleteProject('e2e-test-deploy-project'))
```
umutuzgur, line 30: So we won't be able to run this in parallel

clample: Running the e2e tests in parallel could cause some errors.

I can limit the concurrency in the GH Action https://docs.github.com/en/actions/using-jobs/using-jobs-in-a-workflow. I think that this would be a nice straightforward solution. I would just limit the concurrency for e2e tests rather than for the whole PR check.

We could also support concurrent runs by using UUIDs prefixes in the project name.

I'll add the UUID prefixes.

umutuzgur: > I'll add the UUID prefixes.

I think that would be the best solution. We can then also have smarter cleanup script with prefix matching and created_at < 10m ago

clample: Sounds good, I added that to the PR

umutuzgur (approved):
LGTM

PR #403, opened 2023-01-16
fix: use Checkly auth0 domain

clample:

### Notes for the reviewer
image

Currently logging in with `npx checkly login` and using GitHub OAuth fails. After logging in with GitHub, the webpage redirects to the error page shown above. The reason is that the initial `/authorize` request to Auth0 goes to `checkly.eu.auth0.com`, but the `/callback` call to Auth0 goes to `auth.checklyhq.com`. This setup isn't supported by Auth0 - both requests are required to same domain ([relevant forum post](https://community.auth0.com/t/you-may-have-pressed-the-back-button-but-i-really-didnt/23877/16)). 

To fix the issue, we can just use the `auth.checklyhq.com` domain. This is the one that's already used when you do a normal login from the Checkly webapp.

I've tested with GitHub OAuth, Email+Password, and Google OAuth.

### DIFF
--- a/package/src/auth/index.ts
+++ b/package/src/auth/index.ts
@@ -3,11 +3,10 @@ import * as os from 'os'
 import * as http from 'http'
 import * as crypto from 'crypto'
 
-const AUTH0_DOMAIN = 'checkly'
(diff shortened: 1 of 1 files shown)


umutuzgur (approved):
LGTM

PR #413, opened 2023-01-18
docs: adds new docs + updates examples

tnolet:

### Notes for the reviewer

as per the title.


README.md
```diff
@@ -11,10 +11,7 @@ The Checkly CLI and Constructs in this repo together form the basic building blo
 This goal of this repo and the larger MaC project is to deliver a Javascript/Typescript-native workflow for creating,
 debugging, deploying and life cycling synthetic monitors (checks) at scale, from your code base.
 
-This project is now in `alpha`:
-1. Our goal is to first make a great DX / workflow for writing, debugging and deploying checks: focus on terminal, CLI and programming model.
-2. Not all TS types are done yet and also UI integration is largely not there.
-3. New check types, resources (like alerting channels) will follow soon.
+This project is now in `alpha`, we aim for `beta` mid-February '23. `GA` should be before end Q1 '23.
```
clample, line 14: Should we communicate a bit what users should expect from "alpha", "beta", and "GA"? Maybe something like "During the alpha phase there may be some bugs, missing features, and occasionally a breaking change."

README.md
```diff
@@ -430,15 +389,113 @@ When **running the CLI from your CI pipeline** you will need to export two varia
 Go to your Settings page in Checkly and grab a fresh API key from [the API keys tab](https://app.checklyhq.com/settings/user/api-keys) and your
 Account ID from the [Account settings tab](https://app.checklyhq.com/settings/account/general).
 
-# Constructs API
+# Using the Constructs API
+
+All resources you can create and manage using the Checkly CLI are derived from "constructs". These constructs are just 
+[TypeScript classes](https://github.com/checkly/checkly-cli/tree/main/package/src/constructs) like `ApiCheck` in `api-check.ts` and
```
clample, line 395: I'm not sure if users will actually want to look at the code to get started, and I'm not sure if it's worth mentioning the file names.

What about something like:

> To manage resources in Checkly from the Checkly CLI you can use "constructs". These constructs are JS/TS classes that represent resources like checks, alert channels, and check groups in Checkly.  You can use standard JS/TS programming when creating constructs: loops, variables, if-statements, file imports, extensions etc. When running `npx checkly deploy`, your monitoring setup in Checkly will be updated to match the constructs defined in your code.

clample: Assuming that we keep the README up to date, I think that most users won't need to look at the code.

README.md
```diff
@@ -430,15 +389,113 @@ When **running the CLI from your CI pipeline** you will need to export two varia
...
+- `runtimeId`: The ID of which [runtime](https://www.checklyhq.com/docs/runtimes/specs/) to use for this check.
+
+Note that most properties have sane default values and do not need to be specified.
+
+### API Checks
+
+API checks are a good fit for monitoring typical HTTP based endpoints like REST APIs and GraphQL APIs, but can also be
+used for form encoded payloads. [Read more about API checks in our docs](https://www.checklyhq.com/docs/api-checks/)
+
+- It defines the basic check properties like `name`, `activated` etc.
```
clample, line 424: I think that the same example appears earlier in the readme under the "Creating an API Check" section. I think that it's a bit confusing having almost the same example twice. 

To keep the readme size manageable, I think that it makes sense to have one section that gives a getting started example, then another section that is more of a reference.

umutuzgur:
Should we remove examples from this repo and point to the examples repo in the README?

umutuzgur:
Btw, I talked to Jan yesterday and he informed me that they working on moving all documentation to typescript to make checkly a typescript first company. Should we also focus mainly on TS examples in the readme as well?

README.md
stefanjudis: Should we have a  "Known limitations" section close to the `alpha` statement?

tnolet: @stefanjudis @clample let's add that later when we have more clarity around this. Might be next week already

README.md
tnolet: @clample I cleaned up the first example, it's much simpler now.

tnolet:
> Should we remove examples from this repo and point to the examples repo in the README?

@umutuzgur not right now, it's still useful to have this here.

umutuzgur (approved):

clample (approved):

PR #418, opened 2023-01-19
fix: use a new session parameter for absolute path [sc-0]

umutuzgur:

### Notes for the reviewer
Fix the relative path support. I was under the false assumption that checkFilePath was absolute


clample (approved):

clample:
@umutuzgur I think that something happened to the tests

umutuzgur:
It is the snapshots. The error changed now that relative path is working correctly

package/e2e/__tests__/__snapshots__/check-parse-error.spec.ts.snap
```diff
@@ -11,14 +11,17 @@ exports[`check parse error "checkly test" should return a clear error when there
     The following NPM dependencies were used, but aren't supported in the 
     runtimes.
     For more information, see https://www.checklyhq.com/docs/runtimes/.
-    	entrypoint.js imports unsupported dependencies:
+    	/home/runner/work/checkly-cli/checkly-cli/package/e2e/__tests__/fixtures/c
```
clample, line 14: I think that it's nicer to have the filename in the error path be relative to the root of the project like it was before. Also since this is an absolute path, it's going to be failing outside of GH CI.

umutuzgur: Gotcha. That requires some work but I will take a look at it

umutuzgur: This is seriously annoying... Now, I broke other tests

umutuzgur: @clample the check-dependency-parser.spec.ts expects absolute path errors. We gotta make a decision here which way we want to support

clample: Got it 👍 I think that we can merge the PR with either absolute or relative path errors for now, whichever is easier.

We can discuss more with Tim and Stefan later on in #proj-local-workflow about if we should go with relative or absolute path errors in the long run.

umutuzgur: I'm gonna switch it back to absolute path given it is a smaller change

clample (approved):

PR #422, opened 2023-01-20
feat: add query and header support for webhook alerts [sc-0]

umutuzgur:

### Notes for the reviewer
* Add headers and query params to webhook alerts
* Turn env vars also into an interface given it was the only `KeyValuePair` that was a class

### DIFF
--- a/package/src/constructs/query-param.ts
+++ b/package/src/constructs/query-param.ts
@@ -1,0 +1,2 @@
+import KeyValuePair from './key-value-pair'
+export type QueryParam = KeyValuePair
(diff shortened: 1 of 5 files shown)


package/src/constructs/query-param.ts
```diff
@@ -0,0 +1,2 @@
+import KeyValuePair from './key-value-pair'
+export type QueryParam = KeyValuePair
```
clample, line 2: Should we export these types from constructs/index.ts to make them public?

umutuzgur: Oh, good catch. I forgot to add them there

clample (approved):

PR #428, opened 2023-01-24
feat: expose `defineConfig` helper to leverage typings [sc-00]

kiroushi:

### New dependency submission
Expose `defineConfig` so users can have a typed noop when declaring configs:

Before:
```ts
const config = {
  // ...
}
```

After:
```ts
const config = defineConfig({
  // autocomplete works here
})
```


umutuzgur:
Is this a common way of doing this?

package/src/config.ts
```diff
@@ -0,0 +1,5 @@
+import type { ChecklyConfig } from './services/checkly-config-loader'
+
+export function defineConfig(config: ChecklyConfig): ChecklyConfig {
+  return config
+}
```
umutuzgur, line 5: Could you add this to the checkly-config-loader?

kiroushi: You mean relocating the helper?

umutuzgur (approved):

PR #429, opened 2023-01-24
feat: add `LocationCode` type for locations prop sanity [sc-00]

kiroushi:

### New dependency submission
Hint the users into using proper location codes instead of the unhelpful `string[]`.


umutuzgur:
Is this restrictive to the list there or can a user add newer ones if they are on an older cli and we just released a new DC? If it

kiroushi:
> Is this restrictive to the list there or can a user add newer ones if they are on an older cli and we just released a new DC?

@umutuzgur yes. Users can add a `global.d.ts` override like so:

```ts
declare module '@checkly/cli' {
  type LocationCode = string
}
```

I can add that to the docs.

kiroushi:
I stand corrected: type augmentation doesn’t seem to be really straightforward, closing this for now.

clample:
> I stand corrected: type augmentation doesn’t seem to be really straightforward, closing this for now.

@kiroushi does using `any` solve the issue? That should let users just bypass the type validation altogether.

Issue #426, opened 2023-01-23
Support ES Modules import syntax (for JavaScript)


clample:
We should already support ES Modules import syntax for TypeScript files. For JavaScript files, though, changes are needed in check dependency parsing.

PR #436, opened 2023-01-26
feat: add ES Modules support

clample:

### Notes for the reviewer

Resolves https://github.com/checkly/checkly-cli/issues/426

This PR adds support for ES Modules `import`+`export` syntax. The backend already supports this since we run all check files (even JS files) through the TypeScript compiler, which then outputs commonJS. All that's needed is to detect dependencies which are imported with ES Modules in `check-dependency-parser`.

This will only generally work for runtimes >= 2022.10, since we don't always apply the TypeScript compiler. Testing on an older runtime gave the following error:
```
    ERROR, 5:45:02 PM, Failed to compile check script: /tmp/1kqb2b8lddbsuwy/src/services/docs/__checks__/docs-search.spec.js:1
     import { defaults } from './defaults'
     ^^^^^^

    SyntaxError: 'import' and 'export' may only appear at the top level
        at Object.<anonymous> (/tmp/1kqb2b8lddbsuwy/script.js:61:47)
        at Module._compile (internal/modules/cjs/loader.js:1085:14)
        at Object.Module._extensions..js (internal/modules/cjs/loader.js:1114:10)
        at Module.load (internal/modules/cjs/loader.js:950:32)
```

It might make sense to add a better error message in the future. We could track the runtime version in `check-dependency-parser` and throw an error if we encounter an import expression. On the other hand, I'm not sure whether it's worth the extra effort.

### DIFF
--- a/package/src/services/__tests__/check-dependency-parser-fixtures/esmodules-example/dep1.js
+++ b/package/src/services/__tests__/check-dependency-parser-fixtures/esmodules-example/dep1.js
@@ -1,0 +1,2 @@
+export const value = 'Hello World'
+
--- a/package/src/services/__tests__/check-dependency-parser-fixtures/esmodules-example/dep2.js
+++ b/package/src/services/__tests__/check-dependency-parser-fixtures/esmodules-example/dep2.js
@@ -1,0 +1,2 @@
+const value = 'Another value'
+export default value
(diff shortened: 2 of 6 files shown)


umutuzgur (approved):
LGTM

PR #444, opened 2023-01-30
feat: add config properties descriptions [sc-0]

nahuelon:

### Notes for the reviewer
- Add configuration properties descriptions
- Remove duplicated properties and now `CheckConfigDefaults` picks properties from the base `interface CheckProps`


package/src/constructs/api-check.ts
```diff
@@ -17,16 +17,30 @@ enum BodyType {
...
+
 interface BasicAuth {
   username: string
   password: string
 }
 
 export interface Request {
   url: string,
-  method: string,
+  method: HttpRequestMethod,
```
kiroushi, line 37: I would use a union literal type here, is much more ergonomic from a user perspective. Even Axios does it that way: https://github.com/axios/axios/blob/18c1710f244f54d5891a017714bb58f21f5039bd/index.d.ts#L191

package/src/services/checkly-config-loader.ts
```diff
@@ -1,36 +1,40 @@
...
+   * Checks default configuration properties.
+   */
   checks?: CheckConfigDefaults & {
-    locations?: string[],
-    runtimeId?: string,
     checkMatch?: string,
     ignoreDirectoriesMatch?: string[],
     browserChecks?: CheckConfigDefaults & {
       testMatch?: string,
     },
```
kiroushi, line 33: Nit: it would be good to have a description on these too.

kiroushi (approved):
Looks neat 👍

package/src/constructs/alert-channel.ts
```diff
@@ -1,13 +1,36 @@
...
+  sslExpiryThreshold: number
+}
 
+/**
+ * Creates an Alert Channels
+ *
+ * @remarks
+ *
+ * This class make use of the Alert Channels endpoints
+ * listed {@link https://developers.checklyhq.com/reference/postv1alertchannels here}
```
umutuzgur, line 32: This is not true. We have different schemas for these constructs. I would not put the public api docs here. We can add the link when we put project/deploy behind v1 as well

umutuzgur: Sounds good 👍

umutuzgur (approved):

PR #446, opened 2023-01-31
chore: use collector class to gather parsing info [sc-0]

umutuzgur:

### Notes for the reviewer
* Moving check parsing to a full directory
* Split errors and helper types to different files
* Create a collector to collect parsing info. There will be more work about this in the following PR


package/src/constructs/browser-check.ts
```diff
@@ -2,7 +2,7 @@ import * as fs from 'fs'
 import * as path from 'path'
 import { Check, CheckProps } from './check'
 import { Session } from './project'
-import { Parser } from '../services/check-dependency-parser'
+import { Parser } from '../services/check-parser/parser'
```
clample, line 5: Nit: It could be handy to have an `index.ts` in `check-parser` that re-exports everything.

umutuzgur: I will most likely add that with the next PR

package/src/services/check-parser/collector.ts
```diff
@@ -0,0 +1,59 @@
...
+      throw new DependencyParseError(
+        this.entrypoint,
+        this.missingFiles,
+        this.unsupportedNpmDependencies,
+        this.parseErrors,
+      )
+    }
+  }
+
+  collect () {
```
clample, line 55: I think that the naming `collector.collect()` could be made a bit more precise. Maybe something like `collector.getDependencies()`?

umutuzgur: Sure

package/src/services/check-parser/parser.ts
```diff
@@ -106,11 +55,9 @@ export class Parser {
   * We can find all of the files we need to run the check by traversing this graph.
   * In this implementation, we use breadth first search.
   */
-    const missingFiles: string[] = []
-    const unsupportedNpmDependencies: UnsupportedNpmDependencies[] = []
-    const parseErrors: ParseError[] = []
-    const dependencies = new Set<string>()
-    dependencies.add(entrypoint)
+    const collector = new Collector(entrypoint)
+    // Stop adding entrypoint as dependency later on
```
clample, line 59: What do you mean by `Stop adding entrypoint as dependency later on`? Is this a TODO?

We'll need to do something to handle circular dependencies on the `entrypoint`, since you can have circular dependencies in JS. I think that just adding the `entrypoint` to the dependencies and removing it later is a nice solution.

clample: Ok, sounds good 👍

clample (approved):

Issue #423, opened 2023-01-20
TypeScript code can't import JS code


umutuzgur:
We currently can't import JS files from TS code. This is a common case from TS but our parser has a strict type assumption

PR #447, opened 2023-01-31
fix: allow ts files to import js [sc-0]

umutuzgur:

### Notes for the reviewer
* Allow more flexible setup for different extensions
* Allow TS files to import JS files
* Inline parsing errors to remove top try/catch

Closes https://github.com/checkly/checkly-cli/issues/423

### DIFF
--- a/package/src/services/check-parser/parser.ts
+++ b/package/src/services/check-parser/parser.ts
@@ -4,6 +4,7 @@ import * as acorn from 'acorn'
 import * as walk from 'acorn-walk'
 import { TSESTree } from '@typescript-eslint/typescript-estree'
 import { Collector } from './collector'
+import { DependencyParseError } from './errors'
 
 enum SupportedExtensions {
   JS = '.js',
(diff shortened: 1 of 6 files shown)


package/src/services/check-parser/parser.ts
```diff
@@ -46,89 +56,110 @@ export class Parser {
...
+        // @ts-ignore
+        walk.simple(ast, Parser.tsNodeVisitor(tsParser, localDependencies, npmDependencies))
+      } else {
+        throw new Error(`Unsupported file extension for ${filePath}`)
+      }
+    } catch (err) {
+      return {
+        localDependencies: Array.from(localDependencies),
+        npmDependencies: Array.from(npmDependencies),
+        error: err,
```
clample, line 161: Very go-lang style 😄

package/src/services/check-parser/parser.ts
```diff
@@ -46,89 +56,110 @@ export class Parser {
...
-    if (!filePath.endsWith(extension)) {
-      return filePath + extension
+  static readDependency (filePath: string, preferedExtenstion: SupportedExtensions) {
+    // Read the specific file if it has an extension
+    if (path.extname(filePath).length) {
+      const content = fs.readFileSync(filePath, { encoding: 'utf-8' })
+      return { filePath, content }
     } else {
-      return filePath
+      if (preferedExtenstion === SupportedExtensions.JS) {
```
clample, line 114: Could it ever be the case that the main entrypoint check is in JS, but it imports TS files? Or is it only possible for TS to import JS?

package/src/services/check-parser/parser.ts
umutuzgur: Yeah. Error probe operation, like parsing, are really hard to achieve with the try/catch syntax

package/src/services/check-parser/parser.ts
umutuzgur: That shouldn't be allowed. We will experience probably different resolving issues in the future but js import ts in a pure matter shouldn't be possible

clample (approved):

PR #457, opened 2023-02-01
feat: Allow configuring default alert channels in checkly config

### LINKED ISSUE #408: RFC: Allow for configuring default alert channels in checkly.config.js
Currently we allow for users to configure default check settings in `checkly.config.js`. We should also allow users to define alert channels in `checkly.config.js` and configure these to be defaults for all checks:

(`checkly.config.js`)
```
const sendDefaults = {
  sendFailure: true,
  sendRecovery: true,
  sendDegraded: false,
}

const smsChannel = new SmsAlertChannel('sms-channel-1', {
  phoneNumber: '123',
  ...sendDefaults
})

const emailChannel = new EmailAlertChannel('email-channel-1', {
  address: 'alerts@acme.com',
  ...sendDefaults
})

const config = {
  projectName: 'Website Monitoring',
  logicalId: 'website-monitoring-1',
  repoUrl: 'https://github.com/acme/website',
  checks: {
    alertChannels: [smsChannel, emailChannel],
  }
}

module.exports = config;
```

clample:

### Notes for the reviewer
Resolves https://github.com/checkly/checkly-cli/issues/408

Users might want to configure default alert channels which are applied to all checks. This currently isn't possible, though - mostly because users can't create constructs in the Checkly config file.

This PR allows users to create _any_ constructs in the Checkly config file [1]. Users can then add an `alertChannels` section to their config to be applied as a default. This works just like other check defaults.

In the implementation, I needed to make some changes to how constructs are registered. Previously, we would just register the constructs with the `Session.project`. The `Project` doesn't exist when the config file is being parsed, though. With this PR - we first try to register the constructs with the `Session.project`. If it doesn't exist, we try to register them with the `Session.checklyConfigConstructs` - an array that is only present when the Checkly config file is being parsed. After the Checkly config file is parsed and a project is created, all of the `checklyConfigConstructs` are registered with `Session.project`.

[1] Another option could be to only allow alert channels to be created in the config file

### DIFF
--- a/package/src/commands/deploy.ts
+++ b/package/src/commands/deploy.ts
@@ -29,7 +29,7 @@ export default class Deploy extends Command {
     const { flags } = await this.parse(Deploy)
     const { force, preview } = flags
     const cwd = process.cwd()
-    const checklyConfig = await loadChecklyConfig(cwd)
+    const { config: checklyConfig, constructs: checklyConfigConstructs } = await loadChecklyConfig(cwd)
     const { data: avilableRuntimes } = await runtimes.getAll()
     const project = await parseProject({
       directory: cwd,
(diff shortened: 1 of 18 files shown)


clample:
As discussed in Slack - we'll only allow alert channels to be created in the checkly.config.js for now. I'll update the PR accordingly.

PR #459, opened 2023-02-01
feat: add support for default alert channels

clample:
#### Notes for the reviewer

Resolves https://github.com/checkly/checkly-cli/issues/408

Users might want to configure default alert channels which are applied to all checks. This currently isn't possible, though - mostly because users can't create constructs in the Checkly config file.

This PR allows users to create alert channels in the Checkly config file [1]. Users can then add an alertChannels section to their config to be applied as a default. This works just like other check defaults.

In the implementation, I needed to make some changes to how constructs are registered. Previously, we would just register the constructs with the Session.project. The Project doesn't exist when the config file is being parsed, though. With this PR - we first try to register the constructs with the Session.project. If it doesn't exist, we try to register them with the Session.checklyConfigConstructs - an array that is only present when the Checkly config file is being parsed. After the Checkly config file is parsed and a project is created, all of the checklyConfigConstructs are registered with Session.project.

[1]  _only_ alert channels can be created in the Checkly config file. Creating other constructs will result in a clear error:
```
$ npx checkly deploy
    Error: Creating a ApiCheck construct in the Checkly config file isn't supported.
```

### DIFF
--- a/package/src/constructs/construct.ts
+++ b/package/src/constructs/construct.ts
@@ -1,20 +1,16 @@
-import { Session } from './project'
 import { Ref } from './ref'
 
-export class Construct {
+export abstract class Construct {
+  type: string
   logicalId: string
-  constructor (logicalId: string) {
+  constructor (type: string, logicalId: string) {
     this.logicalId = logicalId
+    this.type = type
   }
 
   ref () {
     return Ref.from(this.logicalId)
   }
 
-  register (type: string, logicalId: string, resource: any) {
-    if (!Session.project) {
-      throw new Error('A project is not registered to the session')
-    }
-    Session.project.addResource(type, logicalId, resource)
-  }
+  abstract synthesize(): any
 }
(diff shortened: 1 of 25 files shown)


package/src/constructs/construct.ts
```diff
@@ -1,20 +1,22 @@
 import { Session } from './project'
 import { Ref } from './ref'
 
-export class Construct {
+export abstract class Construct {
+  type: string
   logicalId: string
-  constructor (logicalId: string) {
+  constructor (type: string, logicalId: string) {
+    Session.validateCreateConstruct(this)
```
clample, line 8: Why validate this here, rather than in `Session.registerConstruct`?

If a user is trying to create an unsupported construct (f.ex. a `BrowserCheck`) in the checkly config file, some necessary `Session` properties won't be configured. Validating that we can register the construct early on makes sure that the user get's a clear error, rather than an error about a missing `Session` property - f.ex:
 https://github.com/checkly/checkly-cli/blob/827cf71b4e8a46648c0763a2ad685166948439b9/package/src/constructs/browser-check.ts#L59

umutuzgur (approved):
LGTM

PR #471, opened 2023-02-03
feat: support barrel files for ts and js [sc-0]

umutuzgur:

### Notes for the reviewer
Support barrel files for TS and JS

### DIFF
--- a/packages/cli/src/services/check-parser/parser.ts
+++ b/packages/cli/src/services/check-parser/parser.ts
@@ -22,24 +22,42 @@ type Module = {
   npmDependencies: Array<string>
 }
 
-enum SupportedExtensions {
+enum FileExtensions {
   JS = '.js',
-  TS = '.ts'
+  TS = '.ts',
+  PACKAGE = '/package.json',
+  INDEX_JS = '/index.js',
+  INDEX_TS = '/index.ts'
 }
+
+const JS_RESOLVE_ORDER = [
+  FileExtensions.JS,
+  FileExtensions.PACKAGE,
+  FileExtensions.INDEX_JS,
+]
+const TS_RESOLVE_ORDER = [
+  FileExtensions.TS,
+  FileExtensions.JS,
+  FileExtensions.PACKAGE,
+  FileExtensions.INDEX_TS,
+  FileExtensions.INDEX_JS,
+]
+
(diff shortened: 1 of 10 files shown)


packages/cli/src/services/check-parser/parser.ts
```diff
@@ -88,6 +106,11 @@ export class Parser {
     // We can add a not-null assertion operator (!).
     // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
       const item = bfsQueue.shift()!
+
+      if (item.filePath.endsWith(FileExtensions.PACKAGE)) {
+        // Hold info about the main file and doesn't need to be parser
```
clample, line 111:
```suggestion
        // Hold info about the main file and doesn't need to be parsed
```

packages/cli/src/services/check-parser/parser.ts
```diff
@@ -125,22 +148,35 @@ export class Parser {
...
+        deps.push({ filePath: fullPath, content })
+        if (extension === FileExtensions.PACKAGE) {
+          const { main } = JSON.parse(content)
+          if (!main || !main.length) {
+            // No main is defined. This means package.json doesn't have a specific entry
+            continue
+          }
+          const mainFile = path.join(filePath, main)
+          deps.push({
+            filePath: mainFile, content: fs.readFileSync(mainFile, { encoding: 'utf-8' }),
```
clample, line 176: Do we need to automatically import the `main` file from a `package.json`?

umutuzgur: Yeah. For example with the cli, it is `"main": "dist/index.js"`. If main doesn't exist, you go for `index.js` directly

clample: I'm still confused, though.  In a check, what's the advantage of importing `package.json` and having the `main` automatically imported as well?

umutuzgur: If we don't have the package.json, `import './module'` will point to `module/index.js`. If we have the package.json in the runtime, it should point to wherever the package.json specifies

clample: Ah got it. Thank you for clarifying.

clample (approved):
Looks nice!

Issue #445, opened 2023-01-31
`checkly login` server sometimes not closed


clample:
For `checkly login` the checkly-cli starts a server on port 4242. In the case of a failed login, it looks like this server isn't correctly closed. Later login attempts then fail with `Unable to start an HTTP server on port 4242. Error: listen EADDRINUSE: address already in use :::4242`. We should make sure that the login server is always closed.

clample:
Unfortunately, I was unable to reproduce this issue. The server on port 4242 is also part of the same process as the `npx checkly login` command. The port should therefore be freed whenever the `npx checkly login` command exits. It might be the case that an old `npx checkly login` command was still running in another terminal tab.

As a fix to allow multiple `npx checkly login` processes to run at once, I tried updating the server to bind to port 0. This would avoid the `EADDRINUSE` issue since we would always bind to a free port. It turns out that this won't work, though, since Auth0 requires us to specify a fixed port for the Auth0 redirect. They don't allow wildcards for the port when configuring a redirect ([relevant support ticket from 2020](https://community.auth0.com/t/are-port-wildcards-supported-pkce-for-desktop-apps/34132/5)).

I'll mark this as closed for now since I don't see anything that we can do to improve this.

clample:
Even if we can't fix the issue, we should have a better error message in case there's an `EADDRINUSE` error.

PR #475, opened 2023-02-03
chore: provide better error message for EADDRINUSE

clample:

### Notes for the reviewer
Resolves https://github.com/checkly/checkly-cli/issues/445

Running `npx checkly login` may fail if port 4242 is already in use. As a simple improvement, this PR gives a more clear error message in this case. If users continue running into this error, we can invest more effort into removing the error case altogether.

Screenshot 2023-02-03 at 14 03 17


umutuzgur (approved):
LGTM

PR #479, opened 2023-02-06
chore: cleanup constructs, command and synthesize [sc-0]

nahuelon:

### Notes for the reviewer
- Fix `npx checkly deploy --preview` output
- Add flag `--output` to show `deploy` response
- Enhance `npx checkly logout` when no user is logged in
- Remove internal `__checkFilePath` and `sourceFile` before the project deploy (avoid sending unnecessary properties to the API)


packages/cli/src/constructs/browser-check.ts
```diff
@@ -113,7 +113,6 @@ export class BrowserCheck extends Check {
       script: this.script,
       scriptPath: this.scriptPath,
       dependencies: this.dependencies,
-      sourceFile: this.__checkFilePath ?? this.scriptPath,
```
umutuzgur, line 116: We used to use these for the `checkly test` command. Isn't that the case anymore?

nahuelon: @umutuzgur , I started to refactor the Check constructs to handle internal attributes properly, but I think it will require more time to investigate all backward compatibility. Now, I simply loop the project's checks before deploying them and remove `__checkFilePath` and `sourceFile` interval attributes.

packages/cli/src/commands/logout.ts
```diff
@@ -15,10 +15,12 @@ export default class Logout extends Command {
   async run (): Promise<void> {
     const { flags } = await this.parse(Logout)
     const { force } = flags
+    const accountName = config.data.get('accountName')
 
     if (!force) {
-      const message = `You are about to clear your local session of
-        \`${config.data.get('accountName')}\`, do you want to continue?`
+      const message = `You are about to clear your local session${accountName
```
clample, line 21: Looks good 👍

packages/cli/src/commands/deploy.ts
```diff
@@ -16,6 +16,11 @@ export default class Deploy extends Command {
       description: 'Show state preview',
       default: false,
     }),
+    output: Flags.boolean({
```
clample, line 19: Do we have an issue or Slack discussion about `--output`? I'm just wondering how it should behave.

We should maybe add this to the README for `npx checkly deploy` as well https://github.com/checkly/checkly-cli/tree/b609488f438721e78f56a5a180d813e57a239268#npx-checkly-deploy

packages/cli/src/commands/deploy.ts
```diff
@@ -50,7 +55,7 @@ export default class Deploy extends Command {
     // We can use a null-assertion operator safely since account ID was validated in auth-check hook
     const { data: account } = await api.accounts.get(config.getAccountId()!)
 
-    if (!force) {
+    if (!force && !preview) {
```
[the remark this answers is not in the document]
nahuelon: @clample , my plan was to not ask users to confirm (a deploy) if they are only requesting the `preview`.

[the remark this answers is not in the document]
tnolet: Was not a specific request but makes sense. If there is nothing to destroy, why ask for confirmation?

packages/cli/src/commands/deploy.ts
nahuelon: @clample , no we don't. I found it useful because the `/next/project/deploy` returns all the project information and the `diff` applied. So, I found it useful to debug which resources were modified (created/updated/deleted).

packages/cli/src/commands/deploy.ts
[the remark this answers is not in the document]
clample: I was misunderstanding the code. This makes sense now, sorry for the comment.

clample (approved):

packages/cli/src/commands/deploy.ts
clample: Thanks for clarifying. That makes sense.

Issue #472, opened 2023-02-03
add `optional` and `default` values to some properties


tnolet:
The following properties can easily be given defaults and should be marked optional in the types

# ApiCheck

https://github.com/checkly/checkly-cli/blob/5a18ea807be9aeb9000cf7b803c5ecc56be3269c/packages/cli/src/constructs/api-check.ts#L81

- `degradedResponseTime: 5`
- `maxResponseTime: 20`
-  `followRedirects`: true,
-  `skipSsl`: false,
-  `assertions`: []

```
export class ApiCheck extends Check {
  request: Request
  localSetupScript?: string
  localTearDownScript?: string
  degradedResponseTime: number
  maxResponseTime: number
```

# AlertChannelProps

https://github.com/checkly/checkly-cli/blob/5a18ea807be9aeb9000cf7b803c5ecc56be3269c/packages/cli/src/constructs/alert-channel.ts#L3

I think all the props here can have sane defaults and should be optional:

- `sendRecovery: true`
- `sendFailure: true`
- `sendDegraded: true`
- `sslExpiry: true`
- `sslExpiryThreshold: 30`

```
export interface AlertChannelProps {
  /**
   * Determines if an alert should be send for check recoveries.
   */
  sendRecovery: boolean
  /**
   * Determines if an alert should be send for check failures.
   */
  sendFailure: boolean
  /**
   * Determines if an alert should be send when a check is degraded.
   */
  sendDegraded: boolean
  /**
   * Determines if an alert should be send for expiring SSL certificates.
   */
  sslExpiry: boolean
  /**
   * At what moment in time to start alerting on SSL certificates.
   */
  sslExpiryThreshold: number
```

# CheckGroup

https://github.com/checkly/checkly-cli/blob/5a18ea807be9aeb9000cf7b803c5ecc56be3269c/packages/cli/src/constructs/check-group.ts#L30

these should be optional like they are for the `Check` interface

 - `activated`: true,
-  `muted`: false,
-  `runtimeId`: account default,
-  `tags`: []
-   `environmentVariables`: [],
-  `apiCheckDefaults`: {},
-  `browserCheckDefaults`: {},

clample:
For the implementation, I think that we should try to put these in Joi defaults in the backend.

We could also add defaults in the checkly-cli, but I think that having the defaults in the backend gives us more flexibility for updating things later on.

tnolet:
[...]

PR #480, opened 2023-02-06
feat: enable default values for API checks

clample:

### Notes for the reviewer
Relates to https://github.com/checkly/checkly-cli/issues/472

We already have sensible defaults for API check values in the backend. These are added via Joi and the DB (in the case of `followRedirects`). This PR makes some fields optional - the defaults will be used in this case.

* `degradedResponseTime`: 10000 (higher than the one mentioned in the issue, but it's worth the consistency with the public API + TF)
* `maxResponseTime`: 20000
* `followRedirects`: true
* `skipSsl`: false
* `assertions`: []


umutuzgur (approved):
LGTM

PR #482, opened 2023-02-06
feat: enable default values for groups

clample:

## Affected Components
* [x] CLI
* [ ] Create CLI
* [ ] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer
Relates to https://github.com/checkly/checkly-cli/issues/472

We already have sensible defaults for check groups in the backend - added in the level Joi and DB level. With this PR, users can more easily create check groups with the default values:

`activated`: true
`muted`: false
`runtimeId`: `null`. In this case the runtimeId from the account will be used as a fallback.
`tags`: `[]`
`concurrency`: 3. This field wasn't mentioned in the GH issue, but I think that it makes sense to enable the default here as well.
`environmentVariables`: `[]`

`apiCheckDefaults` and `browserCheckDefaults` can also be skipped. This default comes from checkly-cli rather than the backend.

### DIFF
--- a/packages/cli/src/constructs/check-group.ts
+++ b/packages/cli/src/constructs/check-group.ts
@@ -35,15 +35,15 @@ export interface CheckGroupProps {
   /**
    * Determines if the checks in the group are running or not.
    */
-  activated: boolean
+  activated?: boolean
   /**
    * Determines if any notifications will be send out when a check in this group fails and/or recovers.
    */
-  muted: boolean
+  muted?: boolean
   /**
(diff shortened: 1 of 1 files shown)


umutuzgur:
Do we have a `concurrency` default? We didn't have it in the past so that is why I'm asking

umutuzgur (approved):

clample:
> Do we have a concurrency default? We didn't have it in the past so that is why I'm asking

We do have one in the public API Joi schema. I sent more details in private chat.

Issue #399, opened 2023-01-12
Improve error handling for non-allowed runLocation


clample:
### Context
When a user runs `checkly test` with a non-supported run location, we currently print the full failed HTTP response for each check that runs:

```
$ npx checkly test --location does-not-exist

Running 4 checks in does-not-exist.

src/__checks__/home.check.js
  - 404 page
  - Homepage
  - Homepage - fetch stats
src/services/docs/__checks__/docs-search.spec.js
  - docs-search.spec.js

4 pending, 4 total

Scheduler failure AxiosError: Request failed with status code 400
    at settle (/Users/clample/projects/checkly-cli/package/node_modules/axios/dist/node/axios.cjs:1855:12)
    at IncomingMessage.handleStreamEnd (/Users/clample/projects/checkly-cli/package/node_modules/axios/dist/node/axios.cjs:2712:11)
    at IncomingMessage.emit (node:events:525:35)
    at endReadableNT (node:internal/streams/readable:1358:12)
    at processTicksAndRejections (node:internal/process/task_queues:83:21) {
  code: 'ERR_BAD_REQUEST',
... lot't of additional output...
```

### Proposal
We should print something like:
```
$ npx checkly test --location does-not-exist

Unable to run checks on unsupported location "does-not-exist". Supported locations are:

eu-central-1
us-west-2
...print all locations...
```

### Implementation
One question is whether we want to keep a list of all supported locations in the client, or if we want to rely on information from the backend. I think that relying on information from the backend would be better, since users can use new locations without needing to upgrade the checkly-cli version.

tnolet:
@clample I think relying on the backend is also better.

PR #483, opened 2023-02-06
feat: refactor to add error handling [sc-0]

### LINKED ISSUE #398: Improve error message when an API key is deleted
Currently, when the API key has been deleted on Checkly we fail with the following error messages:
```
$ npx checkly whoami
Failed to find an account corresponding to the account id
```

```
$ npx checkly deploy
    AxiosError: Request failed with status code 401
    Code: ERR_BAD_REQUEST
```

Most users will be developers and will be able to figure out the issue with the 401 authentication failure. Still, it could improve DX if we return a more specific error for the case where an API key and Account ID is set, but we still get a 401:

> Authentication failed with Account ID "<account UUID>" and API key "<partly redacted API key>"

nahuelon:

## Affected Components
* [x] CLI
* [ ] Create CLI
* [ ] Test
* [x] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer
- Refactor Commands extending from a `BaseCommand` or `AuthCommand`
- Remove `Hooks` and move the logic to the `init()` method in the Base Commands
- Validate credentials and session on every auth command (`whoami`, `switch`, `test`, `deploy`)
- Add centralized error handling and standarize them, replacing all `console.error` and `Command.error` with `throw new Error()`
- Validate supported location when `npx checkly test --location=<one_location>`

- Complete `README.md` command arguments.

> Resolves #398 #399


umutuzgur:
You can also remove the `auth` fields in the commands. That was my custom solution because I didn't know about the init commands

packages/cli/src/commands/switch.ts
```diff
@@ -22,37 +24,41 @@ export default class Switch extends Command {
...
 
-    config.data.set('accountId', id)
-    config.data.set('accountName', name)
+      config.data.set('accountId', id)
+      config.data.set('accountName', name)
 
-    console.info(`Account switched to ${chalk.bold.blue(name)}`)
+      this.log(`Account switched to ${chalk.bold.blue(name)}`)
+    } catch (err: any) {
+      throw new Error(`Failed to switch account. ${err.response.data.message}`)
```
clample, line 61: Same comment as for `deploy.ts` - I think that `err.response.data.message` won't always be defined.

packages/cli/src/commands/test.ts
```diff
@@ -164,45 +163,51 @@ export default class Test extends Command {
         hasFailures: true,
         runError: message,
       })
-      process.exitCode = 1
+      throw new Error(`Scheduler failure:, ${message}.`)
```
clample, line 166: I think that we shouldn't throw an error in this case. Other checks are still running as part of `checkly test` - I think that we should let them all finish.

packages/cli/src/commands/test.ts
```diff
@@ -164,45 +163,51 @@ export default class Test extends Command {
...
   async prepareRunLocation (
     configOptions: { runLocation?: string, privateRunLocation?: string } = {},
     cliFlags: { runLocation?: string, privateRunLocation?: string },
   ): Promise<RunLocation> {
     // Command line options take precedence
     if (cliFlags.runLocation) {
-      return { type: 'PUBLIC', region: cliFlags.runLocation }
+      const { data: availableLocations } = await api.locations.getAll()
+      if (availableLocations.some(l => l.region === cliFlags.runLocation)) {
+        return Promise.resolve({ type: 'PUBLIC', region: cliFlags.runLocation })
```
clample, line 180: Since this an `async` function, can we just return the value without calling `Promise.resolve`?

packages/cli/src/commands/test.ts
```diff
@@ -164,45 +163,51 @@ export default class Test extends Command {
...
-    const privateLocation = privateLocations.find(({ slugName }) => slugName === privateLocationSlugName)
-    if (privateLocation) {
-      return { type: 'PRIVATE', id: privateLocation.id, slugName: privateLocationSlugName }
+    try {
+      const { data: privateLocations } = await api.privateLocations.getAll()
+      const privateLocation = privateLocations.find(({ slugName }) => slugName === privateLocationSlugName)
+      if (privateLocation) {
+        return { type: 'PRIVATE', id: privateLocation.id, slugName: privateLocationSlugName }
+      }
+      // We can use a null-assertion operator safely since account ID was validated in auth-check hook
```
clample, line 206: This comment isn't needed anymore now that the `!` operator is gone from `config.getAccountId()!`

clample (changes requested):

packages/cli/src/commands/deploy.ts
```diff
@@ -83,16 +83,17 @@ export default class Deploy extends Command {
...
+        this.log(data)
       }
       if (!preview) {
-        console.info(`Successfully deployed project "${project.name}" to account "${account.name}".`)
+        this.log(`Successfully deployed project "${project.name}" to account "${account.name}".`)
       }
     } catch (err: any) {
       if (err?.response?.status === 400) {
-        console.error(`Failed to deploy the project due to a missing field. ${err.response.data.message}`)
+        throw new Error(`Failed to deploy the project due to a missing field. ${err.message}`)
```
clample, line 93: I know that I mentioned we shouldn't use `err.response.data.message`. In this case, though, (when there's a `400` response code) - I think that it's OK to leave the error as it was before.

```suggestion
        throw new Error(`Failed to deploy the project due to a missing field. ${err.response.data.message}`)
```

packages/cli/src/rest/api.ts
```diff
@@ -33,7 +34,34 @@ export function getDefaults () {
...
+
+  const accountId = config.getAccountId()
+  const apiKey = config.getApiKey()
+
+  try {
+    // check if credentials works
+    await accounts.get(accountId)
+    return true
+  } catch (err: any) {
+    // if fails remove stored credentials
```
clample, line 54: I think that we can just leave the credentials. Probably the user will run another `npx checkly login` command and login again.

I'm just thinking that we should avoid taking a permanent action like wiping the auth config without the user explicitly requesting that. Maybe there was just a temporary bug that caused the 401 for example.

packages/cli/src/services/check-runner.ts
```diff
@@ -91,7 +91,7 @@ export default class CheckRunner extends EventEmitter {
       await checksApi.run(checkRun)
     } catch (err: any) {
       if (err?.response?.status === 402) {
-        const errorMessage = `Failed to run a check. ${err.response.data.message}`
+        const errorMessage = `Failed to run a check. ${err.message}`
```
clample, line 94: I think that we can leave the `err.response.data.message` in this case. I think that it should be fine since we're already checking that the error has a specific `402` status code.

clample (approved):

PR #488, opened 2023-02-07
chore: call synthesize() later on

clample:

## Affected Components
* [x] CLI
* [ ] Create CLI
* [ ] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer
Currently the `Project` tracks all of the `synthesize()` result of checks, groups, and alert channels rather than tracking the original construct. This means that after parsing the project, we only have access to the results of `synthesize()`.

This has worked OK so far, but it also has a few small issues. In particular, we need to put internal properties in the `synthesize()` result. These internal properties need to be removed later on in `checkly deploy`  before messaging the backend https://github.com/checkly/checkly-cli/blob/f2ab503c5997de1fa9b1f5565bd3872395e1b48e/packages/cli/src/commands/deploy.ts#L72-L82

In this PR, I make `Project` track the original construct objects. We can then access the construct method and any internal properties throughout the CLI codebase. `synthesize()` is only called at the last moment - when serializing a request before sending to the backend.

### DIFF
--- a/packages/cli/src/commands/deploy.ts
+++ b/packages/cli/src/commands/deploy.ts
@@ -68,19 +68,6 @@ export default class Deploy extends AuthCommand {
 
     try {
       const projectPayload = project.synthesize()
-
-      // TODO: refactor Check construct to handle internal attributes properly
-      projectPayload.checks = Object.keys(projectPayload.checks).reduce((acc, checkLogicalId) => {
-        const check = projectPayload.checks[checkLogicalId]
-        delete check.__checkFilePath
-        delete check.sourceFile
-        acc = {
-          ...acc,
-          [checkLogicalId]: check,
-        }
-        return acc
-      }, {})
-
       const { data } = await api.projects.deploy(projectPayload, { dryRun: preview })
       if (preview || output) {
         this.log(data)
--- a/packages/cli/src/commands/test.ts
+++ b/packages/cli/src/commands/test.ts
@@ -12,6 +12,7 @@ import { loadChecklyConfig } from '../services/checkly-config-loader'
 import { filterByFileNamePattern, filterByCheckNamePattern } from '../services/test-filters'
 import type { Runtime } from '../rest/runtimes'
 import { AuthCommand } from './authCommand'
+import { BrowserCheck } from '../constructs'
 
 const DEFAULT_REGION = 'eu-central-1'
 
@@ -113,13 +114,16 @@ export default class Test extends AuthCommand {
       }, <Record<string, Runtime>> {}),
       checklyConfigConstructs,
     })
-    const { checks: checksMap, groups: groupsMap } = project.data
-    const checks = Object.entries(checksMap)
-      .filter(([_, check]) => {
-        return filterByFileNamePattern(filePatterns, check.scriptPath) ||
-          filterByFileNamePattern(filePatterns, check.__checkFilePath)
+    const checks = Object.entries(project.data.checks)
+      .filter(([, check]) => {
+        if (check instanceof BrowserCheck) {
+          return filterByFileNamePattern(filePatterns, check.scriptPath) ||
+            filterByFileNamePattern(filePatterns, check.__checkFilePath)
+        } else {
+          return filterByFileNamePattern(filePatterns, check.__checkFilePath)
+        }
       })
-      .filter(([_, check]) => {
+      .filter(([, check]) => {
         return filterByCheckNamePattern(grep, check.name)
       })
       .map(([key, check]) => {
@@ -135,11 +139,6 @@ export default class Test extends AuthCommand {
             })
           }
         }
-        // TODO: Add the group to check in a cleaner form
-        if (check.groupId) {
-          check.group = groupsMap[check.groupId.ref]
-          delete check.groupId
-        }
         return check
       })
     const reporter = isCI ? new CiReporter(location, checks, verbose) : new ListReporter(location, checks, verbose)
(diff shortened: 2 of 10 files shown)


packages/cli/src/commands/test.ts
```diff
@@ -108,11 +109,14 @@ export default class Test extends Command {
       }, <Record<string, Runtime>> {}),
       checklyConfigConstructs,
     })
-    const { checks: checksMap, groups: groupsMap } = project.data
-    const checks = Object.entries(checksMap)
+    const checks = Object.entries(project.checks)
       .filter(([_, check]) => {
```
nahuelon, line 113:
```suggestion
      .filter(([, check]) => {
```

packages/cli/src/commands/test.ts
```diff
@@ -108,11 +109,14 @@ export default class Test extends Command {
...
-        return filterByFileNamePattern(filePatterns, check.scriptPath) ||
-          filterByFileNamePattern(filePatterns, check.__checkFilePath)
+        if (check instanceof BrowserCheck) {
+          return filterByFileNamePattern(filePatterns, check.scriptPath) ||
+            filterByFileNamePattern(filePatterns, check.__checkFilePath)
+        } else {
+          return filterByFileNamePattern(filePatterns, check.__checkFilePath)
+        }
       })
       .filter(([_, check]) => {
```
nahuelon, line 121:
```suggestion
      .filter(([, check]) => {
```

nahuelon (dismissed):

packages/cli/src/commands/test.ts
clample: Good idea, we can get rid of the lint error this way.

packages/cli/src/commands/deploy.ts
```diff
@@ -68,19 +68,6 @@ export default class Deploy extends Command {
 
     try {
       const projectPayload = project.synthesize()
-
-      // TODO: refactor Check construct to handle internal attributes properly
-      projectPayload.checks = Object.keys(projectPayload.checks).reduce((acc, checkLogicalId) => {
-        const check = projectPayload.checks[checkLogicalId]
-        delete check.__checkFilePath
```
clample, line 75: We no longer add `__checkFilePath` and `sourceFile` in `synthesize()`, so this isn't necessary.

packages/cli/src/commands/test.ts
```diff
@@ -108,13 +109,16 @@ export default class Test extends Command {
...
       checklyConfigConstructs,
     })
-    const { checks: checksMap, groups: groupsMap } = project.data
-    const checks = Object.entries(checksMap)
-      .filter(([_, check]) => {
-        return filterByFileNamePattern(filePatterns, check.scriptPath) ||
-          filterByFileNamePattern(filePatterns, check.__checkFilePath)
+    const checks = Object.entries(project.checks)
+      .filter(([, check]) => {
+        if (check instanceof BrowserCheck) {
```
clample, line 114: `check` now has the type `Check` rather than `any`. I changed this since TS had an error about `scriptPath` not being on `Check`. Another solution would be to just use `any` here.

packages/cli/src/commands/test.ts
```diff
@@ -130,11 +134,6 @@ export default class Test extends Command {
             })
           }
         }
-        // TODO: Add the group to check in a cleaner form
```
clample, line 133: I've moved this into `CheckRunner`

packages/cli/src/services/test-filters.ts
```diff
@@ -3,9 +3,9 @@ export function filterByCheckNamePattern (checkNamePattern = '', checkName: stri
   return re.test(checkName)
 }
 
-export function filterByFileNamePattern (filePatterns: Array<string> = [], path: string) {
+export function filterByFileNamePattern (filePatterns: Array<string> = [], path: string|undefined) {
```
clample, line 6: `path` was actually `string|undefined`before (it was called with `scriptPath`, which is null for API checks). Typechecking was still passing, though, since we were using `any` to represent checks.

`re.test(undefined)` works fine and returns `true`, so this is OK. There are no behavioral changes here - this just updates types to satisfy TS.

packages/cli/src/constructs/check.ts
```diff
@@ -76,10 +76,8 @@ export abstract class Check extends Construct {
   alertChannels?: Array<AlertChannel>
   __checkFilePath?: string // internal variable to filter by check file name from the CLI
 
-  static readonly __checklyType = 'checks'
```
umutuzgur, line 79: Any reason we removed these as well?

clample: This was used for registering the construct with the `Project`.

Before, project stored all of the `synthesized()` constructs - so `Project` wouldn't know what type it was receiving. 

Now that we're actually registering a `Construct` object. We don't need the type anymore since we can just check the type with `instanceof`.

https://github.com/checkly/checkly-cli/blob/903f12f6a95ea7b8d2556fe61f50515a6860d6c9/packages/cli/src/constructs/project.ts#L48-L60

umutuzgur: This works in our case since we have a few types but becomes just too much we have 20-30 different construct classes. This was the reason I added this as a "premature optimization"

umutuzgur: @clample I think I wanna keep the types. I have some plans with them in the future to add further types to the return values for `synthesize` and https://github.com/checkly/checkly-cli/pull/488/files#diff-f71a863e9534b5c479e26efeae202cafb5f80f992820b15a57d657e2e6a360e6R68-R73

clample: I'll try to switch it back to types

umutuzgur:
@clample The code looks great! If you resolve the merge conflict, we can merge it today

clample:
Thanks @umutuzgur! I fixed the merge conflicts, I think that it should be all-set.

umutuzgur (approved):
LGTM

PR #493, opened 2023-02-08
feat: add logicalId validation [sc-0]

### LINKED ISSUE #487: story: Restrict logical id to alphanumeric with special characters [a-zA-Z0-9_-/#]
### Story description

We want to help our users avoid using ids prone to making mistakes, like empty spaces or tabs. I suggest using `[a-zA-Z0-9_-/#]` or simply `[\w-/#]` to make sure they have a simpler but still extensive logical ids combination. This should be a FE and a BE restriction

### Shortcut link

_No response_

### Additional resources

_No response_

nahuelon:

## Affected Components
* [x] CLI
* [ ] Create CLI
* [ ] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer
- Add `logicalId` validation.
- Now the `Project` construct class extends from `Construct`

> Resolves #487

## New Dependency Submission
- Important: we have to confirm all Production/Staging projects compliant with the new `logicalId` format restriction.

### DIFF
--- a/packages/cli/src/constructs/construct.ts
+++ b/packages/cli/src/constructs/construct.ts
@@ -6,12 +6,9 @@ export abstract class Construct {
   type: string
   logicalId: string
   constructor (type: string, logicalId: string) {
-    if (!/^[A-Za-z0-9_\-/#.]+$/.test(logicalId)) {
-      throw new ValidationError('The `logicalId` must includes only allowed characters [A-Za-z0-9_-/#.]')
-    }
-    Session.validateCreateConstruct(this)
     this.logicalId = logicalId
     this.type = type
+    Session.validateCreateConstruct(this)
   }
 
   ref () {
--- a/packages/cli/src/constructs/construct.ts
+++ b/packages/cli/src/constructs/construct.ts
@@ -7,7 +7,7 @@ export abstract class Construct {
   logicalId: string
   constructor (type: string, logicalId: string) {
     if (!/^[A-Za-z0-9_\-/#.]+$/.test(logicalId)) {
-      throw new ValidationError('The `logicalId` must includes only allowed characters [A-Za-z0-9_-/#]')
+      throw new ValidationError('The `logicalId` must includes only allowed characters [A-Za-z0-9_-/#.]')
     }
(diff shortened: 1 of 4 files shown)


packages/cli/src/constructs/construct.ts
```diff
@@ -1,10 +1,14 @@
 import { Session } from './project'
 import { Ref } from './ref'
+import { ValidationError } from './validator-error'
 
 export abstract class Construct {
   type: string
   logicalId: string
   constructor (type: string, logicalId: string) {
+    if (!/^[A-Za-z0-9_\-/#.]+$/.test(logicalId)) {
```
umutuzgur, line 9: Any reason to add `.` support as well? Did you find anyone using it?

umutuzgur: I just saw your comment in the issue, now I got it 👍

umutuzgur:
Could you also add it here @nahuelon https://github.com/checkly/checkly-cli/blob/main/packages/cli/src/constructs/project.ts#L35-L43?

umutuzgur:
Can we just a create helper that checks the logical id and verifies it in both classes? We can also remove project from the constructs folder as the user can't do anything with it atm

nahuelon:
@umutuzgur , we don't need a helper since the `logicalId` validation is centralized in one place (for all resource contructs including the `Project`). To remove the `project` from the constructs folder we need a small refactor.

packages/cli/src/constructs/project.ts
```diff
@@ -26,13 +26,16 @@ export class Project {
...
   }
 
+  static readonly __checklyType = 'project'
+
   /**
-   * Constructs the Email Alert Channel instance
+   * Constructs the Project instance
    *
-   * @param logicalId unique project name
+   * @param logicalId unique project identificator
```
umutuzgur, line 34: identifier

umutuzgur:
I suggested that because Project is not really a construct. However, the solution solves the problem so I'm cool with it

umutuzgur (dismissed):

umutuzgur (approved):

Issue #489, opened 2023-02-08
story: Add assertion types for api checks


umutuzgur:
### Story description

It is currently quite hard to set assertions using MaC or Terraform in that matter as well given the object is quite complex. We should add different types for different assertions and use the power of TS interfaces. Our backend should have more info about these assertion types

PR #500, opened 2023-02-09
feat: add `AssertionBuilder` to improve API check assertion usability

clample:

## Affected Components
* [x] CLI
* [ ] Create CLI
* [ ] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer

> Resolves https://github.com/checkly/checkly-cli/issues/489

Currently it's tricky to create API check assertions. We don't list the valid `source` and `comparison` values anywhere. It's also easy to create invalid assertions. For example, the `regex` field isn't needed when `source` is `STATUS_CODE`.

This PR introduces `AssertionBuilder`. This should make the possible assertion types more discoverable from IDE's autocomplete (see screenshot below). It also aims to avoid invalid assertion types from being created. 

It's still possible for users to use the old method of creating assertions. This is not a breaking change at all - just a new builder for `Assertion` objects.

#### Test Plan

I've updated the examples and made sure that `npx checkly test` and `npx checkly deploy` work. I've also checked that examples with the old assertion setup still work.

I'll also do a close review of the code to make sure that it matches our assertion checking logic and the assertion types already produced by webapp.

#### IDE Completion:

Screenshot 2023-02-09 at 16 11 31


tnolet:
@clample this is already awesome. Please do not update the docs in the `README.md` as I'm moving the docs to the website. Updating examples is great.

clample:
Thanks for checking it out @tnolet 👍 

I'll leave the docs untouched.

packages/cli/src/constructs/api-check.ts
```diff
@@ -2,12 +2,166 @@ import { Check, CheckProps } from './check'
...
+    return this.toAssertion(AssertionComparison.NOT_NULL)
+  }
+
+  /** @private */
+  private toAssertion (comparison: AssertionComparison, target?: string|number|boolean): Assertion {
+    return {
+      source: this.source,
+      comparison,
+      property: this.property ?? '',
+      target: target !== undefined ? target.toString() : '',
```
umutuzgur, line 161: target?.toString() ?? ''

clample: Nice, that makes sense

umutuzgur (approved):
Looks great!

clample:
I'll merge it now. I didn't see that it was approved.

Issue #498, opened 2023-02-09
bug: unclear error when `cli` section not set in Checkly config


clample:
### Node.js version

v16.17.0

### NPM version

8.15.0

### @checkly/cli version

@checkly/cli/0.3.14

### Steps to reproduce

Omit the `cli` section from a `checkly.config.js` file:

```
const config = {
    projectName: 'My First Checkly Project',
    logicalId: 'checkly-project-1',
    checks: {
        activated: true,
        muted: false,
        runtimeId: '2022.10',
        frequency: 10,
        locations: ['us-east-1', 'eu-west-1'],
        browserChecks: {
            testMatch: '**/*.spec.ts',
        },
    }
}

module.exports = config;
```

Then run `npx checkly test`

### What is expected?

Ideally, `npx checkly test` should pass. If the `cli` section is _really_ necessary, then there should be a clear validation error.

In particular, I think that we should aim for something like:
```
$ npx checkly test
    Required field `cli` was not found in the Checkly config file. 
```

### What is actually happening?

There is an obscure error:

```
$ npx checkly test
    TypeError: Cannot read properties of undefined (reading 'cli')
```

Running with `DEBUG=*` yields more info:
```
$ DEBUG=checkly* npx checkly test  
  checkly:@oclif/plugin-warn-if-update-available:hooks:init start (require) ...
  /node_modules/@oclif/plugin-warn-if-update-available/lib/hooks/init/check-update.js +0ms
  checkly:@oclif/plugin-warn-if-update-available:hooks:init done +2ms
  checkly:test init version: @oclif/core@2.0.7 argv: [] +0ms
TypeError: Cannot read properties of undefined (reading 'cli')
    at Test.run (.../node_modules/@checkly/cli/src/commands/test.ts:98:66)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at async Test._run (.../node_modules/@oclif/core/lib/command.js:107:22)
    at async Config.runCommand (.../node_modules/@oclif/core/lib/config/config.js:296:25)
```

### Any additional comments?

This is likely an issue for other fields in the `checkly.config.js` file as well. We should probably add some validation with clear errors here.

clample:
Taking a closer look at the error, I think that I may have misinterpreted the error message in this issue.
```
TypeError: Cannot read properties of undefined (reading 'cli')
    at Test.run (.../node_modules/@checkly/cli/src/commands/test.ts:98:66)
```

[...]

PR #503, opened 2023-02-10
fix: handle no cli option and push verbose config [sc-0]

umutuzgur:

## Affected Components
* [x] CLI
* [ ] Create CLI
* [ ] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer
* Handle the missing cli field in the config
* Add the verbose flag support to the config as well
* Update the test to use the verbose flag as well to avoid false negatives

Tested the cases manually using the advanced-project

### DIFF
--- a/packages/cli/src/commands/test.ts
+++ b/packages/cli/src/commands/test.ts
@@ -63,8 +63,8 @@ export default class Test extends AuthCommand {
     }),
     verbose: Flags.boolean({
       char: 'v',
-      default: false,
       description: 'Always show the logs of the checks.',
+      allowNo: true,
     }),
   }
 
@@ -89,7 +89,7 @@ export default class Test extends AuthCommand {
       'env-file': envFile,
       list,
       timeout,
-      verbose,
+      verbose: verboseFlag,
     } = flags
     const cwd = process.cwd()
     const filePatterns = argv as string[]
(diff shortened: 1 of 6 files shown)


packages/cli/src/commands/test.ts
```diff
@@ -89,14 +89,15 @@ export default class Test extends AuthCommand {
...
       timeout,
-      verbose,
+      verbose: verboseFlag,
     } = flags
     const cwd = process.cwd()
     const filePatterns = argv as string[]
 
     const testEnvVars = await getEnvs(envFile, env)
     const { config: checklyConfig, constructs: checklyConfigConstructs } = await loadChecklyConfig(cwd)
     const location = await this.prepareRunLocation(checklyConfig.cli, { runLocation, privateRunLocation })
```
clample, line 99: Taking a closer look, I'm a bit confused by the error that I reported in https://github.com/checkly/checkly-cli/issues/498

The exact error message was
```
TypeError: Cannot read properties of undefined (reading 'cli')
    at Test.run (.../node_modules/@checkly/cli/src/commands/test.ts:98:66)
```

That must mean that `checklyConfig` here was undefined - not that `checklyConfig` was missing the `cli` property.

clample: Or is this issue fixed by adding the default `cliFlags` value of `{}` for `prepareRunLocation`?

clample: Were you able to reproduce the issue by removing the `cli` section?

Do you have any thoughts on how we can avoid a regression here? Will the TS compiler help us at all? Should we add a test case?

umutuzgur: Actually, you are right. This is about config being null or undefined. I need to check the value returned from the config files. This PR doesn't close that issue but we can still merge it

clample: Sounds good 👍 

Could you please remove the `Resolves https://github.com/checkly/checkly-cli/issues/498` from the PR description? Otherwise I think that the issue will be automatically closed and we might lose track of it.

umutuzgur: Makes sense

clample (approved):
Looks good!

PR #505, opened 2023-02-10
fix: handle missing or empty config exports [sc-0]

umutuzgur:

## Affected Components
* [x] CLI
* [ ] Create CLI
* [ ] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer
* Make repoUrl optional as it is not required and didn't want to add validation
* Add projectName and logicalId validations to the config

> Resolves #[498]

### DIFF
--- a/packages/cli/src/constructs/project.ts
+++ b/packages/cli/src/constructs/project.ts
@@ -16,7 +16,7 @@ export interface ProjectProps {
   /**
    * Git repository URL.
    */
-  repoUrl: string
+  repoUrl?: string
 }
 
 interface ProjectData {
@@ -28,7 +28,7 @@ interface ProjectData {
 
 export class Project extends Construct {
   name: string
-  repoUrl: string
+  repoUrl?: string
   logicalId: string
   data: ProjectData = {
     checks: {},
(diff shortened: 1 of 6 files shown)


clample:
> Resolves #[498]

I think that this could be `Resolves #498` in order to auto-link the GH issue and have it automatically close when the PR is merged.

packages/cli/src/constructs/project.ts
```diff
@@ -16,7 +16,7 @@ export interface ProjectProps {
   /**
    * Git repository URL.
    */
-  repoUrl: string
+  repoUrl?: string
```
clample, line 19: Makes sense ✅

umutuzgur:
Hmm, another weird error...

clample (approved):
Nice stuff 👍  These error messages are much nicer.

clample (approved):

packages/cli/src/services/checkly-config-loader.ts
```diff
@@ -74,6 +74,10 @@ function loadFile (file: string) {
   }
 }
 
+function isString (obj: any) {
+  return (Object.prototype.toString.call(obj) === '[object String]')
```
clample, line 78: Why doesn't `instanceof` work?

umutuzgur: Short answer, it is javascript. It doesn't consider const example = "string" as a member of the String class

PR #508, opened 2023-02-10
feat: check if resource was already added [sc-0]

### LINKED ISSUE #491: bug: 2 groups with the same locator ID doesn't throw an error but creates 1 group containing all the targeted checks
### Node.js version

v16.17.0

### NPM version

8.15.0

### @checkly/cli version

@checkly/cli/0.3.10 darwin-arm64 node-v16.17.0

### Steps to reproduce

1. Create 2 different groups targeting different browser checks using the `testMatch` property
2. Assign the same locator Id to the 2 groups
3. Deploy using `npx checkly deploy`

### What is expected?

It should throw an error & prevent the deployment. 

On `app.checklyhq.com`:
- only one of the 2 groups is created (most likely the one from the file run last during the deployment)
- all the targeted checks (from step 1) will be assigned to only 1 group
- there is no trace of the second group that should be created

### Any additional comments?

_No response_

nahuelon:

## Affected Components
* [x] CLI
* [ ] Create CLI
* [ ] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer
- throw error if a resource (`type` and `logicalId`) was previously added.

> Resolves #491


umutuzgur:
@nahuelon it looks good but please add a test for this

packages/cli/src/constructs/project.ts
```diff
@@ -58,6 +58,9 @@ export class Project extends Construct {
   }
 
   addResource (type: string, logicalId: string, resource: Construct) {
+    if (this.data[type as keyof ProjectData][logicalId]) {
+      throw new Error(`Resource of type '${type}' with logical id '${logicalId}' is duplicated.`)
```
umutuzgur, line 62: I would replace "is duplicated" to "already exists"

umutuzgur (approved):
LGTM

PR #511, opened 2023-02-10
feat: handle test command error [sc-0]

### LINKED ISSUE #497: bug: `npx checkly test` doesn't terminate when starting a check fails
### Node.js version

v16.17.0

### NPM version

8.15.0

### @checkly/cli version

@checkly/cli/0.3.14

### Steps to reproduce

Currently `npx checkly test` will hang indefinitely when a request to `POST /next/checks/runs/{checkType}` fails. This case can happen when an account is out of credit. For reproducing the issue, you can also simply add a `throw new Error('failed to start check')` here:

https://github.com/checkly/checkly-cli/blob/0130c022396fd8642cf870ca639a4d9bc823a938/packages/cli/src/services/check-runner.ts#L95

### What is expected?

`npx checkly test` should report any errors (for example, that the account is out of credit), then exit.

`npx checkly test` hangs.

### Any additional comments?

_No response_

nahuelon:

## Affected Components
* [x] CLI
* [ ] Create CLI
* [ ] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer
- fix `test` command hangs moving the timeout creation after getting the API request response. If the request fails, no timeout is created
- remove `this.emit(Events.CHECK_REGISTERED)` because no one is listening to it.

> Resolves #497

### Solution details (OLD PREVIOUS APPROACH)

The first issue found was resolved fixing the `CHECK_FAILED` event payload (replacing `checkRun` with 
`check`):

image

After that was fixed, the `test` command started to hang when one or more checks failed:

image

So, after debugging the `scheduleAllChecks()` I applied some changes to ensure all requests are sent to the API and get all Promises (`fulfilled` with a `Check` or `rejected` with an `Error`).

image


nahuelon:
@clample @umutuzgur , FYI, I used a `random` error to test this:

image

packages/cli/src/services/check-runner.ts
```diff
@@ -104,19 +104,19 @@ export default class CheckRunner extends EventEmitter {
...
-        this.emit(Events.CHECK_FINISHED, check)
-        // TODO: Find a way to abort. The latest version supports this but doesn't work with TS
-        return
+        this.emit(Events.CHECK_FAILED, check, new Error(errorMessage))
+      } else {
+        this.emit(Events.CHECK_FAILED, check, err)
       }
-      this.emit(Events.CHECK_FAILED, checkRun, err)
       this.emit(Events.CHECK_FINISHED, check)
+      throw new Error('Failed to run a check.')
```
umutuzgur, line 119: The catch block should handle the api issues. I actually don't get why we need to throw another error here

nahuelon: @umutuzgur , we weren't throwing the error when `status !== 402`. It's important to differentiate the `fulfilled` from the `rejected` promise.

packages/cli/src/commands/test.ts
```diff
@@ -179,7 +179,10 @@ export default class Test extends AuthCommand {
       })
       process.exitCode = 1
     })
-    runner.on(Events.RUN_FINISHED, () => reporter.onEnd())
+    runner.on(Events.RUN_FINISHED, () => {
+      reporter.onEnd()
+      this.exit()
```
clample, line 184: I don't think that we should need this. `await runner.run()` on the next line should already wait until all of the checks are finished. Is there a bug where `await runner.run()` doesn't actually terminate when all of the checks are finished? If so, then I think that we should address that.

clample:
@nahuelon could you please summarize what was causing the bug in the PR description, and how this fix addresses the bug? That would make the PR easier to review. It's hard to figure out from the diff alone.

packages/cli/src/services/check-runner.ts
clample: I'm also confused here. Since we catch the error, shouldn't the promise from `scheduleCheck()` always be fulfilled? I don't see how the promise would become stuck in a pending state.

nahuelon: @umutuzgur , @clample , I updated the PR comments explaining the solution. Let me know if you need more details.

clample: @nahuelon thanks for clarifying. Nice find with the `getSourceFile()` issue. 

I'm not sure what you and @umutuzgur think, but I think that it would still be worth looking into this bug further. The current PR seems to fix it, but I think it's worth making sure we understand 100% what happened. The main open questions for me are why the `runner.run()` promise isn't being fulfilled when all checks are finished, and why  `scheduleCheck` wasn't being fulfilled before.

Let me know if you'd like to pair on this. I'm happy to help take a look.

nahuelon:
@clample @umutuzgur , I found the reason for the command hangs. I used a package `wtfnode` to search for pending 'tasks' in the event loop. Here is the result:

image

So, I fixed the issue moving the timeout creation after the API request is valid. No timeout is created, if the request fails.

umutuzgur:
Wow, that is really interesting. Great investigation 👏

tnolet:
Damn. Nice one @nahuelon

clample:
Nice one 🎉

clample (approved):
Really nice find!

umutuzgur:
Nice catch!

MarineGEO circle logo

PR #513, opened 2023-02-10
fix: change test filter for spec file checks [sc-0]

### LINKED ISSUE #509: bug: running `test` with a filter for a file that is added to a group, it runs duplicate checks
### Node.js version

18.12.1

### NPM version

8.19.2

### @checkly/cli version

0.3.14

### Steps to reproduce

Here is a link to a setup that reproduces this. Run `npx checkly test one.spec.ts` and it will run that file, but also double it as it's created via glob in the group, and via an explicit `BrowserCheck` construct.

[https://codesandbox.io/p/sandbox/strange-bhabha-2gltvq?selection=%5B%7B%22endColumn%[…]A7%2C%22startColumn%22%3A19%2C%22startLineNumber%22%3A7%7D%5D](https://codesandbox.io/p/sandbox/strange-bhabha-2gltvq?selection=%5B%7B%22endColumn%22%3A19%2C%22endLineNumber%22%3A7%2C%22startColumn%22%3A19%2C%22startLineNumber%22%3A7%7D%5D)

### What is expected?

I expected only the file to run I specifically indicate.

It runs the same check, but in two "instances"

### Any additional comments?

This was reported by a customer

nahuelon:

## Affected Components
* [x] CLI
* [ ] Create CLI
* [ ] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer
- workaround to avoid duplicated checks when using `npx checkly test [FILEARGS]`. The only way I found to differentiate a spec file check from a `browserChecks: testMatch:` (or `code: entrypoint:`) was to use the `check.name`
- error handling wrong `RegEx` string arguments.

> Resolves #509


umutuzgur:
@clample could you check this one? I don't really know this filtering part very well

clample:
I have some comments on the original GH issue here - https://github.com/checkly/checkly-cli/issues/509#issuecomment-1427984795

clample:
By the way, @nahuelon could you please help me? I'm having difficulty understanding what the new behavior of `npx checkly test some-test.spec.js` would be. It's a bit tricky to tell for me to tell based on the current description and just reading the diff.

nahuelon:
Decided to keep the actual filtering behavior.

PR #514, opened 2023-02-10
feat: add testOnly flag for checks [sc-0]

### LINKED ISSUE #510: feat: add `testOnly` boolean to Check construct

Some checks are not meant to be deployed as monitors, but only run as tests. This is currently not super easy to do. In the end I should be able to use the CLI to cover both testing and monitoring use cases without stressing that I might deploy a check that was not meant for monitoring.

This was discussed internally and we opted for making the decision if a check is eligible for deployment at the check level, using a `testOnly` boolean.

nahuelon:

## Affected Components
* [x] CLI
* [ ] Create CLI
* [ ] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer
- add `testOnly` (default: `false`) in `Check` construct to discard the check on `deploy` command
- prepare `synthesizeRecord()` to easily add `testOnly` flag into other resources/construts, if we require this in the future
- include `testOnly` flag in docs

> Resolves #510

### DIFF
--- a/packages/cli/e2e/__tests__/deploy.spec.ts
+++ b/packages/cli/e2e/__tests__/deploy.spec.ts
@@ -50,4 +50,18 @@ describe('deploy', () => {
     })
     expect(result.stderr).toBe('')
   })
+
+  it('Shouldn\'t include a testOnly check', () => {
+    const result = runChecklyCli({
+      args: ['deploy', '--preview'],
+      apiKey: config.get('apiKey'),
+      accountId: config.get('accountId'),
+      directory: path.join(__dirname, 'fixtures/test-only-project'),
+      env: { PROJECT_LOGICAL_ID: projectLogicalId },
+    })
+    expect(result.stdout).toContain('not-testonly-default-check')
+    expect(result.stdout).toContain('not-testonly-false-check')
+    expect(result.stdout).not.toContain('testonly-true-check')
+    expect(result.status).toBe(0)
+  })
 })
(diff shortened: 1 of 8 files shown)


packages/cli/e2e/__tests__/deploy.spec.ts
```diff
@@ -50,4 +50,18 @@ describe('deploy', () => {
...
     expect(result.stderr).toBe('')
   })
+
+  it('Shouldn\'t include a testOnly check', () => {
+    const result = runChecklyCli({
+      args: ['deploy', '--preview'],
+      apiKey: config.get('apiKey'),
+      accountId: config.get('accountId'),
+      directory: path.join(__dirname, 'fixtures/test-only-project'),
+      env: { PROJECT_LOGICAL_ID: projectLogicalId },
```
clample, line 60: Nit: 

This PR currently doesn't use the `PROJECT_LOGICAL_ID` env var. It's there in the other test since it's read here: https://github.com/checkly/checkly-cli/blob/d102c1b043d878c1335c28b0b996680ae7bdd204/packages/cli/e2e/__tests__/fixtures/deploy-project/checkly.config.ts#L3

Currently the test doesn't do any cleanup of the project. The checks will just remain in the E2E account and keep running. Maybe it makes sense to add some cleanup similar to what we have with the other project with `cleanupProjects`:

https://github.com/checkly/checkly-cli/blob/d102c1b043d878c1335c28b0b996680ae7bdd204/packages/cli/e2e/__tests__/deploy.spec.ts#L41

On the other hand, adding some cleanup isn't super important. The test looks good.

clample (approved):

clample:
Nice changes and nice tests 👍 

Added a comment, but the PR is also OK as-is.

packages/cli/e2e/__tests__/deploy.spec.ts
nahuelon: @clample , thanks, I didn't realize about the `env.PROJECT_LOGICAL_ID`. Anyway, I'm using `deploy --preview` in the test.

PR #518, opened 2023-02-14
Tnolet comment copy tweaks

tnolet:

## Affected Components
* [ ] CLI
* [ ] Create CLI
* [ ] Test
* [x] Docs
* [ ] Examples
* [ ] Other

### DIFF
--- a/packages/cli/src/constructs/check-group.ts
+++ b/packages/cli/src/constructs/check-group.ts
@@ -63,7 +63,7 @@ export interface CheckGroupProps {
   concurrency?: number
   environmentVariables?: Array<EnvironmentVariable>
   /**
-   * List of alert channels this check should subscribe to.
+   * List of alert channels to be alerted when checks in this group fail or recover.
    */
   alertChannels?: Array<AlertChannel>
   browserChecks?: BrowserCheckConfig,
--- a/packages/cli/src/constructs/check-group.ts
+++ b/packages/cli/src/constructs/check-group.ts
@@ -23,7 +23,7 @@ const defaultApiCheckDefaults = {
 
 type BrowserCheckConfig = CheckConfigDefaults & {
   /**
-   * Glob pattern to include multiple files, i.e. all `.spec.js` files
+   * Glob pattern to include multiple files, i.e. all `.spec.ts` files
    */
   testMatch: string,
 }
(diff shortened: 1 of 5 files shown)


packages/cli/src/constructs/check-group.ts
```diff
@@ -63,7 +63,7 @@ export interface CheckGroupProps {
   concurrency?: number
   environmentVariables?: Array<EnvironmentVariable>
   /**
-   * List of alert channel subscriptions.
+   * List of alert channels this check should subscribe to.
```
clample, line 66: This is a check group rather than a check. Also I'm not sure if it makes sense to say that the group is subscribed to the alert channel, it's the other way around?

```suggestion
   * List of alert channels to be alerted when checks in this group fail or recover.
```

packages/cli/src/constructs/check.ts
```diff
@@ -65,11 +66,11 @@ export interface CheckProps {
   frequency?: number
   environmentVariables?: Array<EnvironmentVariable>
   /**
-   * The id of the check group this check is part of.
+   * The id of the check group this check is part of. Set this by calling `someGroup.ref()`
    */
   groupId?: Ref
   /**
-   * List of alert channel subscriptions.
+   * List of alert channels this check should subscribe to.
```
clample, line 73: Nit - It doesn't seem to me like checks are subscribed to alert channels. The alert channel gets notified when the check changes, rather than the other way around.

Maybe "List of alert channels to notify when the check fails or recovers"

clample, line 73: Nit - It doesn't seem to me like checks are subscribed to alert channels. The alert channel gets notified when the check changes, rather than the other way around.

What about:
```suggestion
   * List of alert channels to notify when the check fails or recovers.
```

packages/cli/src/services/checkly-config-loader.ts
```diff
@@ -27,7 +27,7 @@ export type ChecklyConfig = {
    */
   checks?: CheckConfigDefaults & {
     /**
-     * Glob pattern to where Checkly looks for checks.
+     * Glob pattern where the CLI looks for file containing Check constructs, i.e. all `.checks.ts` files
```
clample, line 30:
```suggestion
     * Glob pattern where the CLI looks for files containing Check constructs, i.e. all `.checks.ts` files
```

tnolet:
@clample good suggestions, I committed them.

nahuelon (approved):

PR #531, opened 2023-02-15
feat: detect construct usage in spec file [sc-0]

### LINKED ISSUE #387: RFC: Detect spec<->check file mixups and return a clear error message
Users might attempt to create constructs in browser check files. Currently this fails with an error about `@checkly/cli/constructs` not being supported in the runtime:
image

### Proposal
We can create a more tailored error message for this case which describes the difference between check files and spec/test files. This can let users find a solution more easily.

The exact copy might need some more thought, but maybe something like:
> It looks like you're trying to use `@checkly/cli/constructs` in a browser check file. `@checkly/cli/constructs` should only be used in check files: files ending in `.check.ts` and `.check.js`, or the `checkMatch` pattern set in your configuration. For more information see on the difference between test files and check files, see link-to-docs.

nahuelon:

## Affected Components
* [x] CLI
* [ ] Create CLI
* [ ] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer
- detect CLI construct usage in spec file and returns a detailed error message.

> Resolves #387

### DIFF
--- a/packages/cli/src/services/check-parser/errors.ts
+++ b/packages/cli/src/services/check-parser/errors.ts
@@ -19,12 +19,19 @@ export class DependencyParseError extends Error {
       }
     }
     if (unsupportedNpmDependencies.length) {
-      message += '\n\nThe following NPM dependencies were used, but aren\'t supported in the runtimes.\n'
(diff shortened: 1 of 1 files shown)


umutuzgur (approved):
LGTM

PR #540, opened 2023-02-20
fix: handle Windows path separator for logical IDs

### LINKED ISSUE #539: bug: invalid logical ID's created on Windows
### Node.js version

-

### NPM version

-

### @checkly/cli version

0.4.0

### Steps to reproduce

On windows, have a browser check created from a `testMatch` file.

### What is expected?

The browser check should be created when running `npx checkly deploy`.

There's a validation error related to the logical ID:
```
  Error: The 'logicalId' must includes only allowed characters [A-Za-z0-9_-/#.]. (logicalId='src\__checks__\check.spec.ts')
```

### Any additional comments?

_No response_

clample:

## Affected Components
* [x] CLI
* [ ] Create CLI
* [ ] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer

> Resolves #539 

On Windows, the path separator is `\` rather than `/`. The relative path of `testMatch` files is used in the logical ID. This causes problems, though, since `\` isn't an allowed character in logical IDs. Users on Windows will receive the following error:
```
  Error: The 'logicalId' must includes only allowed characters [A-Za-z0-9_-/#.]. (logicalId='src\__checks__\check.spec.ts')
```

Additionally, logical ID's should be consistent across platforms. If `npx checkly deploy` produces different logical IDs in Linux CI vs a Windows developer laptop, then users would constantly see checks being deleted and recreated with each `npx checkly deploy`.

This PR resolves the issue by normalizing the logical IDs for `testMatch` files.

#### Test Plan

I've added a simple unit test to check that this works. 

I've also checked that this doesn't cause any breaking changes in logical ID's on macOS / Linux by running `npx checkly deploy` first with CLI version 0.4.0, then with this version. I've confirmed that the checks aren't recreated.

### DIFF
--- a/packages/cli/src/constructs/check-group.ts
+++ b/packages/cli/src/constructs/check-group.ts
@@ -8,6 +8,7 @@ import { AlertChannel } from './alert-channel'
 import { EnvironmentVariable } from './environment-variable'
 import { AlertChannelSubscription } from './alert-channel-subscription'
 import { CheckConfigDefaults } from '../services/checkly-config-loader'
+import { pathToLogicalId } from '../services/util'
 import type { Region } from '..'
 
 // TODO: turn this into type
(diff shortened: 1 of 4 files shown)


umutuzgur (approved):
LGTM

Issue #527, opened 2023-02-15
feat: show fine-grained detail of changes in `npx checkly deploy --preview`


clample:
### What problem does this feature solve?

This issue relates to https://github.com/checkly/checkly-cli/issues/526, but I think that it's a distinct issue and a bit harder to fix.

Currently `npx checkly deploy --preview` only shows high-level information about the changes: which checks/resources were created, updated, and deleted. It currently doesn't show _which_ fields are changed. For example, it won't show that a check goes from `activated: true` to `activated: false`. It also doesn't detect the case where a check isn't changed at-all - the check will still show as updated in the diff.

It would be great if `npx checkly deploy --preview` gave more fine-grained detail:

```
$ npx checkly deploy --preview

Deleted:
    (Check) 'My Check'
    
Updated:
    (Check) 'Other Check'
        + activated: true ~> false
        + frequency: 10 ~> 15
        
Unchanged:
    (Check) 'another check'
```

### How would you implement this feature?

Maybe [jest-diff](https://jestjs.io/docs/jest-platform#jest-diff) is useful. We can also look at `terraform plan` for inspiration on how the output should be formated.

nahuelon:
@clample , I was trying a package `deep-object-diff` to get the `added`, `updated` and `deleted` dif. I'll continue working in the `/next/deploy` endpoint, adding a detailed JSON we could manipulate in the CLI to display user-friendly response. Maybe it's related with the issue #526 you are working on.

clample:
Thanks for checking this out @nahuelon. `deep-object-diff` looks useful.

The two issues are really closely related. We could maybe even tackle them both at once and just completely rework the `--preview`.

I created two issues since I want to have the `--preview` output cleaned up in a human readable form before the launch week. I'm not sure that we'll be able to finish all of the changes to `/next/deploy` in time, though, so I created https://github.com/checkly/checkly-cli/issues/526 to clean up the output just using the data that we already have.

tnolet:
@clample can we close this one?

clample:
@tnolet personally I think that it would be a useful feature. It would take a day or two to do, though.

Please feel free to close it if you don't think that it's worth doing.

tnolet:
@clample I mixed this up with the other ticket. We can leave this open. I love the output suggested here.

nahuelon:
[...]

PR #542, opened 2023-02-21
feat: add improved output for checkly deploy --preview

### LINKED ISSUE #526: feat: polish output from `npx checkly deploy --preview`

The current output for `npx checkly deploy --preview` is just the raw response from the backend:
```
$ npx checkly deploy --preview
{
  diff: {
    checks: {
      'public-resource_/public-stats': 'UPDATE',
      'public-resource_/v1/runtimes': 'UPDATE',
      '__checks__/homepage.spec.ts': 'UPDATE'
    }
  }
}
```

This is clear enough and users will likely understand it, but it's not consistent with the human readable output used in the rest of the CLI. We should consider printing this with different formatting and maybe use colored output (red for deletion, green for creation, etc). I'm not sure what exactly the best format would be, but maybe something like:

```
Deleting the following resources:
   (Check) 'My Browser Check'
   (Check Group) 'Production Check Group'
   
Creating the following resources:
    (Check) 'New Browser Check' my-project/__checks__/new.check.ts
```

clample:

## Affected Components
* [x] CLI
* [ ] Create CLI
* [ ] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer

> Resolves #526 

Currently `npx checkly deploy --preview` simply prints raw data from the backend:
Screenshot 2023-02-21 at 11 40 25

This output only includes the logical ID and the resource type, but it would also be nice to have information like the file that the object was created in. In other commands, the CLI also gives human readable output by default. It would be nice to make `--preview` also give a human readable output, for consistency.

I think that there's a lot of room to improve the `--preview` output, but it will require changes to the backend and significant work (I created a separate ticket for this: https://github.com/checkly/checkly-cli/issues/527). This PR simply makes some minor polishing to the `--preview` output - using only information already being returned by the backend. I'm not sure that it's worth investing too much effort into perfecting this PR - since I think it will be worth doing a major overhaul later on anyway.

Screenshot 2023-02-21 at 12 18 44

#### Notes
* Since `UPDATE` also includes resources that didn't change, I think that it's the least interesting category. I put `CREATED` and `DELETED` at the top.
* For `DELETED` resources, we only know the logical ID and resource type. We can't log any extra information for these.
* The resources are sorted to give the output a bit more structure.
* We already keep track of the source file for `Check` constructs. This PR tracks this information for _all_ constructs. It also makes sure to account for constructs created in the checkly config file.

#### Testing
I simply ran `npx checkly deploy --preview` locally.

### DIFF
--- a/packages/cli/src/commands/deploy.ts
+++ b/packages/cli/src/commands/deploy.ts
@@ -7,6 +7,10 @@ import { parseProject } from '../services/project-parser'
 import { loadChecklyConfig } from '../services/checkly-config-loader'
 import { runtimes } from '../rest/api'
 import type { Runtime } from '../rest/runtimes'
+import { AlertChannelSubscription, CheckGroup, Project, ProjectData } from '../constructs'
+import chalk = require('chalk')
+import { Check } from '../constructs/check'
+import { AlertChannel } from '../constructs/alert-channel'
 
 export default class Deploy extends AuthCommand {
   static hidden = false
@@ -70,7 +74,7 @@ export default class Deploy extends AuthCommand {
       const projectPayload = project.synthesize(false)
       const { data } = await api.projects.deploy(projectPayload, { dryRun: preview })
       if (preview || output) {
-        this.log(data)
+        this.log(this.formatPreview(data, project))
       }
       if (!preview) {
         this.log(`Successfully deployed project "${project.name}" to account "${account.name}".`)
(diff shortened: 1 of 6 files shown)


packages/cli/src/commands/deploy.ts
```diff
@@ -83,4 +87,72 @@ export default class Deploy extends AuthCommand {
...
+      output.push(chalk.bold.green('Create:'))
+      for (const { logicalId, construct } of creating) {
+        const name = (construct as any).name
+        output.push(`    ${construct.constructor.name}: ${construct.getSourceFile()} > ${name ?? logicalId}`)
+      }
+      output.push('')
+    }
+    if (deleting.length) {
+      output.push(chalk.bold.red('Delete:'))
+      const prettyResourceTypes: any = {
```
umutuzgur, line 138: Why any? Record<string, string>?

packages/cli/src/commands/deploy.ts
```diff
@@ -83,4 +87,72 @@ export default class Deploy extends AuthCommand {
...
+    const creating = []
+    const deleting = []
+    for (const [resourceType, resourceStatuses] of Object.entries(previewData?.diff ?? {})) {
+      if (resourceType === AlertChannelSubscription.__checklyType) {
+        // Don't report changes to alert channel subscriptions.
+        // User's don't create these directly, so it's more intuitive to consider it as part of the check.
+        continue
+      }
+      for (const [logicalId, resourceStatus] of Object.entries(resourceStatuses ?? {})) {
+        if (resourceStatus === 'UPDATE') {
```
umutuzgur, line 105: Maybe do an enum of these?

packages/cli/src/commands/deploy.ts
clample: Without the `any`, then there's a TS compile error on line 144 - `resourceType` doesn't have type `keyof prettyResourceTypes`.

I can update it to `Record<string, string>`, that should also solve it.

umutuzgur: Please avoid using `any` as much as you can

packages/cli/src/constructs/construct.ts
```diff
@@ -4,9 +4,11 @@ import { Ref } from './ref'
 export abstract class Construct {
   type: string
   logicalId: string
+  __checkFilePath?: string
   constructor (type: string, logicalId: string) {
     this.logicalId = logicalId
     this.type = type
+    this.__checkFilePath = Session.checkFilePath
```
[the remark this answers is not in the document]
clample: Hmm, good point.

`main` is currently using this for `checkly test` and only assigning it for `Check` constructs. I think that it works in `main`, and we don't really need to change it. I think that most users will just define the checks directly in the  check file and there won't be many edge cases. I previously looked into getting the actual file where the construct is initialized and didn't see a nice solution.

Adding this to `CheckGroup` and `AlertChannel` like I do in this PR might be problematic, though. I think that it's much more likely these will be defined in a separate file and imported into multiple check files. 

I think the easiest solution is to just remove the source file from the `--preview` output. I can go ahead and change all of that back so that the output is:

```
$ npx checkly deploy --preview

Create:
    BrowserCheck: my-browser-check-logical-id
    BrowserCheck: another-browser-check

....
```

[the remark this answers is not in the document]
clample: I've updated the PR to remove all of the source file handling:

Screenshot 2023-02-21 at 16 12 17

umutuzgur (approved):
LGTM

PR #546, opened 2023-02-22
fix: log clear error in case of internet connection issues

### LINKED ISSUE #545: bug: unclear message in API error edge case
### Node.js version

-

### NPM version

-

### @checkly/cli version

0.4.2

### Steps to reproduce

Run `npx checkly test` without an internet connection.

### What is expected?

There is a clear error message. For example:

> Unable to connect to Checkly services. Please check the internet connection.

```
$ npx checkly test
    TypeError: Cannot destructure property 'status' of 'err.response' as it is undefined.
```

### Any additional comments?

This is probably coming from here:

https://github.com/checkly/checkly-cli/blob/f8ac506dcee73d8242ca7ba9e11532f601045ac2/packages/cli/src/rest/api.ts#L54

We can't assume that `err.response` is defined - there might not have been a response from the API.

clample:

## Affected Components
* [x] CLI
* [ ] Create CLI
* [ ] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer

> Resolves #545 

When there's no internet connection, the CLI fails with an unclear error. This can be reproduced by disabling WIFI.
```
$ npx checkly test
    TypeError: Cannot destructure property 'status' of 'err.response' as it is undefined.
```

This PR detects this case by checking whether `err.response` is defined, and returns a more specific error message. The error messages returned are now: 

Incorrect API key (unchanged):
```
$ npx checkly test
    Error: Authentication failed with Account ID "bf7babe3-0589-4cc9-9358-f706b3b11362"
    and API key "...8b0f"
```

Wifi Disabled:
```
$ npx checkly test
    Error: Encountered an error connecting to Checkly. Please check that the internet
    conenction is working.
    Error: getaddrinfo ENOTFOUND api.checklyhq.com
```

Unexpected error (for example, a 404 from Checkly):
```
$ npx checkly test
    Error: Encountered an unexpected error connecting to Checkly: Request failed with
    status code 404
```

Implementing this, I referred to [Axios error handling docs](https://axios-http.com/docs/handling_errors)

I also made some small changes to `isAuthenticated`. The boolean return value is never used, so I changed it to `validateAuthentication`. It throws an error if there's an issue, otherwise it returns.

### DIFF
--- a/packages/cli/src/commands/authCommand.ts
+++ b/packages/cli/src/commands/authCommand.ts
@@ -6,6 +6,6 @@ export abstract class AuthCommand extends BaseCommand {
 
   protected async init (): Promise<any> {
     super.init()
-    await api.isAuthenticated()
+    await api.validateAuthentication()
   }
 }
--- a/packages/cli/src/commands/login.ts
+++ b/packages/cli/src/commands/login.ts
@@ -73,7 +73,7 @@ export default class Login extends BaseCommand {
   }
 
   private _isLoginSuccess = async () => {
-    await api.isAuthenticated()
+    await api.validateAuthentication()
     this.log('Welcome to @checkly/cli 🦝')
   }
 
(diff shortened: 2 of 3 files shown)


umutuzgur (approved):
LGTM

PR #547, opened 2023-02-22
feat: allow setting check group without calling ref()

clample:

## Affected Components
* [x] CLI
* [ ] Create CLI
* [ ] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer
The `Check` construct already accepts `alertChannels` as actual `AlertChannel` constructs, but for `groupId` it accepts a `Ref` rather than a `CheckGroup`. This is a bit inconsistent, and it would be nice to follow one pattern for linking constructs. 

https://github.com/checkly/checkly-cli/blob/d1a1bed14d68fba64103b916bc8f90fedbec37d3/packages/cli/src/constructs/check.ts#L62-L65

This PR updates the preferred method of assigning a check to a group from passing a ref:
```
new BrowserCheck('logicalId', { groupId: group.ref() })
```

to passing the group construct:
```
new BrowserCheck('logicalId', { group })
```

This is done in a backwards compatible way. The old `groupId` property is marked as deprecated.
Screenshot 2023-02-22 at 15 34 03

This PR also updates the examples since it's just one line. To avoid a race condition, a release should be made right after this is merged. Docs are updated in https://github.com/checkly/checklyhq.com/pull/683.

### DIFF
--- a/examples/advanced-project/src/__checks__/group.check.ts
+++ b/examples/advanced-project/src/__checks__/group.check.ts
@@ -3,7 +3,7 @@ import { smsChannel, emailChannel } from '../alert-channels'
 const alertChannels = [smsChannel, emailChannel]
 /*
 * In this example, we bundle checks using a Check Group. We add checks to this group in two ways:
-* 1. By calling the ref() method for the groupId property of the check.
+* 1. By passing the `CheckGroup` object for the `group` property of the check.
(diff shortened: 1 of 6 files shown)


umutuzgur (approved):
LGTM

PR #550, opened 2023-02-23
fix: use correct base path for CheckGroup testMatch

### LINKED ISSUE #549: bug: __checkFilePath points to the wrong path when checkly resources are imported from different files
### Node.js version

16

### NPM version

8

### @checkly/cli version

0.4.3

### Steps to reproduce

Let's say you have your checkly resources in file called `resources.js|ts` and these are imported by a check file, group.check.js, the relative path entrypoint calculation will be wrong in this case. `__checkFilePath` will be set as `group.check.js` and the `entrypoint` relative path functionality will break

### What is expected?

That we have a relative path built according to the file which created the resources

Explained in the steps to reproduce

### Any additional comments?

_No response_

clample:

## Affected Components
* [x] CLI
* [ ] Create CLI
* [ ] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer

> Resolves #549

`CheckGroup` supports loading `*.spec.ts` checks using `testMatch`, and `BrowserCheck` supports relative paths for `*.spec.ts` files passed in `code: { entrypoint }`. Internally, these relative files are loaded using the `Session.checkFileAbsolutePath` as the base path. As mentioned in the linked GH issue - this doesn't always give the expected behavior. In particular, if the constructs are instantiated in a helper file then imported in a `*.check.ts` file - then the extra files will be loaded relative to the `*.check.ts` file. For a more clear illustration, see the test in this PR (which fails for `main`).

It's hard to detect the file where the `CheckGroup` constructor is actually called. One possible solution is to require the user to pass the base path explicitly:
```
new CheckGroup('my-group', { basePath: __filename })
```

This adds extra boilerplate for users, though. Instead, this PR uses a library [get-caller-file](https://github.com/stefanpenner/get-caller-file). This library parses the file name of the caller from the `Error` stack trace. This is a bit of a hack, but it works fine and the library is widely used. This let's us correctly detect which file the `CheckGroup` is actually instantiated in.

This will be a bit of a breaking change in case any users were importing `CheckGroup`'s and `Check`'s to other files. I think that it's unlikely that users are doing this, though, so I think it's OK to merge this. Maybe we should consider making the next release `0.5.x`, though.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5995,7 +5995,6 @@
       "version": "2.0.5",
       "resolved": "https://registry.npmjs.org/get-caller-file/-/get-caller-file-2.0.5.tgz",
       "integrity": "sha512-DyFP3BM/3YHTQOCUL/w0OZHR0lpKeGrxotcHWcqNEdnltqFwXVfhEBQ94eIo34AfQpo0rGki4cyIiftY06h2Fg==",
-      "dev": true,
       "engines": {
         "node": "6.* || 8.* || >= 10.*"
       }
@@ -11749,6 +11748,7 @@
         "ci-info": "3.7.1",
         "conf": "10.2.0",
         "dotenv": "16.0.3",
+        "get-caller-file": "2.0.5",
         "glob": "8.0.3",
         "indent-string": "4.0.0",
         "inquirer": "8.2.3",
@@ -12376,6 +12376,7 @@
         "config": "3.3.8",
         "dotenv": "16.0.3",
         "eslint": "8.30.0",
+        "get-caller-file": "2.0.5",
         "glob": "8.0.3",
         "indent-string": "4.0.0",
         "inquirer": "8.2.3",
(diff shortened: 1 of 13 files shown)


umutuzgur:
@clample this also works for `ts`, right? I think it does but wasn't 100% sure

clample:
@umutuzgur correct - it should work for TS. The integration test added here is testing the TS case.

umutuzgur:
What about the `Session.checkFilePath` btw? Is it still needed?

clample:
> What about the Session.checkFilePath btw? Is it still needed?

Good question. It's currently being used for `npx checkly test` for filtering (f.ex `npx checkly test foo.check.ts`) and for organizing the results. I think that we can definitely migrate this to use `get-caller-file` instead.

What do you think? Do you have a preference whether I do it in this PR, or do it in a follow up PR?

clample:
Hmmm, @umutuzgur I think that I found another issue with this approach.

Suppose that a user is making constructs with a helper file:

`check-factory.ts`
```
export function makeCheck(endpoint) {
    return new ApiCheck(`check-${endpoint}`, {
        ....
    })
}
```

`login.check.ts`
```
makeCheck('/login')
```

`signup.check.ts`
```
makeCheck('/signup')
```


In this case, the `getCallerFile` approach will report that all of the checks are coming from `check-factory.ts`. I think that this behavior might end up seeming a bit strange as well.

clample:
Closing this PR for now, I think that this needs a bit more thinking on what the solution should be.

umutuzgur:
@clample We can also just document the corner cases to address the issue. I don't think we will find a solution that solves every case

PR #553, opened 2023-02-24
feat: log error message when no resources in deploy preview

clample:

## Affected Components
* [x] CLI
* [ ] Create CLI
* [ ] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer

Currently `npx checkly deploy --preview` has no output at all when no checks or other resources are detected. This output could be a bit misleading for new users.

This PR adds some output for this case:
```
$ npx checkly deploy --preview

No checks were detected. More information on how to set up a Checkly CLI project is available at https://checklyhq.com/docs/cli/.

```

This issue could happen for a number of reasons, so I'm not sure if it makes sense to give any other info in the output for now.

### DIFF
--- a/packages/cli/src/commands/deploy.ts
+++ b/packages/cli/src/commands/deploy.ts
@@ -155,6 +155,9 @@ export default class Deploy extends AuthCommand {
       }
       output.push('')
     }
+    if (!creating.length && !deleting.length && !updating.length) {
+      output.push('\nNo checks were detected. More information on how to set up a Checkly CLI project is available at https://checklyhq.com/docs/cli/.\n')
+    }
     return output.join('\n')
   }
 }
--- a/packages/cli/src/commands/deploy.ts
+++ b/packages/cli/src/commands/deploy.ts
@@ -120,6 +120,11 @@ export default class Deploy extends AuthCommand {
         }
       }
     }
+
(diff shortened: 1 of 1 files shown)


tnolet:
@clample we need to add more of these messages on different spots.

- no checks configured in otherwise correct setup
- no config at all, but user runs `npx checkly test` (when installed globally) in a project.

See also https://checklyhq.slack.com/archives/C04PFSV5W3B/p1677003160430209

clample:
@tnolet having a `messages.ts` makes sense, good idea 👍

clample:
I think that it could still be nice to merge this one in the meantime

tnolet:
@clample yes please

packages/cli/src/commands/deploy.ts
```diff
@@ -155,6 +155,9 @@ export default class Deploy extends AuthCommand {
       }
       output.push('')
     }
+    if (!creating.length && !deleting.length && !updating.length) {
```
nahuelon, line 158: I think you could move this `if` just after line `131` and, instead of `output.push(<message>)` just `return <message>`.

clample: Good point, that makes sense

nahuelon (approved):
LGTM

PR #563, opened 2023-03-02
feat: add git info into project [sc-14207]

nahuelon:

## Affected Components
* [x] CLI
* [ ] Create CLI
* [ ] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer
- gather git repository information using `git-repo-info` package
- (to be discussed) get the git remote origin URL using the `node:child_process.spawnSync` command
- add a new `project` property (`repoInfo`) to be sent when `test` and `deploy` command

**Important**: https://github.com/checkly/checkly-backend/pull/3924 must be merged first.

> Resolves [[sc-14207](https://app.shortcut.com/checkly/story/14207/as-a-user-i-want-the-cli-to-gather-data-about-the-ci-git-info)]

## New Dependency Submission
- Add `git-repo-info` to get Git repository information

### DIFF
--- a/packages/cli/src/commands/test.ts
+++ b/packages/cli/src/commands/test.ts
@@ -14,6 +14,7 @@ import type { Runtime } from '../rest/runtimes'
 import { AuthCommand } from './authCommand'
 import { BrowserCheck } from '../constructs'
 import type { Region } from '..'
+import { getGitInformation } from '../services/util'
 
 const DEFAULT_REGION = 'eu-central-1'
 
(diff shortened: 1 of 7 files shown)


packages/cli/src/services/util.ts
```diff
@@ -1,6 +1,14 @@
 import * as path from 'path'
 import * as fs from 'fs/promises'
 import { Service } from 'ts-node'
+import * as gitRepoInfo from 'git-repo-info'
+
+export interface GitInformation {
+  commitId: string
```
umutuzgur, line 7: So we don't want the originUrl for now?

nahuelon: @umutuzgur , IMO, we can move on with the `repoUrl` from the project config. Then, we could adjust it to fetch the origin URL and sanitize it.

packages/cli/src/commands/test.ts
```diff
@@ -117,11 +117,14 @@ export default class Test extends AuthCommand {
     })
     const verbose = this.prepareVerboseFlag(verboseFlag, checklyConfig.cli?.verbose)
     const { data: availableRuntimes } = await api.runtimes.getAll()
+    const gitInfo = getGitInformation()
+    console.info(gitInfo)
```
umutuzgur, line 121: Don't forget to remove this

packages/cli/src/services/util.ts
nahuelon: Also, the only way to gather the origin URL, is to run a child process and execute the git config .... I found some NPM packages, but they are doing this under the hood.

umutuzgur: We shouldn't have to. It should be in the config file at `.git/config`

umutuzgur (approved):
LGTM

PR #571, opened 2023-03-03
fix: parse file with extension other than .js or .ts [sc-0]

### LINKED ISSUE #568: bug: TypeScript import fails when using `.x.ts` suffix
### Node.js version

v16.17.0

### NPM version

8.15.0

### @checkly/cli version

v0.4.5

### Steps to reproduce

Create a browser-check `.spec.ts` and a file with name `./<someName>.pages.ts`. Try to import the file using `import {} from './<someName>.pages'`. Run `npx checkly test`.

### What is expected?

All checks to be loaded and run successfully.

The dependency file is not found:

```
The following dependencies weren't found:
        /<somePath>/<someName>.pages
``` 

### Any additional comments?

Slack thread: https://checklycommunity.slack.com/archives/C04349E26EN/p1677757653932529

nahuelon:

## Affected Components
* [x] CLI
* [ ] Create CLI
* [x] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer
- fix the parser when the file path has an extension other than `.js` or `.ts`
- add test with a check `.spec.ts` files importing file with extensions other than `.js` or `.ts`
- fix issue with duplicated resources error when using file patterns in Windows. `pathToLogicalId()` is renamed to `pathToPosix()` and reused in all `path` related operations.

> Resolves #568


nahuelon:
I'm checking why the test is failing with `npm run test:e2e` but works fine with `npx checkly test`.

packages/cli/src/services/project-parser.ts
```diff
@@ -72,7 +72,7 @@ async function loadAllCheckFiles (
   for (const checkFile of checkFiles) {
     // setting the checkFilePath is used for filtering by file name on the command line
     Session.checkFileAbsolutePath = checkFile
-    Session.checkFilePath = path.relative(directory, checkFile)
+    Session.checkFilePath = pathToPosix(path.relative(directory, checkFile))
```
clample, line 75: Do we need `pathToPosix` here (and on line 75 above)? Do we have a bug or is this just to refactor the code?

I'm wondering if in some places we should just keep the platform dependent path. In the case of `checkFilePath` I think it's mainly used for filtering with `npx checkly  test <some-path-filter>`. Maybe it's more intuitive for Windows users to be able to still use the Windows path separator, though: `npx checkly test \path\to\my_test.spec.js`.

packages/cli/src/services/check-parser/parser.ts
```diff
@@ -144,7 +144,8 @@ export class Parser {
 
   static readDependency (filePath: string, preferedExtenstion: SupportedFileExtension) {
     // Read the specific file if it has an extension
-    if (path.extname(filePath).length) {
+    const fileExtension = path.extname(filePath)
```
umutuzgur, line 147: I think we gotta add the extension to the `tryReadFileExt` instead. So if fileExtention exists, we do `Parser.tryReadFileExt(filePath, [fileExtention, ...JS_RESOLVE_ORDER])`. This is the same for TS resolve

packages/cli/e2e/__tests__/fixtures/test-project/src/__checks__/homepage.test.ts
```diff
@@ -0,0 +1,16 @@
+// @ts-ignore
```
clample, line 1: Nit:

Would it be possible to move this test to be a unit test in `check-parser.spec.ts` as a separate unit test case and explicitly asserting that `external.first.page.js` and `external.second.page` are correctly picked up as dependencies?

The e2e tests are a bit heavier weight, so I don't typically run them locally. Having this covered as a unit test would make it a bit easier to run. I think that having the explicit assertion will also help make sure that we don't accidentally break the test.

packages/cli/src/services/project-parser.ts
nahuelon: @clample , I convert all paths because I read the docs for `glob` and they suggest to always use the Unix path separator, even on Windows:
image

clample: @nahuelon interesting catch 🤔 

 I think that `npx checkly  test <some-path-filter>` doesn't actually go through the `glob` package. Instead, it uses a regex (code snippets below). It might still make sense to switch to `/`, though, like you mention.
 
 Either way, if we don't really need this to fix https://github.com/checkly/checkly-cli/issues/568, I propose just moving the `pathToPosix` change out of this PR. It would be easier to discuss what the behavior should be with a separate Slack thread or PR, and we could just merge the `.page.js` fix right away.
 
https://github.com/checkly/checkly-cli/blob/60a3464d0d73566b464249377c231adf099dedbd/packages/cli/src/commands/test.ts#L126-L128

https://github.com/checkly/checkly-cli/blob/60a3464d0d73566b464249377c231adf099dedbd/packages/cli/src/services/test-filters.ts#L6-L10

packages/cli/src/services/check-parser/parser.ts
nahuelon: Mmmm, I'm not sure. Because in that case, we should explicitly add all the 'compatible extensions' to the list. In this specific case, it would be the `.page`, but maybe there is some other user using `.resource`, `.site` or any other option.
But, as you say, adding the `.page` extension into the `JS_RESOLVE_ORDER` and `TS_RESOLVE_ORDER` would fix the issue too.

packages/cli/e2e/__tests__/fixtures/test-project/src/__checks__/homepage.test.ts
nahuelon: @clample , IMO, I think it's better to keep it in the E2E. Having a `test-project` with more complexity helps us to identify errors during the `test:e2e`. The previous `test-project` wasn't covering the 'missing extension' issue neither the 'duplicated resource' in Windows.
However, let me add a unit test to cover the file without extension dependencies pick up.

clample: Thanks @nahuelon, that sounds good to me

nahuelon: @clample , done.

PR #573, opened 2023-03-06
fix: transform all path to POSIX [sc-0]

### LINKED ISSUE #572: bug: get
### Node.js version

v16.16.0

### NPM version

8.11.0

### @checkly/cli version

v0.4.5

### Steps to reproduce

On Windows system, use `browserChecks.testMatch` glob to fetch the same `.spec.ts` file from two different places (i.e. a `CheckGroup` and the Checkly config file). Run `npx checkly test`.

### What is expected?

The checks to run successfully.

`Error: Resource of type 'checks' with logical id 'src/__checks__/<somefile>.spec.ts' already exists.`

image

### Any additional comments?

The problem happens only in Windows and the reason is because the `checkFilePath` is not a POSIX path. The `preexistingCheckFiles.has(relPath)` always returns `false` and the parser doesn't know it was already included.

nahuelon:

## Affected Components
* [x] CLI
* [ ] Create CLI
* [x] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer
- fix duplicated `logicalId` error converting all path to POSIX. This fixes the problem checking if the check was already added with `preexistingCheckFiles.has(relPath)`.
- add checks into the `test-project` for the E2E test.
- `checkFileAbsolutePath` is not modified/covered by this PR.

> Resolves #572

### DIFF
--- a/packages/cli/e2e/__tests__/fixtures/test-project/src/__checks__/group.check.ts
+++ b/packages/cli/e2e/__tests__/fixtures/test-project/src/__checks__/group.check.ts
@@ -13,6 +13,10 @@ const group = new CheckGroup('check-group-1', {
   apiCheckDefaults: {},
   concurrency: 100,
   alertChannels,
+  browserChecks: {
+    // using .test.ts suffix (no .spec.ts) to avoid '@playwright/test not found error' when Jest transpile the spec.ts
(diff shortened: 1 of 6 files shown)


umutuzgur (approved):
LGTM

PR #580, opened 2023-03-13
feat: add config flag for test and deploy [gh-577]

### LINKED ISSUE #577: story: target a specific configuration with `-c` / `--config` flag
### Story description

We want to allow users to manage different target for test, deployment from one code base. The cleanest solution is to let users define different `checkly.config.ts` files that have different settings. This creates a composable way to:

- target different environments like staging, production etc.
- enables even cross account, cross project workflows by appending different credentials to a `npx checkly <command> -c <config>` invocation

## `-c` / `--config` flag

- Flag applies to `test` and `deploy` commands.
- When no flag is set, defaults to `checkly.config.ts` and then `checkly.config.js` in root directory

### Shortcut link

_No response_

### Additional resources

_No response_

nahuelon:

## Affected Components
* [x] CLI
* [ ] Create CLI
* [x] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer
- add `-c` / `--config` flag for `test`, `deploy` and `destroy` commands allowing to specify the configuration filename.
- add unit tests cases for `loadChecklyConfig()` function
- add E2E test cases pointing to different configuration files for the same project
- add new function to split and get directory and filename from the `--config` flag argument.
- adjust `pathToPosix()` to remove the `C:` prefix in Windows absolute path (TODO: we should handle other drive letters too)

> Resolves #577


packages/cli/src/services/checkly-config-loader.ts
```diff
@@ -79,10 +79,12 @@ function isString (obj: any) {
   return (Object.prototype.toString.call(obj) === '[object String]')
 }
 
-export async function loadChecklyConfig (dir: string, filenames = ['checkly.config.ts', 'checkly.config.js']): Promise<{ config: ChecklyConfig, constructs: Construct[] }> {
+export async function loadChecklyConfig (dir: string, filename = ''):
```
clample, line 82:
```suggestion
export async function loadChecklyConfig (dir: string, filename?: string):
```

`filename = ''` also works, so feel free to leave as-is. I personally prefer `filename?: string`, though, since we want the `filename` parameter to be optional.

packages/cli/src/commands/deploy.ts
```diff
@@ -39,13 +39,18 @@ export default class Deploy extends AuthCommand {
...
+    }),
   }
 
   async run (): Promise<void> {
     const { flags } = await this.parse(Deploy)
-    const { force, preview, output } = flags
+    const { force, preview, output, config: configFilename } = flags
     const cwd = process.cwd()
-    const { config: checklyConfig, constructs: checklyConfigConstructs } = await loadChecklyConfig(cwd)
+    const { config: checklyConfig, constructs: checklyConfigConstructs } = await loadChecklyConfig(cwd, configFilename!)
```
clample, line 53:
```suggestion
    const { config: checklyConfig, constructs: checklyConfigConstructs } = await loadChecklyConfig(cwd, configFilename)
```

We can drop the `!` since `loadChecklyConfig` accepts `undefined`.

clample:
Looks good to me. I think that you're right that this makes sense for the `destroy` command. Could you add it there as well?

packages/cli/src/commands/deploy.ts
```diff
@@ -39,13 +39,18 @@ export default class Deploy extends AuthCommand {
...
+      description: 'The Checkly CLI config filename.',
+      allowNo: true,
+    }),
   }
 
   async run (): Promise<void> {
     const { flags } = await this.parse(Deploy)
-    const { force, preview, output } = flags
+    const { force, preview, output, config: configFilename } = flags
     const cwd = process.cwd()
```
umutuzgur, line 52: @clample @nahuelon we should probably change `cwd` to the config file directory, given built relative paths according to check and config files. What do you think?

clample: Hmm, that makes sense ➕

clample: Good idea

nahuelon: @umutuzgur , during the coding I tried lot of options related with that. I also tried to allow to enter a path (without filename) in the `--config` argument (and search for the `checkly.config.[ts,js]`). But, I realized about the paths in the config file and I reverted it. I'll try to change the `cwd` to the config file directory, and see if it's safe enoght.

nahuelon: @umutuzgur , this change would require more time to code/test it. Maybe we can move on with the `--config` as it is now and then create a second Issue/PR to modify where the `cwd` should point.

umutuzgur: @nahuelon it should be together so we don't create temporary breaking change

umutuzgur:
@nahuelon if you want to do the work in smaller PRs, feel free to make PRs for deploy, test and destroy sequentially

packages/cli/src/commands/destroy.ts
```diff
@@ -8,18 +9,34 @@ export default class Destroy extends AuthCommand {
...
-      type: 'test',
-      message: `Are you sure you want to delete all resources in project "${checklyConfig.projectName}" for account "${account.name}"?\n  Please confirm by typing the project name "${checklyConfig.projectName}":`,
-    }])
-    if (projectName !== checklyConfig.projectName) {
-      this.log(`The entered project name "${projectName}" doesn't match the expected project name "${checklyConfig.projectName}".`)
-      return
+    if (!force) {
+      const { projectName } = await prompt([{
+        name: 'projectName',
+        type: 'test',
```
umutuzgur, line 33: This really didn't have to be in this PR

umutuzgur: Let's push this separately. The main complexity of this PR is the config flag and the project directory

nahuelon: @umutuzgur , I added here because allows me to execute the `destroy` commands E2E tests. I could move all the `destroy` command tests with the `--force` flag to a new PR if you prefer.

packages/cli/src/services/checkly-config-loader.ts
```diff
@@ -79,19 +79,25 @@ function isString (obj: any) {
   return (Object.prototype.toString.call(obj) === '[object String]')
 }
 
-export async function loadChecklyConfig (dir: string, filenames = ['checkly.config.ts', 'checkly.config.js']): Promise<{ config: ChecklyConfig, constructs: Construct[] }> {
+export async function loadChecklyConfig (filename?: string):
+  Promise<{ config: ChecklyConfig, constructs: Construct[], projectCwd: string }> {
+  const cwd = process.cwd()
```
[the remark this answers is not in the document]
nahuelon: @umutuzgur , first I tried that way (having all the logic in the command files), but I faced lot of typing issues with the `filenames` being and `string[]` and the require some conditional block to handle it. Let me refactor it quickly so I can keep the `loadChecklyConfig()` untounch.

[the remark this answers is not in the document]
nahuelon: @umutuzgur , I kept `loadChecklyConfig()` untouched. I use a new `splitConfigFilePath()` to convert the `--config` value to a `dir` and `filenames` values passed as arguments.

packages/cli/src/commands/destroy.ts
nahuelon: @umutuzgur , I removed the `--force` flag for the `destroy` command. I'll add it (including the destroy E2E tests) after merging these changes.

packages/cli/src/commands/test.ts
```diff
@@ -67,6 +68,11 @@ export default class Test extends AuthCommand {
       description: 'Always show the logs of the checks.',
       allowNo: true,
     }),
+    config: Flags.string({
+      char: 'c',
+      description: 'The Checkly CLI config filename.',
+      allowNo: true,
```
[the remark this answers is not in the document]
nahuelon: Sorry. It was copy&paste issue.

umutuzgur (approved):
LGTM

packages/cli/src/services/util.ts
```diff
@@ -73,5 +73,12 @@ export function pathToPosix (relPath: string): string {
   // Windows uses \ rather than / as a path separator.
   // It's important that logical ID's are consistent across platforms, though.
   // Otherwise, checks will be deleted and recreated when `npx checkly deploy` is run on different machines.
-  return relPath.split(path.sep).join(path.posix.sep)
+  return path.normalize(relPath).split(path.sep).join(path.posix.sep).replace(/^C:/, '')
+}
+
+export function splitConfigFilePath (configFile?: string): { configDirectory: string, configFilenames?: string[] } {
+  return {
+    configDirectory: configFile ? pathToPosix(path.join(process.cwd(), path.dirname(configFile))) : process.cwd(),
```
umutuzgur, line 81: This won't work with absolute paths I imagine?

nahuelon: @umutuzgur , exactly, only relative path at the moment. I had the 'absolute path support' in mind all during the coding, but it would require more caution and validation before allowing to use it. Maybe we can support it in the future.

PR #589, opened 2023-03-16
feat: allow --config with absolute path [gh-0]

nahuelon:

## Affected Components
* [x] CLI
* [ ] Create CLI
* [ ] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer
- allow and handle absolute path in `--config` flag
- add test for `--config` with absolute path.

### DIFF
--- a/packages/cli/src/services/__tests__/checkly-config-loader.spec.ts
+++ b/packages/cli/src/services/__tests__/checkly-config-loader.spec.ts
@@ -26,15 +26,16 @@ describe('loadChecklyConfig()', () => {
     }
   })
   it('config TS file should export an object', async () => {
-    const configFile = './fixtures/configs/good-config.ts'
+    const filename = 'good-config.ts'
+    const configFile = `./fixtures/configs/${filename}`
     const { configDirectory, configFilenames } = splitConfigFilePath(configFile)
 
-    expect(configFilenames).toEqual(['good-config.ts'])
+    expect(configFilenames).toEqual([filename])
     expect(configDirectory).toEqual(pathToPosix(path.dirname(path.join(process.cwd(), configFile))))
 
     const {
       config,
-    } = await loadChecklyConfig(path.join(__dirname, 'fixtures', 'configs'), ['good-config.ts'])
+    } = await loadChecklyConfig(path.join(__dirname, 'fixtures', 'configs'), [filename])
 
     expect(config).toMatchObject({
       checks: {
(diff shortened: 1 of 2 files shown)


packages/cli/src/services/__tests__/checkly-config-loader.spec.ts
```diff
@@ -45,16 +46,38 @@ describe('loadChecklyConfig()', () => {
...
+    const configFile = `./fixtures/configs/${filename}`
+    const { configDirectory, configFilenames } = splitConfigFilePath(path.join(process.cwd(), configFile))
+
+    expect(configFilenames).toEqual([filename])
     expect(configDirectory).toEqual(pathToPosix(path.dirname(path.join(process.cwd(), configFile))))
 
     const {
       config,
-    } = await loadChecklyConfig(path.join(__dirname, 'fixtures', 'configs'), ['good-config.js'])
+    } = await loadChecklyConfig(path.join(__dirname, 'fixtures', 'configs'), [filename])
```
clample, line 80: Why don't we pass the output of `splitConfigFilePath` to the `loadChecklyConfig` call? If the goal of the test is to make sure that `splitConfigFilePath` and `loadChecklyConfig` work nicely together, then that would make more sense to me.

packages/cli/src/services/util.ts
```diff
@@ -77,8 +77,16 @@ export function pathToPosix (relPath: string): string {
 }
 
 export function splitConfigFilePath (configFile?: string): { configDirectory: string, configFilenames?: string[] } {
+  if (configFile) {
+    const dirname = path.dirname(configFile)
+    const cwd = path.isAbsolute(configFile) ? dirname : path.join(process.cwd(), dirname)
```
clample, line 82: I think that the code that you have works. I think that [path.resolve](https://nodejs.org/api/path.html#pathresolvepaths) is a bit more concise, though. It will also remove all of the `..` parts of the path.

```suggestion
    const cwd = path.resolve(path.dirname(configFile))
```

packages/cli/src/services/util.ts
```diff
@@ -77,8 +77,16 @@ export function pathToPosix (relPath: string): string {
 }
 
 export function splitConfigFilePath (configFile?: string): { configDirectory: string, configFilenames?: string[] } {
```
clample, line 79: How do we handle paths with `~` for the homedirectory? For example `~/my-mac-project/checkly.config.ts`?

I think that this isn't supported by Node.js ([here](https://github.com/nodejs/node/issues/684)) so we might need to pull in a library. Maybe we should pull in [untildify](https://www.npmjs.com/package/untildify) as a dependency.

clample: Nevermind, I tested it out. `~` is already handled by the shell.

clample (dismissed):
Looks good 👍 

I added two suggestions, but I think this is good to go as-is.

packages/cli/src/services/util.ts
nahuelon: Good suggestion! Thanks.

clample (approved):

umutuzgur (approved):

umutuzgur:
Should I approve it again?

clample (approved):

PR #592, opened 2023-03-17
Feat/allow signup via cli

### LINKED ISSUE #574: story: enable signup via `login` command
### Story description

With the CLI, using Checkly is terminal focussed. We want to enable users to install the CLI and signup from the CLI in one go. This extends the `login` command that already exists. Adding a separate `signup` commands feels bloated.

## `login` command

We have multiple paths when executing `login`

### No local user credentials

prompt should be split in two options:

1. `Do you want to log in with an existing Checkly account?`
2. `Do you want to sign up for a new Checkly account?`

Case 1: we follow the current, existing path. User logs in via browser session, API key is created.

Case 2: 
- User is directed to `checklyhq.com/signup` with some query param. 
- User signs up but is not directed to the in-app onboarding flow.
- API key is created just like case 1.

End result: User is signed up to a fresh Checkly account and is ready to use the CLI

## Existing user, locally logged in.

Behaviour stays the same, i.e. we ask the user if they want to continue

```
 npx checkly login
? Existing session with account "Production E2E MaC", do you want to continue? (Y/n)                                                                                                                                                               
```

### Shortcut link

_No response_

### Additional resources

_No response_

kiroushi:

## Affected Components
* [x] CLI
* [ ] Create CLI
* [ ] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer
Consolidates all the code related to the auth context in `auth/index.ts` so we can clean up `LoginCommand`.

When retrieving the API key, it will try first to fetch the user, and if it gets a 401 back, it will then register a new one.

> Resolves #574

### DIFF
--- a/packages/cli/src/auth/index.ts
+++ b/packages/cli/src/auth/index.ts
@@ -1,4 +1,4 @@
-import axios from 'axios'
+import axios, { type AxiosError } from 'axios'
 import * as os from 'os'
 import * as http from 'http'
 import * as crypto from 'crypto'
@@ -124,6 +124,20 @@ export async function getAccessToken (code: string, codeVerifier: string) {
 
 // TODO: Move this to API SDK
 export async function getApiKey ({ accessToken, baseHost }: {accessToken: string, baseHost: string}) {
+  try {
+    await fetchUser({
+      accessToken,
+      baseHost,
+    })
+  } catch (error: unknown) {
+    if ((error as AxiosError).response?.status === 401) {
+      console.log(error)
+      await registerUser({ baseHost, accessToken })
+    } else {
+      throw error
+    }
+  }
+
   const { data } = await axios.post(
       `${baseHost}/users/me/api-keys?name=${
         'CLI User Key (' + os.hostname() + ')'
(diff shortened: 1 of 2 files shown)


packages/cli/src/auth/index.ts
```diff
@@ -43,98 +32,170 @@ export function generatePKCE () {
...
+      baseURL: getApiDefaults().baseURL,
       headers: {
-        'Content-Type': 'application/x-www-form-urlencoded',
-        'Accept-Encoding': '*',
+        Accept: 'application/json, text/plain, */*',
+        Authorization: `Bearer ${this.#accessToken}`,
       },
-    },
-  )
+    })
```
umutuzgur, line 62: I have 2 problems with this:
1. Assigning values to the object in between steps usually leads to null|undefined errors
2. You are assigning values in a `get` method which hides the axios creation

umutuzgur: Why don't we just pass axios as function paramters? This would be avoided then

umutuzgur (approved):
LGTM

PR #593, opened 2023-03-17
feat: use group frequency as fallback value for checks [gh-591]

### LINKED ISSUE #591: bug: check frequency property is overwritten
### Node.js version

v16.17.0

### NPM version

8.15.0

### @checkly/cli version

0.4.8

### Steps to reproduce

Having some `.spec.ts` check files:

1. Create a group.check.js with

```js
browserChecks: {
  testMatch: '**/*.spec.ts',
  frequency: 60,
}
```
2. deploy the project

### What is expected?

The browser checks should have a `frequency: 60`

The browser checks have `frequency: 600` (ten minutes), the default frequency from the config file.

### Any additional comments?

The check configuration should be assigned in order:

1. assign config defaults
2. then assign group defaults
3. then assign check specific values.

nahuelon:

## Affected Components
* [x] CLI
* [ ] Create CLI
* [ ] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer
The PR adds the `frequency` as `CheckGroup` and `CheckGroup.browserChecks` property. The new prop is used as a default check group setting, so if a check doesn't have a `frequency` setting can be used.
The behavior is like the `project` setting where the `frequency` in the root is applied to all types of checks and the `browserChecks.frequency` is used just for browser-checks.
Every time a `Check` is instanciated the following logic is used:
1. **(new step added)** Apply `CheckGroup.frequency` if no `Check.frequency` is configured
2. Apply `Project.checks.frequency` if still no `frequency` is configured
3. Apply/Overide with `Check.frequency` if it exists
4. **(new step added)** If it's a `BrowserCheck`, apply `CheckGroup.browserChecks.frequency` if still no `frequency` is set
5. Apply `Project.browserChecks.frequency` if still no `frequency` is configured
6. Apply/Overide with `BrowserCheck.frequency` if it exists

At the moment, only the `frequency` property is taken, but we could add all the `CheckConfigDefaults` properties (`activated`, `muted`, `doubleCheck`, `shouldFail`, etc.). 

`TODO`: add tests when the we confirm the logic is OK.

> Resolves #591


nahuelon:
@tnolet @umutuzgur , I modified the first approach and now the `frequency` is added to the `CheckGroup` and `CheckGroup.browserChecks` following the same pattern as the `project` settings. IMO, this behavior is more clear and easy to follow. Let me know if you are OK with the approach.

tnolet (approved):
This works as described functionally. We need to update the docs after merging this. Please use the `Frequency` construct in any examples.

umutuzgur:
@tnolet I think this is becoming a circular dependency. I don't think we should tackle this case but just document that we do defaults with test matching only

tnolet:
> @tnolet I think this is becoming a circular dependency. I don't think we should tackle this case but just document that we do defaults with test matching only

can you explain a bit more on the circular dependency part? I'm not sure I understand

PR #599, opened 2023-03-20
fix: add correct skipSSL property [gh-597]

### LINKED ISSUE #597: bug: assertion builder error when checking a site with a bad SSL certificate
### Node.js version

v16.17.0

### NPM version

8.15.0

### @checkly/cli version

v0.4.8

### Steps to reproduce

Execute a test with the following API check:

```ts
new ApiCheck('uptime-api-1', {
  name: 'Hello Uptime',
  activated: true,
  request: {
    method: 'POST',
    url: 'https://admin-service-uat.caseclinical.com/graphql',
    bodyType: 'GRAPHQL',
    body: 'query { uptime }',
    assertions: [
        AssertionBuilder.jsonBody('$.data').notEmpty(),
        AssertionBuilder.statusCode().equals(200),
    ]
  },
})
```

### What is expected?

Get the result or a handled error when the SSL is wrong.

`npx checkly test` output:

```
Running 1 checks in eu-west-1.

✖ Hello Uptime (397ms)

/home/yosoy/checkly/checkly-cli/packages/cli/src/reporters/util.ts:208
    return val.toString()
               ^
TypeError: Cannot read properties of undefined (reading 'toString')
    at toString (/home/yosoy/checkly/checkly-cli/packages/cli/src/reporters/util.ts:208:16)
    at truncate (/home/yosoy/checkly/checkly-cli/packages/cli/src/reporters/util.ts:187:16)
    at formatHttpResponse (/home/yosoy/checkly/checkly-cli/packages/cli/src/reporters/util.ts:140:45)
    at formatCheckResult (/home/yosoy/checkly/checkly-cli/packages/cli/src/reporters/util.ts:56:9)
    at ListReporter.onCheckEnd (/home/yosoy/checkly/checkly-cli/packages/cli/src/reporters/list.ts:31:51)
    at CheckRunner.<anonymous> (/home/yosoy/checkly/checkly-cli/packages/cli/src/commands/test.ts:183:16)
    at CheckRunner.emit (node:events:513:28)
    at CheckRunner.emit (node:domain:489:12)
    at MqttClient.<anonymous> (/home/yosoy/checkly/checkly-cli/packages/cli/src/services/check-runner.ts:141:14)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
```

### Any additional comments?

_No response_

nahuelon:

## Affected Components
* [x] CLI
* [ ] Create CLI
* [ ] Test
* [ ] Docs
* [x] Examples
* [ ] Other

## Notes for the Reviewer
- add `skipSSL` request property for API Checks (the `skipSsl` has wrong case)
- set `skipSsl` as to-be-deprecated
- rename property in the example projects.
- handle certificate issues in `formatCheckResult()`.

> Resolves #597

### DIFF
--- a/packages/cli/src/constructs/api-check.ts
+++ b/packages/cli/src/constructs/api-check.ts
@@ -193,7 +193,11 @@ export interface Request {
   url: string,
   method: HttpRequestMethod,
   followRedirects?: boolean,
+  /**
+   * To be deprecated, use `skipSSL`
+   */
   skipSsl?: boolean,
+  skipSSL?: boolean,
   /**
    * Check the main Checkly documentation on assertions for specific values like regular expressions
    * and JSON path descriptors you can use in the "property" field.
@@ -252,7 +256,12 @@ export class ApiCheck extends Check {
 
   constructor (logicalId: string, props: ApiCheckProps) {
     super(logicalId, props)
-    this.request = props.request
+    this.request = {
+      ...props.request,
+      // TODO: remove the `skipSsl` property after deprecate it
+      skipSSL: props.request.skipSSL ?? props.request.skipSsl,
+      skipSsl: undefined,
+    }
     this.localSetupScript = props.localSetupScript
     this.localTearDownScript = props.localTearDownScript
     this.degradedResponseTime = props.degradedResponseTime
--- a/examples/advanced-project/src/services/api/api.check.ts
+++ b/examples/advanced-project/src/services/api/api.check.ts
@@ -10,7 +10,7 @@ new ApiCheck('homepage-api-check-1', {
     url: 'https://api.checklyhq.com/public-stats',
     method: 'GET',
     followRedirects: true,
-    skipSsl: false,
+    skipSSL: false,
     assertions: [
       AssertionBuilder.statusCode().equals(200),
       AssertionBuilder.jsonBody('$.apiCheckResults').greaterThan(0),
--- a/examples/boilerplate-project/__checks__/api.check.ts
+++ b/examples/boilerplate-project/__checks__/api.check.ts
@@ -9,7 +9,7 @@ new ApiCheck('homepage-api-check-1', {
     url: 'https://api.checklyhq.com/public-stats',
     method: 'GET',
     followRedirects: true,
-    skipSsl: false,
+    skipSSL: false,
     assertions: [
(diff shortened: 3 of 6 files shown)


packages/cli/src/constructs/api-check.ts
```diff
@@ -193,7 +193,11 @@ export interface Request {
   url: string,
   method: HttpRequestMethod,
   followRedirects?: boolean,
+  /**
+   * To be deprecated, use `skipSSL`
+   */
   skipSsl?: boolean,
```
umutuzgur, line 199: Since we directly have an error when we set this value atm, we can just delete it already without deprecation

umutuzgur (changes requested):
Code looks good. Let's go ahead with the deprecation in this PR already

clample:
@nahuelon did we have a regression in v0.4.8 or was this bug always there? It seems like this was always an issue, but Tim reported in Slack that earlier versions work fine.

packages/cli/src/constructs/api-check.ts
nahuelon: @umutuzgur , now, the `skipSsl` is deprecated in this PR, if someone is using the `skipSsl` will get:

image

Is this OK?

clample:
> @clample , I tested in v0.4.6 and v0.4.7 locally, the issue is present in those previous versions too.

Thanks for checking that out!

packages/cli/src/constructs/api-check.ts
umutuzgur: Let me triple check with Tim

umutuzgur: Wasn't also `skipSsl: false` throwing the error Tim reported?

nahuelon: @umutuzgur , there were two related issues:
1. if the SSL is wrong, the response format crashed because there is no `response.body`, fixed in `reporters/util.ts` file
2. the `skipSsl` wasn't working as expected (wrong property case), fixed be renaming the prop

If the `skipSsl` would have been working, the `skipSsl: true` would avoid returning an error. Using `skipSsl: false` would have crashing the CLI too.

umutuzgur (approved):
LGTM

PR #609, opened 2023-03-24
feat: introduce checkly trigger command [sc-14206]

clample:

## Affected Components
* [x] CLI
* [ ] Create CLI
* [ ] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer
This PR introduces a new command for running checks, `npx checkly trigger`.  Unlike `npx checkly test`, this lets you run checks that are stored on your Checkly account and not necessarily managed with MaC. It's possible to specify checks to run using tags: `npx checkly trigger -t target-tag`. For the most part, `npx checkly trigger` aims to be as similar as possible to `npx checkly test`.

```
$ npx checkly trigger --help
Trigger checks on Checkly

USAGE
  $ checkly trigger [--private-location <value> | -l <value>] [-t <value>] [-c <value>] [--timeout <value>] [-v] [-r list|dot|ci|github] [-e <value> | --env-file <value>] [--record] [-n <value>]

FLAGS
  -c, --config=<value>             The Checkly CLI config filename.
  -e, --env=<value>...             [default: ] Env vars to be passed to the check run.
  -l, --location=<value>           The location to run the checks at.
  -n, --test-session-name=<value>  A name to use when storing results in Checkly with --record.
  -r, --reporter=<option>          A list of custom reporters for the test output.
                                   <options: list|dot|ci|github>
  -t, --tags=<value>...            Filter the checks to be run using a comma separated list of tags. Checks will only be run if the contain all of the specified tags. Multiple --tags flags can be passed, in which case checks will be run if they match any of the --tags
                                   filters. F.ex. `--tags production,webapp --tags production,backend` will run checks with tags (production AND webapp) OR (staging AND backend).
  -v, --[no-]verbose               Always show the full logs of the checks.
  --env-file=<value>               dotenv file path to be passed. For example --env-file="./.env"
  --private-location=<value>       The private location to run checks at.
  --record                         Record check results in Checkly as a test session with full logs, traces and videos.
  --timeout=<value>                [default: 240] A timeout (in seconds) to wait for checks to complete.

DESCRIPTION
  Trigger checks on Checkly
```

### DIFF
--- a/packages/cli/src/commands/trigger.ts
+++ b/packages/cli/src/commands/trigger.ts
@@ -28,7 +28,7 @@ export default class Trigger extends AuthCommand {
     tags: Flags.string({
       char: 't',
       description: 'Filter the checks to be run using a comma separated list of tags.' +
-        ' Checks will only be run if the contain all of the specified tags.' +
+        ' Checks will only be run if they contain all of the specified tags.' +
         ' Multiple --tags flags can be passed, in which case checks will be run if they match any of the --tags filters.' +
         ' F.ex. `--tags production,webapp --tags production,backend` will run checks with tags (production AND webapp) OR (production AND backend).',
       multiple: true,
--- a/packages/cli/src/commands/trigger.ts
+++ b/packages/cli/src/commands/trigger.ts
@@ -1,0 +1,201 @@
+import { Flags } from '@oclif/core'
+import { isCI } from 'ci-info'
+
+import * as api from '../rest/api'
+import { AuthCommand } from './authCommand'
+import { loadChecklyConfig } from '../services/checkly-config-loader'
+import { splitConfigFilePath, getEnvs, getGitInformation, getCiInformation } from '../services/util'
+import type { Region } from '..'
+import TriggerRunner from '../services/trigger-runner'
+import { RunLocation, Events, PrivateRunLocation, CheckRunId } from '../services/abstract-check-runner'
+import config from '../services/config'
+import { createReporters, ReporterType } from '../reporters/reporter'
+
+const DEFAULT_REGION = 'eu-central-1'
+
+export default class Trigger extends AuthCommand {
+  static hidden = false
+  static description = 'Trigger checks on Checkly'
+  static flags = {
+    location: Flags.string({
+      char: 'l',
+      description: 'The location to run the checks at.',
+    }),
+    'private-location': Flags.string({
+      description: 'The private location to run checks at.',
+      exclusive: ['location'],
+    }),
+    tags: Flags.string({
+      char: 't',
+      description: 'Filter the checks to be run using a comma separated list of tags.' +
+        ' Checks will only be run if the contain all of the specified tags.' +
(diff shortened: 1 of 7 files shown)


packages/cli/src/rest/api.ts
```diff
@@ -8,11 +8,12 @@ import Assets from './assets'
 import Runtimes from './runtimes'
 import PrivateLocations from './private-locations'
 import Locations from './locations'
+import TestSessions from './test-sessions'
 
 export function getDefaults () {
   const environments = {
     production: {
-      apiUrl: 'https://api.checklyhq.com',
+      apiUrl: 'http://localhost:3000',
```
clample, line 16: Just updated this for some local testing ;)

packages/cli/src/services/config.ts
```diff
@@ -68,7 +68,7 @@ const config = {
       development: 'wss://events-local.checklyhq.com',
       test: 'wss://events-test.checklyhq.com',
       staging: 'wss://events-test.checklyhq.com',
-      production: 'wss://events.checklyhq.com',
+      production: 'wss://events-local.checklyhq.com',
```
clample, line 71: Just updated this for some local testing

packages/cli/src/commands/trigger.ts
```diff
@@ -0,0 +1,129 @@
...
+      config.getAccountId(),
+      config.getApiKey(),
+      targetTags,
+      location,
+      240,
+      false,
+    )
+    // TODO: the reporter is quite coupled to having logical Ids. Need to work on that.
+    // let reporter
+    runner.on(Events.RUN_STARTED, (checks) => {
```
clample, line 64: I didn't get it working, but we should be able to set up the reporter here in the `RUN_STARTED` event. We won't know the checks that will be run beforehand, but in `RUN_STARTED` we receive a list of checks.

packages/cli/src/commands/trigger.ts
```diff
@@ -0,0 +1,201 @@
...
+    const repoInfo = getGitInformation()
+    const ciInfo = getCiInformation()
+
+    const runner = new TriggerRunner(
+      config.getAccountId(),
+      timeout,
+      verbose,
+      shouldRecord,
+      location,
+      targetTags?.map((tags: string) => tags.split(',')) ?? [],
```
nahuelon, line 116: @clample , I know it's weird, but we allow users to include commas `,` inside tags. Is it possible to escape them if the user want to filter by a tag having a comma in it. If not, we could restrict the usage of `,` in tags.

clample: Good point @nahuelon 

I think that we can allow users to escape with `\,` if we come up with a regex.

clample: I played around with it, but didn't get a regex with escaping. Let's wait until someone asks for this.

tnolet:
@clample this works pretty damn well. I could not getting it to break 😆.

tnolet (approved):
tested this and works as described.

PR #611, opened 2023-03-27
chore: improve deploy job [sc-0]

umutuzgur:

## Affected Components
* [ ] CLI
* [ ] Create CLI
* [x] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer
* Use a new project logical id for each test
* Increase duration to 20s
* Delete directly the project instead of fetching it if the project logical id is passed

### DIFF
--- a/packages/cli/e2e/__tests__/deploy.spec.ts
+++ b/packages/cli/e2e/__tests__/deploy.spec.ts
@@ -37,7 +37,7 @@ async function cleanupProjects (projectLogicalId?: string) {
 
 describe('deploy', () => {
   // Create a unique ID suffix to support parallel test executions
-  let projectLogicalId
+  let projectLogicalId: string
   // Cleanup projects that may have not been deleted in previous runs
   beforeAll(() => cleanupProjects())
   beforeEach(() => {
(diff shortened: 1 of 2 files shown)


packages/cli/e2e/__tests__/deploy.spec.ts
```diff
@@ -20,25 +20,32 @@ async function cleanupProjects (projectLogicalId?: string) {
...
+    return
+  }
   const { data: projects } = await projectsApi.getAll()
   for (const project of projects) {
-    const matchesLogicalId = project.logicalId === projectLogicalId
     // Also delete any old projects that may have been missed in previous e2e tests
     const leftoverE2eProject = project.name.startsWith('e2e-test-deploy-project-') &&
-      DateTime.fromISO(project.created_at) < DateTime.now().minus(Duration.fromObject({ minutes: 10 }))
-    if (matchesLogicalId || leftoverE2eProject) {
+      DateTime.fromISO(project.created_at) < DateTime.now().minus(Duration.fromObject({ minutes: 20 }))
```
clample, line 31: I wonder if there's a chance that `created_at` isn't consistent with the time in the GH action. That could delete some projects while they're in-use in other tests.

clample (approved):

packages/cli/e2e/__tests__/deploy.spec.ts
umutuzgur: Valid concern but it is just the test projects. We can detect that with test failures later on

PR #621, opened 2023-03-31
feat: allows file dependencies for setup / teardown scripts

### LINKED ISSUE #620: story: allow `import` / `require` for setup and teardown scripts
### Story description

In the current solution (v0.4.9) we only allow direct JS/TS code to be added as a `localSetupScript` and/or `localTearDownScript`. This is very restrictive when building out more sophisticated login and setup routines.

We want to allow users the same experience as with Browser Checks scripts:
- use `import`/`require`
- dry-up code
- use allow-listed external NPM packages like `axios` in file dependencies.

### Shortcut link

https://app.shortcut.com/checkly/story/14893/enable-setup-teardown-scripts-with-dependencies

### Additional resources

_No response_

tnolet:

## Affected Components
* [x] CLI
* [ ] Create CLI
* [x] Test
* [ ] Docs
* [x] Examples
* [ ] Other

## Notes for the Reviewer

This PR introduces two new concepts:
- `scriptPath` for setup & teardown scripts
- `dependencies` for setup & teardown scripts

Usage for setup / teardown scripts will now follow this pattern
- just use `path.join()` to import an entry file.
- note, the `readFileSync()` option still works, so it's backwards compatible. We just check whether the value is a path or an blob of JS/TS code as a string.

```ts
import { ApiCheck, AssertionBuilder } from '@checkly/cli/constructs'
import * as path from 'path'
import { readFileSync } from 'fs'

new ApiCheck('test-api-check-1', {
  name: 'Test API check SETUP TEARDOWN',
  activated: true,
  muted: false,
  frequency: 1,
  localSetupScript: readFileSync(path.join(__dirname, 'setup.ts'), 'utf-8'), // old way, still works
  localTearDownScript: path.join(__dirname, 'teardown.ts'), // new, better way
  doubleCheck: false,
  request: {
    url: 'https://httpbin.org/post',
    method: 'POST',
    body: JSON.stringify({
      data: 1
    }),
    assertions: [
      AssertionBuilder.statusCode().equals(200)
    ]
  }
})
```

> Resolves #620 

Companion PRs that need to go live first:
- https://github.com/checkly/checkly-backend/pull/4046
- https://github.com/checkly/checkly-lambda-runners/pull/1154


packages/cli/src/constructs/api-check.ts
```diff
@@ -241,6 +257,11 @@ export class ApiCheck extends Check {
   localTearDownScript?: string
   degradedResponseTime?: number
   maxResponseTime?: number
+  private readonly setupScriptDependencies?: Array<ScriptDependency>
+  private readonly tearDownScriptDependencies?: Array<ScriptDependency>
+  private readonly setupScriptPath?: string
```
umutuzgur, line 262: We should make this like BCR IMO, like content or entrypoint. That way, they are identical

tnolet: I'm unsure how to rewrite this while staying backwards compatible.

umutuzgur (approved):
LGTM

PR #628, opened 2023-04-05
feat: add mjs support to the cli [gh-616]

### LINKED ISSUE #616: feat: the ability to import / require mjs files using the CLI

Currently, mjs files are not supported by the CLI.

```
`import { test, expect } from "@playwright/test";
import { myFunction } from '../myModule.mjs';


test("coupon", async ({ page }) => {
  myFunction()
  await page.goto("https://danube-web.shop/");
})
export function myFunction() {
  console.log("Hello World!");
}`
```

The above code will generate this error

> The following NPM dependencies were used, but aren't supported in the runtimes.
> For more information, see https://www.checklyhq.com/docs/runtimes/.
>         /Users/jonathancheckly/Desktop/CODE/solutions/template-solutions-scaffold/tests/e2e-guide-tests/coupon.spec.ts imports unsupported dependencies:
>                 .myModule.mjs

There has been some customer feedback regarding this and how cli/constructs impede some workarounds.

Currently this behavior is supported with [GitHub sync](https://github.com/checkly/checkly-webapp/pull/2901), so it would stand to reason that there is some related solution already in place.

umutuzgur:

## Affected Components
* [x] CLI
* [ ] Create CLI
* [x] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer
* Handle mjs files
* Load the check files using the `import` function which supports both commonjs and esm
* Leave a TODO to handle the extensions custom set by the user in the package.json
* Handle `D:` drive letter paths support on Win systems.

> Resolves #616


tnolet:
@nahuelon @clample @umutuzgur once this PR lands, we will need to update the docs to make sure we note that we support MJS files

nahuelon:
@clample @tnolet , I applied some adjustments because the `.mjs` dynamic importing was not working and it has some limitations:
1. Jest fails with a segmentation fault using dynamic importer with TypeScript (more info [here](https://github.com/nodejs/node/issues/35889))
2. A dynamic imported `.mjs` files doesn't handle import dependencies packages with `/` (it tries to search for a file), so I have to `import` the constructor as follow:
```js
import * as Checkly from 'checkly'
const api = new Checkly.constructs.ApiCheck('api-check', {
  name: 'Api Check',
  activated: false,
})
```

Also, the `BrowserCheckExecuter` must be adjusted, to allow `.mjs` files as browser check dependencies: https://github.com/checkly/checkly-lambda-runners/pull/1190

At the moment, the TODOs are:
- investigate if there is a way to import `'checkly/constructs'` in `.mjs` files
- add docs with `.mjs` examples/project and usage instructions.

nahuelon:
@clample , @umutuzgur after testing different projects structure I found that the `lambdas` require the same adjustment applied for browser checks dependencies in https://github.com/checkly/checkly-lambda-runners/pull/1190
So, I created https://github.com/checkly/checkly-lambda-runners/pull/1194 to allow `.mjs` files for `setup`/`tearDown` scripts and their dependencies.

packages/cli/src/services/util.ts
```diff
@@ -37,8 +41,19 @@ export async function walkDirectory (
...
+    if (exported instanceof Function) {
+      exported = await exported()
+    }
+    return exported
+  } catch (err: any) {
+    throw new Error(`Error loading file ${filepath}\n${err.stack}`)
+  }
+}
+
+export async function loadMjsFile (filepath: string): Promise<any> {
```
umutuzgur, line 54: @nahuelon we need to have a single import for both files. The reason is that if a user has type: module in their package.json, the files don't need to have the extenstion `.mjs`

packages/cli/src/services/check-parser/parser.ts
```diff
@@ -41,7 +47,7 @@ const supportedBuiltinModules = [
 
 function validateEntrypoint (entrypoint: string): {extension: SupportedFileExtension, content: string} {
   const extension = path.extname(entrypoint)
-  if (extension !== '.js' && extension !== '.ts') {
+  if (extension !== '.js' && extension !== '.ts' && extension !== '.mjs') {
```
umutuzgur, line 50:
```suggestion
  if (!['.js', '.ts', '.mjs'].includes(extension)) {
```

nahuelon: @umutuzgur , I tried but there is a TS complaint because the transpiler doesn't understand that the `extensions` filtered values are all within the `SupportedFileExtension` values:

image

packages/cli/src/services/util.ts
nahuelon: @umutuzgur , done.

umutuzgur: Should we delete this function now? Since `loadJsFile` now supports

umutuzgur: @nahuelon ☝️

nahuelon: Your are right. Thanks. Done.

packages/cli/src/services/util.ts
```diff
@@ -37,8 +41,14 @@ export async function walkDirectory (
 
 export async function loadJsFile (filepath: string): Promise<any> {
   try {
-    // eslint-disable-next-line @typescript-eslint/no-var-requires
-    let exported = require(filepath)
+    // There is a Node opened issue related with a segmentation fault using ES6 modules
+    // with jest https://github.com/nodejs/node/issues/35889
+    // As a work around, we check if Jest is running to modify the way to import the module.
+    // TODO: investigate if the issue is fixed to clean up the conditional import
+    let { default: exported } = process.env.JEST_WORKER_ID !== undefined
```
umutuzgur, line 48: Instead of using the process env, should we try this `typeof jest !== 'undefined'`? This is in case anyone sets the jest worked id as an env 🤷 

I got the idea from here https://stackoverflow.com/questions/50940640/how-to-determine-if-jest-is-running-the-code-or-not

nahuelon: Done. Good point. I also got the idea from the same post. I like the way you suggest, it seems more secure.

nahuelon (approved):

PR #629, opened 2023-04-05
feat: add gh actions to publish next version [gh-0]

nahuelon:

## Affected Components
* [ ] CLI
* [ ] Create CLI
* [ ] Test
* [ ] Docs
* [ ] Examples
* [x] Other

## Notes for the Reviewer
- add new Github Actions workflow to publish NPM package with `next` tag (for `main` branch)
- add new step in `test.yml` workflow to publish a new NPM package version `<version>-canary-<pullrequest_id>.0` (for opened Pull Requests)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -41,3 +41,19 @@ jobs:
           CHECKLY_ACCOUNT_NAME: ${{ secrets.E2E_CHECKLY_ACCOUNT_NAME }}
           CHECKLY_ACCOUNT_ID: ${{ secrets.E2E_CHECKLY_ACCOUNT_ID }}
           CHECKLY_API_KEY: ${{ secrets.E2E_CHECKLY_API_KEY }}
+  publish-prerelease:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      # Setup .npmrc file to publish to npm
+      - uses: actions/setup-node@v3
+        with:
+          node-version: '16.x'
+          registry-url: 'https://registry.npmjs.org'
+      # Ensure that the README is published with the package
+      - run: rm -f packages/cli/README.md && cp README.md packages/cli
+      - run: npm ci --workspace packages/cli
+      - run: npm version prerelease --preid=canary-${{github.event.pull_request.number}} --workspace packages/cli
+      - run: npm publish --workspace packages/cli --tag next-canary-${{github.event.pull_request.number}}
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
--- a/.github/workflows/release-next.yml
+++ b/.github/workflows/release-next.yml
@@ -1,0 +1,47 @@
+name: Publish Package to npmjs
+on:
+  push:
+    branches:
+      - main
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      # Setup .npmrc file to publish to npm
+      - uses: actions/setup-node@v3
+        with:
+          node-version: '16.x'
+          registry-url: 'https://registry.npmjs.org'
+      # Ensure that the README is published with the package
+      - run: rm -f packages/cli/README.md && cp README.md packages/cli
+      - run: npm version prerelease --preid=beta --workspace packages/cli
+      - run: npm publish --workspace packages/cli --tag next
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+      # Slack failure alert
+      - name: Slack Failure Notification
+        if: ${{ failure() }}
+        uses: rtCamp/action-slack-notify@v2
+        env:
+          SLACK_USERNAME: Checkly Github Bot
+          SLACK_WEBHOOK: ${{ secrets.SLACK_WEBHOOK }}
+          SLACK_CHANNEL: cli
+          SLACK_ICON: https://github.com/checkly.png?size=48
+          SLACK_COLOR: ${{ job.status }}
+          SLACK_TITLE: ':red_circle: NPM next release failed'
+          SLACK_MESSAGE: by ${{ github.actor }}
+          SLACK_FOOTER: ''
+      # Slack success alert
+      - name: Slack Success Notification
+        if: ${{ success() }}
+        uses: rtCamp/action-slack-notify@v2
+        env:
+          SLACK_USERNAME: Checkly Github Bot
+          SLACK_WEBHOOK: ${{ secrets.SLACK_WEBHOOK }}
+          SLACK_CHANNEL: cli
+          SLACK_ICON: https://github.com/checkly.png?size=48
+          SLACK_COLOR: ${{ job.status }}
+          SLACK_TITLE: ':white_check_mark: NPM next release succeeded'
+          SLACK_MESSAGE: by ${{ github.actor }}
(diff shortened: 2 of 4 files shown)


.github/workflows/test.yml
tnolet, line 1: Let call this `canary.yml`, the name `test` is confusing.

.github/workflows/unpublish.yml
```diff
@@ -0,0 +1,18 @@
+name: test
```
tnolet, line 1: the name `test` is confusing

ndom91:
The verison of this I have experience with also appends the commit hash on the end, so yuo end up wtih version's like `next-auth@0.0.0-pr.7056.ced446b7`. 

Check out: 
- https://github.com/nextauthjs/next-auth/blob/main/.github/workflows/release.yml#L85
- https://github.com/nextauthjs/next-auth/blob/main/.github/version-pr/index.js

Also I would definitely recommend including the exact package name in the slack message that's posted / potentially also comment on the PR (See: https://github.com/nextauthjs/next-auth/blob/main/.github/workflows/release.yml#L114)

.github/workflows/release-canary.yml
```diff
@@ -0,0 +1,33 @@
...
+      - run: npm version $(npm run version --workspace packages/cli --silent)-canary-${{github.event.pull_request.number}}.$(git rev-parse --short HEAD) --workspace packages/cli
+      - run: npm publish --workspace packages/cli --tag canary-${{github.event.pull_request.number}}
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+      - uses: marocchino/sticky-pull-request-comment@v2
+        with:
+          header: NPM package canary
+          hide_and_recreate: true
+          hide_classify: "OUTDATED"
+          message: You can install the canary version using `npm install @checkly/cli@canary-${{github.event.pull_request.number}}`
```
ndom91, line 32:
```suggestion
          message: |
            Successfully published `@checkly/cli@canary-${{github.event.pull_request.number}}` 🎉
            
            ```
            npm install @checkly/cli@canary-${{github.event.pull_request.number}}
            ```
```

Little bit "cleaner" imo, and the triple backticks `npm install ..` command at the bottom of the message will include the code block copy btn, so people can copy it out of the comment with 1 simple click

.github/workflows/release-canary.yml
```diff
@@ -0,0 +1,33 @@
...
+      # Ensure that the README is published with the package
+      - run: rm -f packages/cli/README.md && cp README.md packages/cli
+      - run: npm ci --workspace packages/cli
+      - run: npm version $(npm run version --workspace packages/cli --silent)-canary-${{github.event.pull_request.number}}.$(git rev-parse --short HEAD) --workspace packages/cli
+      - run: npm publish --workspace packages/cli --tag canary-${{github.event.pull_request.number}}
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+      - uses: marocchino/sticky-pull-request-comment@v2
+        with:
+          header: NPM package canary
```
ndom91, line 29:
```suggestion
          header: CLI Canary Version Published
```

.github/workflows/release-canary.yml
```diff
@@ -0,0 +1,33 @@
+name: release-canary
+
+on:
+  pull_request:
+    types: [labeled]
```
ndom91, line 5:
```suggestion
on:
  pull_request:
    types:
      - labeled
    paths:
      - 'packages/cli/src/**.ts'
```

.github/workflows/release-canary.yml
```diff
@@ -0,0 +1,33 @@
...
+      # Setup .npmrc file to publish to npm
+      - uses: actions/setup-node@v3
+        with:
+          node-version: '16.x'
+          registry-url: 'https://registry.npmjs.org'
+      # Ensure that the README is published with the package
+      - run: rm -f packages/cli/README.md && cp README.md packages/cli
+      - run: npm ci --workspace packages/cli
+      - run: npm version $(npm run version --workspace packages/cli --silent)-canary-${{github.event.pull_request.number}}.$(git rev-parse --short HEAD) --workspace packages/cli
+      - run: npm publish --workspace packages/cli --tag canary-${{github.event.pull_request.number}}
```
ndom91, line 24:
```suggestion
      - run: npm version 0.0.0-pr.${{github.event.pull_request.number}}.$(git rev-parse --short HEAD) --workspace packages/cli
      - run: npm publish --workspace packages/cli --tag experimental
```

While I get the idea behind making 1 tag per PR, making it easy for folks to always install the latest commit from a specific PR, it'll result in our NPM versions dashboard getting super cluttered and hard to find other tags. Also, for `pr` specific ones, I'd argue we should use the `0.0.0-pr.123.abc123` pattern, similar to the npm [preid](https://docs.npmjs.com/cli/v9/commands/npm-version#preid) flag's behaviour. 

The `canary` tag is usually considered specific type of release, like `alpha`/`beta`. I think it'd be smart if we left those open for ourselves and not use them up for PR preview releases. next-auth, React and others follow a similar formula.

Us:
image


Others: 
image
image

.github/workflows/release-canary.yml
```diff
@@ -0,0 +1,33 @@
...
+      - uses: actions/checkout@v3
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+      # Setup .npmrc file to publish to npm
+      - uses: actions/setup-node@v3
+        with:
+          node-version: '16.x'
+          registry-url: 'https://registry.npmjs.org'
+      # Ensure that the README is published with the package
+      - run: rm -f packages/cli/README.md && cp README.md packages/cli
```
ndom91, line 21:
```suggestion
      - run: rm -f packages/cli/README.md && cp README.md packages/cli
      - run: echo "PR_VERSION=0.0.0-pr.${{github.event.pull_request.number}}.$(git rev-parse --short HEAD)" >> $GITHUB_ENV
```

We can write the `VERSION` to the actions env var's once, and refer to it again elsewhere, even in places where we can't `run` `git rev-parse..` for example

ndom91 (changes requested):
Some formatting and version number comments, but looks really good so far!

ndom91:
Also, just somethign to think about - Instead of running it on a `pull_request: labeled` event, at next-auth we use Github's `environment`'s and branch protection rules to be able to run that `Publish PR` job any time, but require maintainer approval to do so. So you just have to click into the Action and hit "Review deployments" -> "Approve" to kick it off. 

See: https://github.com/nextauthjs/next-auth/blob/main/.github/workflows/release.yml#L90

image

.github/workflows/release-canary.yml
nahuelon: I didn't add the `paths` because sometimes we might want a experimental version just to test a dependency package update (`package.json` changes only). Neither, I can't test this PR because there is no TS file modified.

ndom91: `paths` is an array, and you can add the root and package `package.json` if you want. But yeah, whatever works best for you guys :+1:

ndom91 (approved):
LGTM :+1: :partying_face: 

I can't see the `package-lock.json` changes in the Github UI, too big :sweat_smile:. However, its a bit fishy to me that it's that many changes.. You only added `oclif` as a dev dependency, right? Hmm..

Is the `package-lock.json` format maybe updated from `v1` to `v2` here by accident?

PR #630, opened 2023-04-06
fix: set timeouts before the run command [gh-0]

umutuzgur:

## Affected Components
* [x] CLI
* [ ] Create CLI
* [ ] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer
Set the timeouts before we hit the run session command. It is async so we might get the check run result before the test session rest call ends

### DIFF
--- a/packages/cli/src/services/check-runner.ts
+++ b/packages/cli/src/services/check-runner.ts
@@ -1,5 +1,6 @@
 import { testSessions, assets } from '../rest/api'
 import { SocketClient } from './socket-client'
+import PQueue from 'p-queue'
 import * as uuid from 'uuid'
 import { EventEmitter } from 'node:events'
 import type { AsyncMqttClient } from 'async-mqtt'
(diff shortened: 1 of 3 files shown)


packages/cli/src/services/check-runner.ts
```diff
@@ -201,6 +196,20 @@ export default class CheckRunner extends EventEmitter {
     })
   }
 
+  private setAllTimeouts () {
+    Array.from(this.checks.entries()).forEach(([checkRunId, check]) =>
+      this.timeouts.set(checkRunId, setTimeout(() => {
+        this.timeouts.delete(checkRunId)
```
nahuelon, line 202: @umutuzgur , I would remove this line and always keep the timeout instance. Then, in the  `disableTimeout()`, I would add an `if` condition like:
```ts
if (timeout) {
  clearTimeout(timeout)
}
```

tnolet (approved):
I've run this on a fairly complex project and it works.

PR #632, opened 2023-04-09
add env command(s)

hlenke:
adding command(s) to ls, add, rm, pull Checkly env commands

## Affected Components
* [ x] CLI
* [ ] Create CLI
* [ ] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer

> Resolves #[issue-number]


packages/cli/src/commands/env/rm.ts
```diff
@@ -0,0 +1,41 @@
...
+    if (!args[0]) {
+      throw new Error('Please provide a variable key to delete')
+    }
+    const envVariableKey = args[0]
+    // check if env variable exists
+    const { data: environmentVariables } = await api.environmentVariables.getAll()
+    const envVariable = environmentVariables.find(({ key }) => key === envVariableKey)
+    if (!envVariable) {
+      throw new Error(`Environment variable ${envVariableKey} not found.`)
+    }
```
nahuelon, line 37: @hlenke , the `DELETE` endpoint return a `404` when the variable is not found. Maybe you can avoid the `getAll()` to check if it's exists.

packages/cli/src/commands/env/pull.ts
```diff
@@ -0,0 +1,68 @@
...
+    const exists = fs.existsSync(filename)
+    // check if filename exists and ask for confirmation to overwrite if it does
+    if (exists && !force) {
+      const { confirm } = await prompt([{
+        name: 'confirm',
+        type: 'confirm',
+        message: `Found existing file ${filename}. Do you want to overwrite?`,
+      }])
+      if (!confirm) {
+        this.log('Canceled. No changes made.')
```
nahuelon, line 52: nit
```suggestion
        this.log('Cancelled. No changes made.')
```

packages/cli/src/commands/env/rm.ts
```diff
@@ -0,0 +1,41 @@
+import * as api from '../../rest/api'
+import { Args } from '@oclif/core'
+import { AuthCommand } from '../authCommand'
+
+export default class EnvRm extends AuthCommand {
```
nahuelon, line 5: @hlenke , shouldn't this command have a confirmation dialog `--force` flag?

packages/cli/src/commands/env/rm.ts
```diff
@@ -0,0 +1,41 @@
...
+  static hidden = false
+  static description = 'Remove environment variable via checkly env rm <key>'
```
clample, line 7: For consistency with our other commands, all of the descriptions should end with a period.

packages/cli/src/commands/env/add.ts
```diff
@@ -0,0 +1,48 @@
+import * as api from '../../rest/api'
+import { Flags, Args, ux } from '@oclif/core'
+import { AuthCommand } from '../authCommand'
+
+export default class EnvAdd extends AuthCommand {
```
clample, line 5: It would also be nice to have a way to update environment variables.

packages/cli/src/commands/env/add.ts
```diff
@@ -0,0 +1,48 @@
...
+    if (!subcommands[0]) {
+      throw new Error('Please provide a variable key to add')
+    }
+    const envVariableName = subcommands[0]
+    let locked = false
+    if (subcommands[1]) {
+      locked = subcommands[1] === 'true'
+    }
+    const envValue = await ux.prompt(`What is the value of ${envVariableName}`, { type: 'mask' })
+    await api.environmentVariables.add(envVariableName, envValue, locked)
```
clample, line 45: The backend will return an error if the environment variable already exists. We should catch this and print a user friendly error.

The current behavior is:
```
$ ./bin/run env add foo false
What is the value of foo: ***
Environment variable foo added.
$ ./bin/run env add foo false
What is the value of foo: *****
    AxiosError: Request failed with status code 409
    Code: ERR_BAD_REQUEST
```

clample (changes requested):

clample:
Nice job @hlenke 👍 . If you want help addressing some of the comments, let me know and I could push some changes.

packages/cli/src/commands/env/ls.ts
```diff
@@ -0,0 +1,20 @@
+import * as api from '../../rest/api'
+import { AuthCommand } from '../authCommand'
+import { escapeValue } from '../../services/util'
+
+export default class EnvLs extends AuthCommand {
```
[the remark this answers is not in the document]
hlenke: @clample I looked at what others are doing and Vercel for example got two commands for it: https://vercel.com/docs/cli/env

tnolet:
Haven't checked the code but the main thing in my head is we should not overwrite or edit existing env files. Maybe check if a file exists and ask the user to confirm.

hlenke:
> Haven't checked the code but the main thing in my head is we should not overwrite or edit existing env files. Maybe check if a file exists and ask the user to confirm.

@tnolet https://github.com/checkly/checkly-cli/blob/9099b6d56924d275b59d780e90616031fb4a7914/packages/cli/src/commands/env/pull.ts#L44
That should be what you had in mind. No?

packages/cli/src/commands/env/rm.ts
hlenke: @nahuelon good catch! I added a --force option to it.

packages/cli/src/commands/env/add.ts
```diff
@@ -0,0 +1,48 @@
+import * as api from '../../rest/api'
+import { Flags, Args, ux } from '@oclif/core'
+import { AuthCommand } from '../authCommand'
+
+export default class EnvAdd extends AuthCommand {
+  static hidden = false
+  static description = 'Add environment variable via checkly env add <key> <locked>'
+
+  static flags = {
+    force: Flags.boolean({
```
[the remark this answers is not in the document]
hlenke: @clample Good catch. I made force a locked flag with the latest commit. Hence this should be resolved now.

packages/cli/src/commands/env/ls.ts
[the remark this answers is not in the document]
hlenke: @clample, I'll keep them separated for now. There may be more syncing logic on the pull command in the future. I could imagine us syncing only a Checkly section of the env files and keeping the rest untouched. That will differ from a simple ls drastically.

packages/cli/src/commands/env/add.ts
hlenke: @clample, that absolutely makes sense. I've added that just now with the latest commit.

packages/cli/src/commands/env/add.ts
hlenke: @clample, this totally makes sense. I just added this command.

hlenke:
@nahuelon & @clample, could you please re-review? I've addressed all comments. 🙏

packages/cli/e2e/__tests__/env/env.pull.spec.ts
```diff
@@ -0,0 +1,79 @@
+// create test for checkly env add
```
nahuelon, line 1:
```suggestion
// create test for checkly env pull
```

packages/cli/e2e/__tests__/env/env.upate.spec.ts
```diff
@@ -0,0 +1,46 @@
+// create test for checkly env add
```
nahuelon, line 1: The filename of the test has a typo (missing `d` in `upate`. Also, the first line has a wrong comment.
```suggestion
// create test for checkly env update
```

packages/cli/src/commands/env/ls.ts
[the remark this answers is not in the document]
clample: @hlenke got it, that sounds good 👍

clample:
Aside from the typos @nahuelon found, it looks good to me 👍

clample (dismissed):

clample:
This is merged with https://github.com/checkly/checkly-cli/pull/647

PR #640, opened 2023-04-13
feat: support existing projects [sc-15038]

maxigimenez:

## Affected Components
* [ ] CLI
* [x] Create CLI
* [ ] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer
Ability to initialize the CLI on existing JavasSciprt projects by detecting if `package.json` already exists.

Also, this PR introduces a new folder structure `/actions` with the ability to reuse part of them at different stages of the installation, for example, install dependencies. 

**Note for reviewer:** I'm unsure of a couple of folders/files, feel free to suggest better placing (for example the actions folder). Same as for how the initial `__checks__` & `checkly.config.ts` are generated, on the initial version had static files along the code, but change it to download the boilerplate template and move files around.

https://user-images.githubusercontent.com/1178139/231795157-86cf7b39-7e93-4a56-b54c-eeb06900d832.mp4

### Dependencies
- [normalize-url](https://github.com/sindresorhus/normalize-url)


packages/create-cli/src/actions/creates.ts
```diff
@@ -0,0 +1,36 @@
...
+    { onCancel },
+    )
+
+    fs.writeFileSync('./__checks__/custom.spec.ts', `import { test, expect } from '@playwright/test'
+
+// You can override the default Playwright test timeout of 30s
+// test.setTimeout(60_000);
+
+test('Custom Browser Check', async ({ page }) => {
+  const response = await page.goto('${userWebsiteResponse.url}')
```
ndom91, line 31: If the user doens't enter the protocol, playwright will fail with `invalid URL`. Should we add some validation here? Somethign like

```js
if (!url.startsWith('https://') && !url.startsWith('http://')) {
  url = `https://${myTestUrl}`
}

const normalizedUrl = new URL(url).href;
```

Or `sindresorhus` has a ~25kb library called [normalize-url](https://www.npmjs.com/package/normalize-url) that would do what we need and could potentially be used elsewhere.

maxigimenez: Good point 👍

packages/create-cli/src/actions/creates.ts
```diff
@@ -0,0 +1,36 @@
+import * as fs from 'fs'
+import prompts from 'prompts'
+
+export async function createCustomBrowserCheck (
+  { onCancel }: { onCancel: () => void },
+) {
+  const createInitialBrowserCheckResponse = await prompts({
+    type: 'confirm',
+    name: 'value',
+    message: 'Would you like to create a custom Browser Check?',
```
tnolet, line 10:
```suggestion
    message: 'Would you like to create a custom Playwright-based Browser Check to check a URL?',
```

packages/create-cli/src/actions/creates.ts
```diff
@@ -0,0 +1,36 @@
...
+    initial: true,
+  },
+  { onCancel },
+  )
+
+  if (createInitialBrowserCheckResponse.value) {
+    const userWebsiteResponse = await prompts({
+      type: 'text',
+      name: 'url',
+      message: 'Could you provide your website homepage URL?',
```
tnolet, line 20:
```suggestion
      message: 'Please provide the URL of the site you want to check.',
```

tnolet (changes requested):
@maxigimenez just provided some copy updates.

ndom91 (approved):
LGTM :+1:

ndom91 (approved):

PR #644, opened 2023-04-14
feat: support fromId for alert channels using the new endpoint [gh-567]

### LINKED ISSUE #567: story: enable shared / externally managed resources through `fromId()`
### Story description

Users should be able to reference existing resources on the Checkly cloud that are not managed by their project, but potentially by another project or not managed at all via the CLI.

Some practical cases:

- A user can have two or more projects that live in separate Git repos that need to target shared resources: `webapp` , `site` and `backend`. They need to use the same, shared alert channels. But also think of other shared, cross project resources like a dashboard.
- A user should be enabled to migrate to CLI without overhauling their setup from the start. This migration is easier when the user can keep their alerting setup and just migrate some Checks to start out with.

## Proposal

```ts
export const OpsGenieChannel = new OpsgenieAlertChannel.fromId(158554)
```

## Limitations
- This only works and is only valid within the scope of one Checkly account for security reasons.
- Removing a shared resource will break the setup on a next `deploy` as the reference no longer exists. We need to have a friendly error message.

### Shortcut link

_No response_

### Additional resources

_No response_

umutuzgur:

## Affected Components
* [x] CLI
* [ ] Create CLI
* [ ] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer
* use the new deploy endpoint, the tests will fail until the BE is deployed
* update __checklyType to match the BE types expected
* add fromId(id) support for alert channels. This will work with `OpsgenieAlertChannel.fromId` as well

> Resolves #567


packages/cli/src/constructs/alert-channel.ts
```diff
@@ -54,11 +66,15 @@ export abstract class AlertChannel extends Construct {
     this.sslExpiryThreshold = props.sslExpiryThreshold
   }
 
+  static fromId (id: string|number) {
```
nahuelon, line 69: @umutuzgur , I think you could specify just `number` here. The table id is a `serial`.

```suggestion
  static fromId (id: number) {
```

packages/cli/src/constructs/alert-channel.ts
```diff
@@ -54,11 +66,15 @@ export abstract class AlertChannel extends Construct {
...
+    return new AlertChannelWrapper(`alert-channel-${id}`, id)
+  }
+
   allowInChecklyConfig () {
     return true
   }
 
-  synthesize () {
+  synthesize (): any|null {
```
nahuelon, line 77:
```suggestion
  synthesize (): {
    sendRecovery?: boolean
    sendFailure?: boolean
    sendDegraded?: boolean
    sslExpiry?: boolean
    sslExpiryThreshold?: number
    } {
```

packages/cli/src/constructs/construct.ts
```diff
@@ -18,7 +22,7 @@ export abstract class Construct {
     return false
   }
 
-  abstract synthesize(): any
+  abstract synthesize(): any|null
```
nahuelon, line 25: @umutuzgur , I think we should export the new synthesized types/interfaces in each constructs and import them here to indicate the synthesize() return types using `|`.

packages/cli/src/constructs/alert-channel.ts
```diff
@@ -54,11 +66,15 @@ export abstract class AlertChannel extends Construct {
     this.sslExpiryThreshold = props.sslExpiryThreshold
   }
 
+  static fromId (id: string|number) {
+    return new AlertChannelWrapper(`alert-channel-${id}`, id)
```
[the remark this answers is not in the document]
clample: Nevermind, since this is the physical UUID, I don't think that users will run into a duplication. It might still be worth changing the formatting, though, IMO (as a nit).

packages/cli/src/constructs/alert-channel.ts
umutuzgur: We should do this but not with this PR IMO

clample (approved):

PR #646, opened 2023-04-14
fix: let parser handle the file missing case [gh-0]

umutuzgur:

## Affected Components
* [x] CLI
* [ ] Create CLI
* [ ] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer
Let the parser handle the missing entrypoint file


tnolet (approved):
yes, this is a better solution after testing it out. Also the error message if you have a typo in the file name is better.

PR #649, opened 2023-04-24
feat: add high frequencies [sc-0]

### LINKED ISSUE #454: Add high frequency support to the check constructs
We haven't added high frequency configuration option to the checks. We need to set the frequency offset to achieve this. We can actually use enums for this maybe? It would be neater for the users

umutuzgur:

## Affected Components
* [ ] CLI
* [ ] Create CLI
* [ ] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer
* Adds a new class to avoid magic numbers as frequencies like, `Frequency.everySecond(10)`
* Add high frequency support using the new `Frequency` class

> Resolves #454


packages/cli/src/constructs/frequency.ts
```diff
@@ -0,0 +1,38 @@
...
+  everyMinute (freq: Minute) {
+    if (!MIN_FREQS.includes(freq)) {
+      throw new Error(`Unsupproted freq ${freq}`)
+    }
+    return new Frequency(freq)
+  }
+
+  everyHour (freq: Hour) {
+    if (!HOUR_FREQS.includes(freq)) {
+      throw new Error(`Unsupproted freq ${freq}`)
```
tnolet, line 34: "unsupported"

tnolet (changes requested):
Ok, I took it for a spin. As we are essentially restricting the possible values the `everyMinute()` like functions can take, we should probably not make that a guess game for the users. 

- Let's go for full ENUMs so the user can just pick it from their code completion list.
- Add a simple unit test
- update the examples

tnolet:
@umutuzgur this works fine: no thinking / guessing from the user required.

umutuzgur:
Nice. I will prep the PR then

tnolet (approved):
Lgtm

PR #662, opened 2023-04-26
refactor: prepare reporters for checkly trigger command [sc-14206]

clample:

## Affected Components
* [x] CLI
* [ ] Create CLI
* [ ] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer
The reporters are currently quite coupled to `checkly test` and will be tricky to work with for `checkly trigger`. This PR refactors the reporters to prepare for the `checkly trigger` command (https://github.com/checkly/checkly-cli/pull/609).

In particular:
1. All of the `Reporters` currently take the list of checks that will be run in the constructor. With `checkly trigger`, though, we won't know in advance which checks will actually be run (since the backend applies tag filters). This PR makes the reporters accept a list of checks that will be run in `onBegin` instead. Unfortunately, this means that we sometimes need to use the null assertion operator `!` in the reporters.
2. `AbstractListReporter` currently tracks the state of checks using a map `checkFilesMap` from `file -> check logicalId -> check+result`. With `checkly trigger`, though, we can't assume that all checks have a logical ID. This PR refactors the map to track the check state using the `CheckRunId` as a key instead, since we always have this for checks.
     * `checkly trigger` won't have a `file` for checks either. This could cause problems for `AbstractListReporter.checkFilesMap`. To work around this, we can simply use `null` as the file for `checkly trigger`. The reporters will need to be updated accordingly, to just not print a filename if the file is `null`.
3. Currently the reporters track the `testResultIds` as a map from logical ID to the `testResultId`. We won't have logical IDs for `checkly trigger`, though. Instead, we can move the `testResultId`'s into `AbstractListReporter.checkFilesMap` which is indexed by `CheckRunId`.

### DIFF
--- a/packages/cli/src/commands/test.ts
+++ b/packages/cli/src/commands/test.ts
@@ -5,7 +5,7 @@ import { isCI } from 'ci-info'
 import * as api from '../rest/api'
 import config from '../services/config'
 import { parseProject } from '../services/project-parser'
-import { Events, RunLocation, PrivateRunLocation } from '../services/abstract-check-runner'
+import { Events, RunLocation, PrivateRunLocation, CheckRunId } from '../services/abstract-check-runner'
 import TestRunner from '../services/test-runner'
 import { loadChecklyConfig } from '../services/checkly-config-loader'
 import { filterByFileNamePattern, filterByCheckNamePattern } from '../services/test-filters'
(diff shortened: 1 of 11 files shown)


nahuelon (approved):
LGTM.

nit: I would say there are lot of null assertion operators `!`. Maybe, in the future we can refactor the reporters to compliant `TS` good practices.

nit: use `import type` when `CheckRunId` type is imported.

Issue #566, opened 2023-03-03
story: move `@checkly/cli` package to `checkly` package at version 4.0.0.


tnolet:
### Story description

We (very) recently got hold of the top-level `checkly` package name. This package name is better and clearer than the scoped `@checkly/cli` package as it does the functionality is larger than only a CLI. Because we inherited this package name, we need to start at version `4.0.0.` as we cannot reuse earlier versions.

This move includes:

- [ ]  start publishing the `@checkly/cli` to the `checkly` package name at 
- [ ] start publishing the `@checkly/create-cli` to the `create-checkly` package name
- [ ] rename the GitHub repo to `checkly`
- [ ] communicate to users that this change happened: email, Slack, Twitter, Changelog and inside the CLI.
- [ ] keep mirroring the packages on the scoped `@checkly/` packages for 1 month.

PR #669, opened 2023-04-28
fix: continue reading config for @checkly/cli

clample:

## Affected Components
* [x] CLI
* [ ] Create CLI
* [ ] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer

> Relates to #566

We currently use [conf](https://www.npmjs.com/package/conf) for persisting credentials.  Conf persists credentials using a filename containing the package name. On MacOS, CLI credentials were persisted in `~/Library/Preferences/@checkly/cli`.

With https://github.com/checkly/checkly-cli/pull/665, we changed the package name to `checkly` and so the CLI now reads credentials from  `~/Library/Preferences/checkly`. This means that users migrating to the new CLI version need to recreate credentials and log in to Checkly!

To avoid this inconvenience and make the upgrade process smoother, we can simply continue using the old config file.

### DIFF
--- a/packages/cli/src/services/config.ts
+++ b/packages/cli/src/services/config.ts
@@ -21,12 +21,14 @@ enum Env {
 
 const config = {
   auth: new Conf({
+    projectName: '@checkly/cli',
     configName: 'auth',
     projectSuffix,
     // @ts-ignore
     schema: authSchema,
   }),
   data: new Conf({
+    projectName: '@checkly/cli',
     configName: 'config',
     projectSuffix,
     // @ts-ignore
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -1,6 +1,6 @@
 {
   "name": "checkly",
-  "version": "4.0.0",
+  "version": "4.0.1",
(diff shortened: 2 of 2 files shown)


packages/cli/src/services/config.ts
```diff
@@ -21,12 +21,14 @@ enum Env {
 
 const config = {
   auth: new Conf({
+    projectName: '@checkly/cli',
```
nahuelon, line 24: @clample , isn't it possible to create a fallback logic? I mean, first check `projectName: 'checkly'`, then the legacy one. If not, users with the new version will still being storing their config in the old project name.
Another option could be to store config in both `projectNames` making easier to deprecate the `@checkly/cli` project name in the future.

nahuelon (approved):
LGTM (but I left a comment above)

packages/cli/src/services/config.ts
clample: @nahuelon that might be possible. For now, though, I don't see any downsides to just only storing the config in the `@checkly/cli` file.

PR #680, opened 2023-05-01
feat: show shortlinks for failed checks and test-session summary [sc-…

nahuelon:
…15066]

## Affected Components
* [x] CLI
* [ ] Create CLI
* [ ] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer
We want to show short-links for assets (trace and video files) and test-results pages, ONLY for failed check results and using the `list` reporter. Also, the test-session URL is shown with a short-link.
I have to request the short-links when processing the socket messages because doing it in the `onCheckEnd()` resulted in getting the terminal prints summaries disorder, caused by the `RUN_FINISHED` is received before the the short-link requests completes.
The `github` reporter is still using the long URL because they are used within markdown links.
The request are wrapped in try/catch blocks so, if there is some problem obtaining the short-links, nothing is shown for check failures and, for test-session summary, it shows the long URL.

> Result 
image

> Resolves [[sc-10566]](https://app.shortcut.com/checkly/story/15066/add-chkly-link-shortlink-to-test-results-link-in-the-cli-terminal-output)

> PR https://github.com/checkly/checkly-backend/pull/4129 contains the new API endpoints.

### DIFF
--- a/packages/cli/src/services/abstract-check-runner.ts
+++ b/packages/cli/src/services/abstract-check-runner.ts
@@ -1,10 +1,11 @@
-import { assets } from '../rest/api'
+import { assets, testSessions } from '../rest/api'
 import { SocketClient } from './socket-client'
 import PQueue from 'p-queue'
 import * as uuid from 'uuid'
 import { EventEmitter } from 'node:events'
 import type { AsyncMqttClient } from 'async-mqtt'
 import type { Region } from '..'
+import { TestResultsShortLinks } from '../rest/test-sessions'
 
 // eslint-disable-next-line no-restricted-syntax
 export enum Events {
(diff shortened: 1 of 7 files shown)


clample (dismissed):

packages/cli/src/services/abstract-check-runner.ts
```diff
@@ -136,22 +144,9 @@ export default abstract class AbstractCheckRunner extends EventEmitter {
...
-          }
-          if (playwrightTestVideoFiles && playwrightTestVideoFiles.length) {
-            result.videoFilesUrls = await Promise
-              .all(playwrightTestVideoFiles.map((t: string) => assets.getAssetsLink(region, t)))
-          }
-        }
-      } catch {
-        // TODO: remove the try/catch after deploy endpoint to fetch assets link
-      }
+      const links = testResultId && result.hasFailures && await this.getShortLinks(testResultId)
```
clample, line 147: More of a product question - should we also set a shortlink if the `verbose` flag is passed? I think that that could be useful. It'll start creating many shortlinks, though.

clample: I asked Tim in Slack. Is it straightforward to also check the `verbose` flag here as well? We could also come back to this in a later PR.

PR #681, opened 2023-05-02
fix: print checks for 'checkly test --list'

### LINKED ISSUE #677: bug: `checkly test --list` runs into type error
### Node.js version

19

### NPM version

9

### @checkly/cli version

4.0.3

### Steps to reproduce

1. Setup some test
2. Run `checkly test --list` inside the project

### What is expected?

The checks setup in the project are listed

An error is thrown

```bash
Parsing your project... done

Listing all checks:

    TypeError: Cannot read properties of undefined (reading 'size')
```

### Any additional comments?

This happens only with the recently released version of the CLI of v4.

clample:

## Affected Components
* [x] CLI
* [ ] Create CLI
* [ ] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

There's been a regression in `checkly test --list` (for more info, see the linked issue). This regression was introduced in https://github.com/checkly/checkly-cli/pull/662, where we required the list of checks to be passed to the reporter in `reporter.onBegin()` rather than in the constructor. Since `onBegin()` wasn't called for `checkly test --list`, the reporter wasn't properly initialized and there would be an error.

This PR resolves the issue by removing `reporter.onBeginStatic()`, which is the reporter-specific implementation for `checkly test --list`. `onBeginStatic()` was nice since it let `checkly test --list` reuse code from the reporter, but I think that it's also awkward having each reporter implement printing for `checkly test --list`. For example, the GitHub reporter would print "Running 5 checks in ...", which doesn't make sense for `--list` since the checks aren't running.

Rather than use `reporter.onBeginStatic()`, we now just have one method `Test.listChecks()` for printing the checks for `checkly test --list`. This new method doesn't require a reporter to be set up.

The integration test is from Tim: https://github.com/checkly/checkly-cli/pull/679

### DIFF
--- a/packages/cli/e2e/__tests__/test.spec.ts
+++ b/packages/cli/e2e/__tests__/test.spec.ts
@@ -63,6 +63,17 @@ describe('test', () => {
     expect(result.status).toBe(0)
   })
 
+  it('Should list checks and not execute them with `--list`', () => {
+    const result = runChecklyCli({
+      args: ['test', '--list'],
+      apiKey: config.get('apiKey'),
+      accountId: config.get('accountId'),
+      directory: path.join(__dirname, 'fixtures', 'test-project'),
+    })
+    expect(result.stdout).toContain('Listing all checks')
+    expect(result.status).toBe(0)
+  })
+
   it('Should terminate with error when JS/TS throws error', () => {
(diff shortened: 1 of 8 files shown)


nahuelon (approved):
LGTM

PR #684, opened 2023-05-03
test: add integration test for checkly trigger

clample:

## Affected Components
* [ ] CLI
* [ ] Create CLI
* [x] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer
We recently added a new command `checkly trigger`, which existing checks stored in Checkly and supports filtering by tags. To move quickly, I simply tested the new command manually but didn't add any integration tests. To avoid future regressions, though, this PR adds an integration test.

The test deploys checks to the e2e checkly account. It then executes them with `checkly trigger`. It asserts that all checks matching the tag filter are run, and checks not matching the tag filter are not run. It also checks that environment variables using the `-e` flag are propagated.

One challenge is that we might have this test running from multiple CI jobs in parallel. To avoid interference, the test generates a random `EXECUTION_ID`. This ID is then used in the project ID and in the tags of the checks.


packages/cli/e2e/__tests__/fixtures/trigger-project/checks.check.ts
```diff
@@ -0,0 +1,51 @@
...
+  request: {
+    url: 'https://www.google.com',
+    method: 'GET',
+    followRedirects: false,
+    skipSSL: false,
+    assertions: [],
+  },
+  tags: ['production', 'frontend', executionId, 'another-tag'],
+  // Log the secret environment variable passed to `checkly trigger`
+  localSetupScript: "console.log(process.env.SECRET_ENV)",
```
nahuelon, line 36: The `localSetupScript` is marked as `deprecated`. Maybe, you can use the new `setupScript`

nahuelon (dismissed):
LGTM

packages/cli/e2e/__tests__/fixtures/trigger-project/checks.check.ts
clample: Good idea, thanks! I updated the PR.

nahuelon (approved):
LGTM

PR #692, opened 2023-05-04
feat: change handling for 'checkly trigger' no matching checks

### LINKED ISSUE #685: feat: add option to pass without tests using `trigger`

Depending on the usage of the `trigger` command it might be expected that no tests can be triggered as none exists. A use case in which this is expected is when running Checkly across many environments from a CI pipeline dynamically while certain non crucial environments might not have any checks defined matching the `--tags` selected.

Failing if no checks are found is a very good default. For instance Jest employs this too but allows opting out with a [--passWithNoTests](https://jestjs.io/docs/cli#--passwithnotests). 

Adding a `--pass-without-checks` could be an option added to the `trigger` command to opt out of failing if no matching checks are found.

clample:

## Affected Components
* [x] CLI
* [ ] Create CLI
* [ ] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer

> Resolves #685

Currently when no checks match the `checkly trigger` tag filters, the backend returns an error and the CLI exits with status code 1. To be consistent with `checkly test` and to support valid CI use cases (for example, dynamic tag filters by environment that don't necessarily match any checks), we should change `checkly trigger` to exit with code 0 in this case.

This PR implements this in the simplest way that I saw possible - `TriggerRunner` throws a special error if it detects no matching checks, and the `Trigger` handler logs an error. In the future, it might make sense to not throw an error at all, and just let the reporters detect and handle this case with the normal `onBegin()`, `onEnd()` hooks.

In the future, I think that it could make sense for the backend to simply return an empty array of checks rather than marking this as an error. Since this would be a breaking change for the CLI, I also add handling for this case to the CLI now. This should give us more flexibility to make this change later on.


packages/cli/e2e/__tests__/trigger.spec.ts
```diff
@@ -50,4 +50,19 @@ describe('trigger', () => {
     expect(result.stdout).not.toContain('Staging Backend Check')
     expect(result.status).toBe(0)
   })
+
+  test('Should return code 0 when no checks match', () => {
```
clample, line 54: This won't pass until https://github.com/checkly/checkly-backend/pull/4139 is deployed

umutuzgur (approved):
LGTM

PR #693, opened 2023-05-04
fix: update checkMatch to support check-group.ts

clample:

## Affected Components
* [ ] CLI
* [ ] Create CLI
* [ ] Test
* [ ] Docs
* [x] Examples
* [ ] Other

## Notes for the Reviewer
The advanced example currently contains a [website.check-group.ts](https://github.com/checkly/checkly-cli/blob/main/examples/advanced-project/src/__checks__/website.check-group.ts) check file. Declaring constructs in `*.check-group.ts` files isn't fully supported, though, since the examples have `checkMatch: *.check.ts` (not matching any check-group.ts files).

The reason why the constructs in [website.check-group.ts](https://github.com/checkly/checkly-cli/blob/main/examples/advanced-project/src/__checks__/website.check-group.ts) are loaded, despite not matching the `checkMatch` pattern, is that the file is imported by other `*.check.ts` files. For users that build on the advanced example, though, and try to use the `*.check-group.ts` pattern, there setup won't work.

To fix this, we could either:
1. rename `website.check-group.ts`, so that there's no suggestion of using a `*.check-group.ts` pattern
2. update `checkMatch` to support `*.check-group.ts` files

This PR takes approach 2 to give full support to `check-group.ts`. In addition to updating the examples, I also update the default `checkMatch` value. This way users don't run into any surprises when using `check-group.ts`.

### DIFF
--- a/examples/advanced-project/checkly.config.ts
+++ b/examples/advanced-project/checkly.config.ts
@@ -25,7 +25,7 @@ const config = defineConfig({
      */
     runtimeId: '2023.02',
     /* A glob pattern that matches the Checks inside your repo, see https://www.checklyhq.com/docs/cli/using-check-test-match/ */
-    checkMatch: '**/__checks__/**/*.check.ts',
+    checkMatch: '**/__checks__/**/*.check?(-group).{js,ts}',
     browserChecks: {
       /* A glob pattern matches any Playwright .spec.ts files and automagically creates a Browser Check. This way, you
       * can just write native Playwright code. See https://www.checklyhq.com/docs/cli/using-check-test-match/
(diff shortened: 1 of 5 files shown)


clample:
@umutuzgur should I check with Tim on this one, or should we just go forward with it?

clample:
Should I go with option 1 then, and just rename `website.check-group.ts` to `website-group.check.ts`? I don't have a strong opinion either way - I just think the examples need to be consistent/functional.

umutuzgur:
I think that is the right way to go

clample:
I'll do that in another PR

PR #698, opened 2023-05-05
chore: handle concurrent e2e test runs for env tests

clample:

## Affected Components
* [ ] CLI
* [ ] Create CLI
* [x] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer
The e2e tests for `checkly env` modify the account environment variables on the e2e test Checkly account. When there are multiple e2e tests ongoing in parallel, they can then interfere with each other and cause failures. For example, one test instance could delete the environment variable in it's cleanup right after it's added in the setup of the other test instance.

Running the e2e tests in parallel locally, this does end up causing failures. Looking at our history of flaky GitHub actions, though, this only caused a few flaky runs. Still, we should fix the issue.

This PR fixes the issue by creating a unique `executionId`. This string is then appended to the environment variable names.

## New Dependencies
`nanoid`: this is used for generating the unique ID's. It's more convenient than `uuid` since we can create shorter IDs.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10877,6 +10877,18 @@
       "version": "0.0.8",
       "license": "ISC"
     },
+    "node_modules/nanoid": {
+      "version": "3.3.4",
+      "resolved": "https://registry.npmjs.org/nanoid/-/nanoid-3.3.4.tgz",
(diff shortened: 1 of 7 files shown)


umutuzgur (approved):
Nice one!

PR #699, opened 2023-05-05
chore: increase timeout for flaky e2e tests

clample:

## Affected Components
* [ ] CLI
* [ ] Create CLI
* [x] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer
Currently the e2e test GH Action can be a bit flaky. After looking at past flakiness, it looks like most of the failures (not all, though) are from `test › Test project should run successfully` and `test › Should use Github reporter`.  Both of these tests run [homepage.test.ts](https://github.com/checkly/checkly-cli/blob/main/packages/cli/e2e/__tests__/fixtures/test-project/src/__checks__/homepage.test.ts), which is an actual PW test taking several seconds. From looking at the failures (f.ex: [here](https://github.com/checkly/checkly-cli/actions/runs/4862530190/jobs/8669023267#step:7:49) or [here](https://github.com/checkly/checkly-cli/actions/runs/4861862224/jobs/8667435359#step:7:27)), it looks like `checkly test` reaches the 30 second timeout for `runChecklyTest`. That would explain the `null` status code for `test › Test project should run successfully` failures and why the GitHub summary isn't written for `test › Should use Github reporter` failures.

To fix the issue, this PR simply bumps the timeout for these two tests. If we continue seeing flaky test failures, we can try other solutions as well.

### DIFF
--- a/packages/cli/e2e/__tests__/test.spec.ts
+++ b/packages/cli/e2e/__tests__/test.spec.ts
@@ -12,6 +12,7 @@ describe('test', () => {
       apiKey: config.get('apiKey'),
       accountId: config.get('accountId'),
       directory: path.join(__dirname, 'fixtures', 'test-project'),
+      timeout: 120000, // 2 minutes
     })
     expect(result.stdout).not.toContain('File extension type example')
     expect(result.stdout).toContain(secretEnv)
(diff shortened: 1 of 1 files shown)


umutuzgur (approved):
LGTM

PR #703, opened 2023-05-08
fix: detect if package.json exists [sc-15315]

maxigimenez:

## Affected Components
* [ ] CLI
* [x] Create CLI
* [ ] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer
For some reason `npm create checkly` doesn't detect `package.json` even if exists. Couldn't reproduce it in local mode, but this fix aims to solve it by having the full path of the file.

### DIFF
--- a/packages/create-cli/src/utils/package.ts
+++ b/packages/create-cli/src/utils/package.ts
@@ -9,7 +9,7 @@ export interface PackageJson {
 }
 
 export function hasPackageJsonFile () {
-  return fs.existsSync(path.join(process.cwd(), './package.json'))
+  return fs.existsSync(path.join(process.cwd(), 'package.json'))
 }
 
 export function readPackageJson (): PackageJson {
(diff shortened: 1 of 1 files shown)


shiini2 (approved):

packages/create-cli/src/utils/package.ts
```diff
@@ -8,7 +9,7 @@ export interface PackageJson {
 }
 
 export function hasPackageJsonFile () {
-  return fs.existsSync('./package.json')
+  return fs.existsSync(path.join(process.cwd(), './package.json'))
```
shiini2, line 12: Is the `./` still needed before the file name now that we're also including the path?

maxigimenez: @shiini2 is not, PR updated

ndom91 (approved):

PR #710, opened 2023-05-10
fix: ensure check run timeouts give clear error message

clample:

## Affected Components
* [x] CLI
* [ ] Create CLI
* [ ] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer
There is currently an issue with timeout support for `checkly test` and `checkly trigger`. This can be reproduced by running with `--timeout 0`:

```
$ npx checkly test --timeout 1
Parsing your project... done

Running 2 checks in eu-west-1.

__checks__/api.check.ts
  - Fetch Book List
__checks__/homepage.spec.ts
  - homepage.spec.ts

2 pending, 2 total

/Users/clample/projects/checkly-cli/packages/cli/src/commands/test.ts:211
      reporters.forEach(r => r.onCheckEnd(checkRunId, {
                ^
TypeError: check.getSourceFile is not a function
    at /Users/clample/projects/checkly-cli/packages/cli/src/commands/test.ts:214:27
    at Array.forEach (<anonymous>)
    at TestRunner.<anonymous> (/Users/clample/projects/checkly-cli/packages/cli/src/commands/test.ts:211:17)
    at TestRunner.emit (node:events:513:28)
    at TestRunner.emit (node:domain:489:12)
    at Timeout._onTimeout (/Users/clample/projects/checkly-cli/packages/cli/src/services/abstract-check-runner.ts:176:14)
    at listOnTimeout (node:internal/timers:559:17)
    at processTimers (node:internal/timers:502:7)
```

After more investigation, this is a regression from https://github.com/checkly/checkly-cli/pull/662. This PR changed the arguments for the `CHECK_FAILED` event, but didn't update the timeout code path.

This PR fixes the issue and adds an integration test. The timeout case is now working as it did before:
```
$ npx checkly test --timeout 1
Parsing your project... done

Running 2 checks in eu-west-1.

✖ homepage.spec.ts

    ──Execution Error───
    Reached timeout of 1 seconds waiting for check result.

__checks__/api.check.ts
  ✔ Fetch Book List (233ms)
__checks__/homepage.spec.ts
  ✖ homepage.spec.ts

1 failed, 1 passed, 2 total
```

When the default timeout is used (300s), this case should only occur for Checkly infrastructure problems. Checkly should always report a result within 240s. In this case, the CLI also points the user to the status page and support email address:

```
✖ homepage.spec.ts

    ──Execution Error────
    Reached timeout of 300 seconds waiting for check result. Checkly may be experiencing problems. Please check https://is.checkly.online or reach out to support@checklyhq.com.
```

### DIFF
--- a/packages/cli/src/commands/test.ts
+++ b/packages/cli/src/commands/test.ts
@@ -6,7 +6,13 @@ import { isCI } from 'ci-info'
 import * as api from '../rest/api'
 import config from '../services/config'
 import { parseProject } from '../services/project-parser'
-import { Events, RunLocation, PrivateRunLocation, CheckRunId } from '../services/abstract-check-runner'
+import {
+  Events,
+  RunLocation,
+  PrivateRunLocation,
+  CheckRunId,
+  DEFAULT_CHECK_RUN_TIMEOUT_SECONDS,
+} from '../services/abstract-check-runner'
 import TestRunner from '../services/test-runner'
 import { loadChecklyConfig } from '../services/checkly-config-loader'
 import { filterByFileNamePattern, filterByCheckNamePattern } from '../services/test-filters'
@@ -56,7 +62,7 @@ export default class Test extends AuthCommand {
       description: 'list all checks but don\'t run them.',
     }),
     timeout: Flags.integer({
-      default: 300,
+      default: DEFAULT_CHECK_RUN_TIMEOUT_SECONDS,
       description: 'A timeout (in seconds) to wait for checks to complete.',
     }),
     verbose: Flags.boolean({
--- a/packages/cli/src/services/abstract-check-runner.ts
+++ b/packages/cli/src/services/abstract-check-runner.ts
@@ -32,6 +32,8 @@ export type RunLocation = PublicRunLocation | PrivateRunLocation
 
 export type CheckRunId = string
 
+export const DEFAULT_CHECK_RUN_TIMEOUT_SECONDS = 300
+
 export default abstract class AbstractCheckRunner extends EventEmitter {
   checks: Map<CheckRunId, { check: any, testResultId?: string }>
   testSessionId?: string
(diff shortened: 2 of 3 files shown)


packages/cli/src/commands/test.ts
```diff
@@ -56,7 +62,7 @@ export default class Test extends AuthCommand {
       description: 'list all checks but don\'t run them.',
     }),
     timeout: Flags.integer({
-      default: 300,
+      default: DEFAULT_CHECK_RUN_TIMEOUT_SECONDS,
```
nahuelon, line 65: @clample , if Checkly should always report a result within 240s, is it possible to limit users to configure a value lower than `240`?

packages/cli/src/services/abstract-check-runner.ts
```diff
@@ -173,7 +175,13 @@ export default abstract class AbstractCheckRunner extends EventEmitter {
     Array.from(this.checks.entries()).forEach(([checkRunId, { check }]) =>
       this.timeouts.set(checkRunId, setTimeout(() => {
         this.timeouts.delete(checkRunId)
-        this.emit(Events.CHECK_FAILED, check, `Reached timeout of ${this.timeout} seconds waiting for check result.`)
+        let errorMessage = `Reached timeout of ${this.timeout} seconds waiting for check result.`
+        // Checkly should always report a result within 240s.
+        // If the default timeout was used, we should point the user to the status page and support email.
+        if (this.timeout === DEFAULT_CHECK_RUN_TIMEOUT_SECONDS) {
```
nahuelon, line 181: Related with previous comment: if the user set `--timeout=300`, in case the check doesn't `FINISH`, he will get the `Checkly may be experiencing problems.` message.

nahuelon (approved):
LGTM (but I left a comment)

packages/cli/src/commands/test.ts
clample: I think that we should continue letting users set any timeout that they'd like. It may make sense for their setup to set a shorter timeout - for example, if they only have API checks which don't take so long.

I think that the default of 300s makes sense, though. This is long enough to make sure that the check runs should be completed, plus a buffer for any latency in sending over the check result. This timeout should only ever be reached if checkly is experiencing issues.

packages/cli/src/services/abstract-check-runner.ts
clample: If a check takes longer than 240s, the Checkly backend should normally be timing out/cancelling the check run and reporting a result to the CLI. This timeout in the CLI is mainly a fallback. If the CLI doesn't receive a result in 300 seconds, then it's likely because something went wrong on the Checkly side.

PR #713, opened 2023-05-10
feat: create git-ignore if does not exists [sc-15241]

maxigimenez:

## Affected Components
* [ ] CLI
* [x] Create CLI
* [ ] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer
Allow the user to generate a `.gitignore` if none were present during the initialization of an existing project

### DIFF
--- a/packages/cli/e2e/__tests__/test.spec.ts
+++ b/packages/cli/e2e/__tests__/test.spec.ts
@@ -130,4 +130,15 @@ describe('test', () => {
     expect(fs.existsSync(path.join(__dirname, 'fixtures', 'test-project', reportFilename))).toBe(true)
     expect(result.status).toBe(0)
   })
+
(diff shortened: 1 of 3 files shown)


ndom91 (commented):
Maybe we just add this `.gitignore` as part of the "would you like us to initialize a git repo for you?" question. Instead of adding a whole new question in the wizard?

maxigimenez:
@ndom91 @clample I've thinking about it after Nico's comment, and the best flow would be to ask if wanna initialize git, if so, we generate the `.git` and the `.gitignore` both for new and existing projects. wdyt?

clample:
@maxigimenez I think that makes sense 👍

clample (approved):

ndom91 (approved):
LGTM :+1:

PR #715, opened 2023-05-12
feat: support top level await in browser check scripts

clample:

## Affected Components
* [x] CLI
* [ ] Create CLI
* [ ] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer
Checks created in Checkly via the UI already support top-level `await`. For example, the following check runs successfully when created from the UI:
```
const sleep = ms => new Promise(r => setTimeout(r, ms));
await sleep(10000);
```

The CLI currently fails to parse checks with top-level `await`, though. It fails with the following error:
```
    Encountered an error parsing check files for /Users/clample/projects/checkly-cli/packages/cli/src/services/check-parser/__tests__/check-parser-fixtures/top-level-await.js.

    The following NPM dependencies were used, but aren't supported in the runtimes.
    For more information, see https://www.checklyhq.com/docs/runtimes/.
        /Users/clample/projects/checkly-cli/packages/cli/src/services/check-parser/__tests__/check-parser-fixtures/top-level-await.js imports unsupported dependencies:
```

This PR fixes the issue by simply adding an additional flag to our javascript parser (docs [here](https://github.com/acornjs/acorn/tree/master/acorn/#interface)). This should make the CLI and the UI more consistent, and it's useful for migrating internal checks we have at Checkly to MaC.

### DIFF
--- a/packages/cli/src/services/check-parser/__tests__/check-parser-fixtures/top-level-await.js
+++ b/packages/cli/src/services/check-parser/__tests__/check-parser-fixtures/top-level-await.js
@@ -1,0 +1,2 @@
+const sleep = ms => new Promise(resolve => setTimeout(resolve, ms))
+await sleep(10000)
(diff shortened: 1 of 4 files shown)


nahuelon (approved):
LGTM.

PR #717, opened 2023-05-16
feat: enhance empty script or env var value error [gh-714]

### LINKED ISSUE #714: bug: no actionable error message given for empty browser check script and other validation issues
### Node.js version

v16.17.0

### NPM version

v8.15.0

### @checkly/cli version

v4.0.5

### Steps to reproduce

Create a browser check with an empty script. This can be done in the boilerplate example by deleting the contents of `__checks__/homepage.spec.ts`, for example. Then run `npx checkly test` or `npx checkly deploy`.

### What is expected?

There should be a clear error message that `__checks__/homepage.spec.ts` is empty. Or, the check could just pass.

This results in the following error message from the backend:
```
$ npx checkly test
Parsing your project... done

Unable to run checks: "checkRunJobs[1].script" is not allowed to be empty
```

With many checks, it will be very challenging for users to debug errors like this. Rather than `checkRunJobs[<some index>]`, we should add the items logical ID.

### Any additional comments?

We should add special error handling for the empty file case. 

At the same time, though, there will always be the chance that the backend reports a validation error that the CLI didn't detect. We should also handle the more general case of a backend validation error, and make sure that the logical ID is reported rather than just an index.

nahuelon:

## Affected Components
* [x] CLI
* [ ] Create CLI
* [x] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer
The PR adds a check for browser-check script payloads and check-group environment variable values during the project parsing. If there is an empty string, it throws a clear error message indicating the `logicalId` and/or environment variable.

image

image

### DIFF
--- a/packages/cli/src/constructs/check-group.ts
+++ b/packages/cli/src/constructs/check-group.ts
@@ -133,7 +133,15 @@ export class CheckGroup extends Construct {
     this.privateLocations = props.privateLocations
     this.concurrency = props.concurrency
     this.apiCheckDefaults = { ...defaultApiCheckDefaults, ...props.apiCheckDefaults }
+
     this.environmentVariables = props.environmentVariables ?? []
+    this.environmentVariables.forEach(ev => {
+      // only empty string is checked because the KeyValuePair.value doesn't allow undefined or null.
+      if (ev.value === '') {
+        throw new Error(`Environment variable "${ev.key}" from check group "${logicalId}" is not allowed to be empty`)
+      }
+    })
+
     this.alertChannels = props.alertChannels ?? []
     this.localSetupScript = props.localSetupScript
     this.localTearDownScript = props.localTearDownScript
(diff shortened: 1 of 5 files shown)


packages/cli/src/constructs/check-group.ts
```diff
@@ -133,7 +133,15 @@ export class CheckGroup extends Construct {
     this.privateLocations = props.privateLocations
     this.concurrency = props.concurrency
     this.apiCheckDefaults = { ...defaultApiCheckDefaults, ...props.apiCheckDefaults }
+
     this.environmentVariables = props.environmentVariables ?? []
+    this.environmentVariables.forEach(ev => {
+      // only empty string is checked because the KeyValuePair.value doesn't allow undefined or null.
+      if (ev.value === '') {
```
clample, line 140: `value` could also be null or undefined. Do you know if our Joi validation throws an error in that case?

clample: I looked, and it seems like `null` is allowed. I wonder if we should just allow empty strings in the API.

clample (approved):

packages/cli/src/constructs/check-group.ts
nahuelon: @clample , `null` is not allowed to be used a `KeyValuePair.value`, just `string`. So, the only 'falsy' value is the empty string `''`.

PR #720, opened 2023-05-17
fix: replace inquirer with prompts [gh-689]

### LINKED ISSUE #689: bug: navigating up and down in the account list hides parts of the accounts list
### Node.js version

v16.17.0

### NPM version

8.15.0

### @checkly/cli version

checkly/4.0.3 darwin-arm64 node-v16.17.0

### Steps to reproduce

- In the terminal enter `npx checkly switch`. The list of account linked to the user is shown.
- Navigate from one account to the next on the list using arrow keys.
- If there are more than 2 accounts the list is suddenly hidden when navigating to the third account on the list.

### What is expected?

The full list of account show be visible at all times. 

The list is fully hidden when the user is navigating to the third list item. 

### Any additional comments?

_No response_

nahuelon:

## Affected Components
* [x] CLI
* [ ] Create CLI
* [ ] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer
We are using [prompts](https://www.npmjs.com/package/prompts) in `create-cli` package, and migrating from [inquirer](https://www.npmjs.com/package/inquirer) to [prompts](https://www.npmjs.com/package/prompts) solved the issue.
The PR changes all the commands prompts (`confirm`, `select` and `text`) to use the new package.

> Resolves #689


clample:
@nahuelon, the diff to `package-lock.json` seems way larger than I would expect for changing one package. Do you have any idea what it might be? 

Could you try resetting it with:
```
$ nvm use
^ should pull in node 16
$ npm cache clean --force
$ git checkout main package-lock.json
^ reset the package-lock.json
$ npm install --workspaces
```

packages/cli/e2e/__tests__/destroy.spec.ts
```diff
@@ -30,7 +30,6 @@ describe('destroy', () => {
       directory: path.join(__dirname, 'fixtures', 'deploy-project'),
       env: { PROJECT_LOGICAL_ID: projectLogicalId },
     })
-    expect(result.status).not.toBe(0)
```
clample, line 33: Does the test fail with this assertion? Normally `npx checkly destroy` won't get the confirmation prompt and the command should timeout. In that case, the exit code should be `null` and the test should pass.

clample:
Looks good to me aside from the package-lock.json 👍

nahuelon:
> @nahuelon, the diff to `package-lock.json` seems way larger than I would expect for changing one package. Do you have any idea what it might be?
> 
> Could you please try resetting it with:
> 
> ```
> $ nvm use
> ^ should pull in node 16
> $ npm cache clean --force
> $ git checkout main package-lock.json
> ^ reset the package-lock.json
> $ npm install --workspaces
> ```

@clample , good catch. I was able to fix it by cleaning the cache and reinstalling all the packages again. Thanks!

packages/cli/e2e/__tests__/destroy.spec.ts
clample: It would be nice to replace this with some other assertion that the timeout is reached. I think that that would help make sure `checkly destroy` keeps working as expected in the future.

clample: I looked and I think it's because prompts is capturing the interrupts and calling this `onCancel` callback https://github.com/terkelg/prompts/issues/27

Issue #516, opened 2023-02-14
story: Improve the CI/CD flow of the cli


umutuzgur:
### Story description
We want to improve the CI/CD workflow of the cli. This involves lint checks, easier releases and auto release message creation using conventional commits.

We can use eslint with simple git-hooks for lint checks. Package versioning and release messages can be achieved with sematic-release. If there is anything better, I'm also open to suggestions

- [x] Linter, local and CI - https://github.com/checkly/checkly-cli/pull/731
- [x] Automatic package versioning upon github releases - https://github.com/checkly/checkly-cli/pull/769
- [x] Automatic release message creation - https://github.com/checkly/checkly-cli/pull/769

nahuelon:
> Just my two cents fro the local lint - I think that having a lint in CI is enough to keep `main` tidy. I like to regularly commit in-progress work, and I wonder if a git lint hook will block that.
> 
> Maybe there's a way to disable the local linting, though.

@clample , if I understood right, you want to commit changes that don't respect lint rules. In that case, you can use `git commit --no-verify` where the `--no-verify` flag tells `simple-git-hooks` to bypass all the checks. For example:

![image](https://github.com/checkly/checkly-cli/assets/5315705/8fb34e35-c17b-4b51-9832-e87fe3c8aaf5)

clample:
Thanks @nahuelon for the workaround 👍

PR #731, opened 2023-05-24
feat: add git hooks and lint staged [gh-516]

nahuelon:

## Affected Components
* [x] CLI
* [x] Create CLI
* [x] Test
* [x] Docs
* [x] Examples
* [x] Other

## Notes for the Reviewer
This PR modifies `cli` and `create-cli` packages adding `lint-staged`, `commitlint`, and `simple-git-hooks` hooks packages in the monorepo `package.json`. Some adjustments were applied to follow a monorepo project structure.

1. remove `examples/` from the workspaces lists
2. remove duplicated configuration for `eslint` (`package.json` section)
3. centralize linting and hooks by moving `eslint` and `simple-git-hooks` related packages from each workspace to the monorepo project
4. install the `simple-git-hooks` and `lint-staged` packages as `devDependencies` (and `prepare` script) within the monorepo project
5. adjust `CONTRIBUTING.md` docs.
6. fix linting small issues for `fixtures` that can compliance the linting rules, and disable `eslint` for files that are used to test 'syntax-errors'. Also, the are some fixtures files with some eslint rules disabled (like console.log, required, no-new
7. fix linting for `/examples` folder
8. adjust Github Actions to use monorepo project

Lot of files were modified (small changes) because I had to compliance all the linting rules before committing. This results in a very cleaned up source code.

All this changes are part of https://github.com/checkly/checkly-cli/issues/516 tasks.


packages/cli/constructs.js
```diff
@@ -1 +1,3 @@
-module.exports = require('./dist').constructs
+import { constructs } from './dist'
```
umutuzgur, line 1: I think this has to stay the way it was, or?

commitlint.config.js
```diff
@@ -0,0 +1,12 @@
+module.exports = {
+  extends: ['@commitlint/config-conventional'],
```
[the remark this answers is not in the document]
umutuzgur: Gotcha, my mistake

packages/cli/constructs.js
nahuelon: I think we can use the `import`, but I'm trying it locally with different type of projects.

nahuelon:
@umutuzgur , I adjusted all the previous linting configuration to use a monorepo approach and take advantage of its benefits. Now, dependency packages were removed from each project (`cli` and `create-cli`) and added to the monorepo project. I updated the PR description adding all changes the PR includes.

packages/cli/constructs.js
umutuzgur: Because this is a module type

nahuelon: I'll revert this change for now.

packages/create-cli/src/actions/template.ts
```diff
@@ -22,7 +22,7 @@ export async function copyTemplate ({ template, templatePath, targetDir }: CopyT
       downloadTemplateSpinner.text = chalk.red(`Couldn't find template "${template}"`)
       downloadTemplateSpinner.fail()
     } else {
-      console.error(e.message)
+      process.stderr.write(e.message)
```
umutuzgur, line 25: There is a difference between console and stderr.write. console operations always add a new line whereas stderr.write doesn't

umutuzgur (approved):
LGTM

PR #737, opened 2023-05-29
chore: remove login command flags [gh-588]

### LINKED ISSUE #588: story: move CHECKLY prefixed envs to `AuthCommand` base class so they are documented
### Story description

We have 2 special env vars for bypassing the auth flow which we use heavily with the CI use case. We are not using the oclif framework to set this env vars which then causes it not to be printed when a user invokes the `--help` command. Instead, we should move these to the `AuthCommand` so they are documented and easier to be discovered by the user

### Shortcut link

_No response_

### Additional resources

_No response_

nahuelon:

## Affected Components
* [x] CLI
* [ ] Create CLI
* [ ] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer
After trying multiple approaches to make the env variables available for each `AuthCommand`, I decided to add a new section into the help information.

image

I tried to add the env variables to the flags but the `--help` doesn't show any environment variable description. I found someone that is asking for that feature in https://github.com/oclif/oclif/issues/691
Also, as we discussed in https://github.com/checkly/checkly-cli/issues/588 , the account Id and API key flags shouldn't be available for any auth command to avoid credentials exposures.

This PR removes the `oclif/help` because it can't be configure to use a different class than the `oclif` `Help` class, making the `help` output differ with `--help` output. I simply created a new `help` command (this is what the plugin does under the hood) to use our help class `ChecklyHelpClass`.

> Resolves #588


tnolet:
@umutuzgur @nahuelon I don't know what we are doing in this PR.
- setting the credentials in CI / non-interactive is clearly described in the docs. Just like all other CLI frameworks do.
- I see no reason for us adding this to our help menu as it's a set & forget thing. 
- I have not encountered one user who did not understand this pattern.
I think we can completely delete this PR as to me it just complicates things.

tnolet:
Ok, then please remove the `CHECKLY_ACCOUNT_NAME`. It should not have been there in the first place.

packages/cli/src/help/help-extension.ts
```diff
@@ -55,8 +55,15 @@ export default class ChecklyHelpClass extends Help {
       this.log('')
     }
 
+    this.log(this.section('ENVIRONMENT VARIABLES',
+    `Running the CLI from your CI pipeline will need to export variables in the shell:\n
+  CHECKLY_ACCOUNT_ID    Checkly account ID.
+  CHECKLY_API_KEY       Checkly User API Key.
+  CHECKLY_ACCOUNT_NAME  (optional) Checkly account Name you want to use.
```
umutuzgur, line 62: Let's remove `CHECKLY_ACCOUNT_NAME`. That is only for local development so not needed for the user help

nahuelon: Yes. I misunderstood why the `CHECKLY_ACCOUNT_NAME` is used. I'll remove it.

packages/cli/src/commands/login.ts
```diff
@@ -66,23 +35,24 @@ export default class Login extends BaseCommand {
...
+    }
 
-    if (apiKey) {
-      if (!accountId) {
-        throw new Error('The flag --account-id (-i) is required when using --api-key (-k)')
-      }
+    const { accountName } = await prompts({
+      name: 'accountName',
+      type: 'select',
+      choices: accounts.map(({ name }) => ({ title: name, value: name })),
```
umutuzgur, line 46: We can simplify this further in a new PR and just pass the whole account as value instead of just the name

packages/cli/src/commands/help.ts
```diff
@@ -0,0 +1,12 @@
+import { BaseCommand } from './baseCommand'
+import ChecklyHelpClass from '../help/help-extension'
+
+export default class Logout extends BaseCommand {
+  static hidden = false
+  static description = 'Display help for <%= config.bin %>.'
+
+  async run (): Promise<void> {
+    const help = new ChecklyHelpClass(this.config)
```
umutuzgur, line 9: Nice solution!

umutuzgur (approved):
LGTM

PR #747, opened 2023-06-01
feat: add PrivateLocation construct [gh-736]

### LINKED ISSUE #736: story: add `PrivateLocation` construct
### Story description

Users should be able to also manage their Private Locations "as code" from the CLI using a construct

### Shortcut link

_No response_

### Additional resources

_No response_

nahuelon:

## Affected Components
* [x] CLI
* [ ] Create CLI
* [x] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer
- add `PrivateLocation` resource construct.
- add error handling for `destroy` command (now we could get an error destroying a project: the only possible error is when you are trying to remove a `PrivateLocation` created using the CLI having external checks/groups assigned through the UI.
- fetch private-locations using 'singleton' approach `Session.getPrivateLocations()`
- add `CHECKLY_CLI_VESION` environment variable for testing purpose, to allow us to override the CLI version when running E2E test.

`Check` and `CheckGroup` `privateLocations` properties are parsed to filter slug name strings and use them to get the private-location from the account and create an instance similar to use `fromId()`. After creating the non member PL resources, it adds the assignment resources for Check and Groups.

As all the PL and assignments are transformed to resources, the `privateLocations` property is not sent to the API after this version is released. Only previous CLI versions still send the property and that will use the old PL assignment flow (using the backend `fetchPrivateLocations()` and `upsertGraph()` functions) without supporting PL creation/reference in the CLI.

> **Note 1**: The new resource include `fromId()` instantiation used in the CLI when processing the slug names string and transform to `PrivateLocation` resource instances

> **Note 2**: The `slugName` cannot be modified after the private-location is created. We have to discuss how can let the user know about this or how to restrict the modification (maybe throw some error if we detect a `slugName` modification for the same private-location.

> **Note 3**: After creating a private-location, the raw-key is not returned. We have to discuss how to tell the user to access the UI and regenerate a new key.

> Depends on https://github.com/checkly/checkly-backend/pull/4249

> Resolves #736


packages/cli/src/constructs/private-location.ts
```diff
@@ -0,0 +1,80 @@
+import { Construct } from './construct'
+import { Session } from './project'
+import { ValidationError } from './validator-error'
+
+export interface PrivateLocationProps {
```
umutuzgur, line 5: Don't forget to add the `icon` field as well. We can add the string literals later on of course

packages/cli/src/constructs/private-location.ts
```diff
@@ -0,0 +1,80 @@
...
+      .test(this.slugName)) {
+      throw new ValidationError(`The "slugName" must differ from all AWS locations. (slugName='${this.slugName}')`)
+    }
+
+    Session.registerConstruct(this)
+  }
+
+  static fromId (id: string|number) {
+    return new PrivateLocationWrapper(`private-location-${id}`, id)
+  }
```
umutuzgur, line 67: Conflicts with `Note 1` in the PR description

nahuelon:
@clample , the PR is already adjusted and with some extra test included. It would be great if we can merge this PR today. I appreciate if you can review it during the day. THANKS!

umutuzgur (approved):

PR #751, opened 2023-06-05
feat: add scheduling messages for a better DX when checks take more t…

Antoine-C:
…ime to be scheduled [sc-16052]

## Affected Components
* [x] CLI
* [ ] Create CLI
* [ ] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer

- We now track which check is still being scheduling and which check is running by relying on the `CHECK_INPROGRESS` and `RUN_START` events.
- When a test start checks are flagged as `SCHEDULING` and after we receive the `CHECK_INPROGRESS` event for a specific check we flag it as `PENDING`.
- After 20s if some checks are still being scheduled we add a message in the summary to let the user know it can take up to two minutes.

https://github.com/checkly/checkly-cli/assets/3541876/d155a203-a44d-4721-95b9-6b8b20a651bc


packages/cli/src/messages/scheduling-messages.ts
```diff
@@ -0,0 +1,7 @@
+const schedulingMessages = {
+  provisioningStarted: 'We are provisioning resources for you',
+  provisioningInProgress: 'Thanks for waiting, the more you use this region, the faster it will be',
```
clample, line 3: > the more you use this region, the faster it will be

I think that this is a bit confusing and we can leave it out

clample (commented):
I'm not convinced about adding the pre-timed messages at 20/30/50 seconds. As a user, I don't know if I would like to see this. 

Do we know how often and how much scheduling delay we're seeing in practice? If it's regularly taking ~60 seconds, then I think that we could just print a single message right away like "Preparing to start checks. This may take up to a minute."

I also think that it could be nice to do something with the `CHECK_INPROGRESS` hook. Currently we just have the `-` icon while waiting for a check to start. Instead, we could start doing a different icon depending on whether the check is waiting to be scheduled or whether it's actually in-progress and running. This would give more information to users about if their checks are taking  a long time, or if it's a scheduling issue.

Antoine-C:
> I'm not convinced about adding the pre-timed messages at 20/30/50 seconds. As a user, I don't know if I would like to see this.
> 
> Do we know how often and how much scheduling delay we're seeing in practice? If it's regularly taking ~60 seconds, then I think that we could just print a single message right away like "Preparing to start checks. This may take up to a minute."
> 
> I also think that it could be nice to do something with the `CHECK_INPROGRESS` hook. Currently we just have the `-` icon while waiting for a check to start. Instead, we could start doing a different icon depending on whether the check is waiting to be scheduled or whether it's actually in-progress and running. This would give more information to users about if their checks are taking a long time, or if it's a scheduling issue.

Right now I think 99% of the time we have a negligible scheduling delay but in some rare cases it can take up to 60 seconds. From a discussion on Slack we wanted to provide something to let the user know that we are provisioning resources and it can take time.

I like the icon idea but I'm afraid that after 10/20s+ waiting with this icon the user thinks his checks are stuck and cancel its run. I don't really have a strong opinion on what we should do DX wise to have something meaningful and clear, the icon feels more "cli native" than the messages but is it clear enough? 

@jan-osch wdyt?

Antoine-C:
> > Right now I think 99% of the time we have a negligible scheduling delay but in some rare cases it can take up to 60 seconds. From a discussion on Slack we wanted to provide something to let the user know that we are provisioning resources and it can take time.
> 
> > I like the icon idea but I'm afraid that after 10/20s+ waiting with this icon the user thinks his checks are stuck and cancel its run. I don't really have a strong opinion on what we should do DX wise to have something meaningful and clear, the icon feels more "cli native" than the messages but is it clear enough?
> 
> Thanks for the extra context. If it's an edge case, I agree - it might be good to have the message be shown with a delay.
> 
> One edge case that makes it tricky - some checks might be scheduled right away, while other checks could take a long time. It looks like the PR currently cancels the message when a single check is scheduled with `CHECK_INPROGRESS`, but other checks might still be hanging.
> 
> What do you think about:
> 
> * Add a separate icon for scheduling vs running (rather than the current solution of always showing `-`)
> * After 20 seconds of some checks not being scheduled, show a message in the status bar "Still waiting to schedule some checks. This may take a minute or two."
> * After all checks are scheduled (if the message was displayed), replace the message with "All checks are running"

Indeed that's one valid edge case, I wanted to make something really simple at first to prevent potential frustration but I like what you're suggesting. 
@jan-osch wdyt about Chris idea? I like it personally, is it okay also product wise?

jan-osch:
@Antoine-C how much extra time will it take to add the in progress state to the test list output? Can you record a short video of running a lot of checks in a cold region?

Antoine-C:
> @Antoine-C how much extra time will it take to add the in progress state to the test list output? Can you record a short video of running a lot of checks in a cold region?

Running 80 checks in **ap-northeast-3**. It gets a bit messy in the terminal I think to use a new loading icon before each check IMO, but I'm also surprised at how fast it was to scale or at least execute all tests. Maybe the region was not cold enough but now I'm really second guessing the need for this PR right now?

https://github.com/checkly/checkly-cli/assets/3541876/37e54225-4dbf-4a44-9905-4965474c1046

clample:
> > @Antoine-C how much extra time will it take to add the in progress state to the test list output? Can you record a short video of running a lot of checks in a cold region?
> 
> Running 80 checks in **ap-northeast-3**. It gets a bit messy in the terminal I think to use a new loading icon before each check IMO, but I'm also surprised at how fast it was to scale or at least execute all tests. Maybe the region was not cold enough but I'm second guessing the need for this PR right now?
> Enregistrement.de.l.ecran.2023-06-13.a.10.52.09.mov

How long are the checks? From the video, it looks like all of the checks are very short. It also looks like they're not being run fully in parallel - they seem to be running just a few at a time.

My suspicion is that if we bump the execution time to 30 seconds, then the checks would clog up ECS quite a bit more.

Antoine-C:
> > > @Antoine-C how much extra time will it take to add the in progress state to the test list output? Can you record a short video of running a lot of checks in a cold region?
> > 
> > 
> > Running 80 checks in **ap-northeast-3**. It gets a bit messy in the terminal I think to use a new loading icon before each check IMO, but I'm also surprised at how fast it was to scale or at least execute all tests. Maybe the region was not cold enough but I'm second guessing the need for this PR right now?
> > Enregistrement.de.l.ecran.2023-06-13.a.10.52.09.mov
> 
> How long are the checks? From the video, it looks like all of the checks are very short. It also looks like they're not being run fully in parallel - they seem to be running just a few at a time.
> 
> My suspicion is that if we bump the execution time to 30 seconds, then the checks would clog up ECS quite a bit more.

Ah yes you're right, let me try it again.

Antoine-C:
> 

Thanks for the video, I will add the new icon then for each check while they're scheduled + the message after 20 seconds if they're still being scheduled in the status bar 👍

Antoine-C:
@clample Sorry for the spam, but what do you call the status bar?

clample:
> @clample Sorry for the spam, but what do you call the status bar?

@Antoine-C no worries! I'll DM you

clample:
@Antoine-C - just watched the video, this looks so cool

packages/cli/src/reporters/abstract-list.ts
```diff
@@ -94,16 +108,20 @@ export default abstract class AbstractListReporter implements Reporter {
...
       }
     }
+
     if (!opts.skipCheckCount) {
       status.push('')
       status.push([
+        counts.scheduling ? chalk.bold.blue(`${counts.scheduling} scheduling`) : undefined,
         counts.numFailed ? chalk.bold.red(`${counts.numFailed} failed`) : undefined,
         counts.numPassed ? chalk.bold.green(`${counts.numPassed} passed`) : undefined,
         counts.numPending ? chalk.bold.magenta(`${counts.numPending} pending`) : undefined,
```
clample, line 118: I think that we could change `pending` to `in-progress` or `running` now. It might make the difference between this category and `scheduling` more clear for users.

packages/cli/src/reporters/abstract-list.ts
```diff
@@ -94,16 +108,20 @@ export default abstract class AbstractListReporter implements Reporter {
         status.push(sourceFile ? indentString(titleString, 2) : titleString)
       }
     }
+
     if (!opts.skipCheckCount) {
       status.push('')
       status.push([
+        counts.scheduling ? chalk.bold.blue(`${counts.scheduling} scheduling`) : undefined,
```
clample, line 115: Super nit: I think that it could be nice to have the `scheduling` count right near the `pending` count. Scheduling and pending are kind of similar, so I think it's nice to have them side by side.

clample:
Just left a few small style comments. Other than that, it looks good to me.

packages/cli/src/reporters/abstract-list.ts
```diff
@@ -94,16 +108,24 @@ export default abstract class AbstractListReporter implements Reporter {
...
+        counts.numPending ? chalk.bold.magenta(`${counts.numPending} pending`) : undefined,
```
clample, line 116: I think that this could be "running"

PR #755, opened 2023-06-07
feat: display testOnly checks in deployment summary

clample:

## Affected Components
* [x] CLI
* [ ] Create CLI
* [ ] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer
Currently checks with `testOnly=true` aren't displayed at all in the deployment summary of `npx checkly deploy --preview` or `npx checkly deploy --output`. To give more info to users, this PR adds a "Skip" section to the deployment summary containing the `testOnly` checks.

There's an edge case where a check was deployed, then in a subsequent deployment the check is marked as `testOnly=true`. In this case, the check should be considered deleted rather than skipped, since it will be removed from Checkly. I added handling for this edge case and an integration test.

Choosing a color is tricky. Depending on the theme, it isn't super clear. I think that the grey is OK for now, though.
Example (dark theme):
Screenshot 2023-06-07 at 11 09 30

Example (light theme):
Screenshot 2023-06-07 at 11 08 53

thanks @stefanjudis for the suggestion 🦝

### DIFF
--- a/packages/cli/e2e/__tests__/deploy.spec.ts
+++ b/packages/cli/e2e/__tests__/deploy.spec.ts
@@ -70,17 +70,51 @@ describe('deploy', () => {
     expect(status).toBe(0)
   })
 
-  it('Shouldn\'t include a testOnly check', () => {
+  it('Should mark testOnly check as skipped', () => {
     const { status, stdout } = runChecklyCli({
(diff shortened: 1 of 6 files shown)


umutuzgur (approved):
LGTM

PR #757, opened 2023-06-07
feat: add improved error message when no constructs found for deploy

clample:

## Affected Components
* [x] CLI
* [ ] Create CLI
* [ ] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer
This PR attempts to improve the handling of `npx checkly deploy` for the case where no constructs are found. 

Currently in `main`, the deployment API is called and there's an error during the backend validation:
image
The CLI also asks for a deployment confirmation - even though the deploy is guaranteed to fail. The same result happens when `--preview` is added.

With this PR, there's an updated error message. The CLI also doesn't bother asking for a deployment confirmation, since it can detect early that the deploy will fail.
image

Screenshot 2023-06-07 at 12 07 01


umutuzgur (approved):
Nice addition 👍

PR #639, opened 2023-04-12
feat: adds autocomplete + suggestions. Tweaks small loading messages …

tnolet:
…for quicker interaction.

I hereby confirm that I followed the code guidelines found at [engineering guidelines](https://www.notion.so/checkly/Engineering-Guidelines-a3a165a203a04dc1a112f0e26b2f2d3f)

## Affected Components
* [x] CLI
* [ ] Create CLI
* [ ] Test
* [ ] Docs
* [x] Examples
* [ ] Other

## Notes for the Reviewer
- adds autocomplete and suggestion plugins. Test locally and works as described.
- adds a small progress messages to `test` and `deploy` commands as project parsing can take a second or two.

Resolves #637 
Resolves #625 

## New Dependency Submission
- "@oclif/plugin-not-found"
-  "@oclif/plugin-autocomplete"

PR #758, opened 2023-06-07
feat: remove autocomplete from the CLI

clample:

## Affected Components
* [x] CLI
* [ ] Create CLI
* [ ] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer
This PR removes the `npx checkly autocomplete` command introduced in https://github.com/checkly/checkly-cli/pull/639. 

Autocomplete was problematic because it requires a global installation of `checkly`. For implementation reasons, though, the NPM library used for the CLI checkly command needs to be the same NPM library used for importing constructs in the `.check.ts` files. For this reason, we should discourage using a global install of `checkly` for the CLI component.

In particular, we need to be able to pass some state between the CLI and the .check.ts files. To make it a bit easier to write .check.ts files, we use a bit of magic with static variables in a `Session` class ([here](https://github.com/checkly/checkly-cli/blob/ceb14592ac29e0357e8af16f83e19d86eb6d4f06/packages/cli/src/constructs/project.ts#L116-L124)).

We can revisit autocomplete and improving support for a global installation of `checkly` in the future.


umutuzgur (approved):

clample:
Sorry for the re-approval request. Needed to tweak a test.

umutuzgur (approved):

stefanjudis:
Thanks for taking care of this!

PR #759, opened 2023-06-07
feat: add Dashboard construct [gh-734]

### LINKED ISSUE #734: story: add `Dashboard` construct.
### Story description

Users should be able to also manage their Public Dashboards "as code" from the CLI using a construct

### Shortcut link

_No response_

### Additional resources

_No response_

nahuelon:

## Affected Components
* [x] CLI
* [ ] Create CLI
* [x] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer
Add `Dashboard` resource construct. It allows to use a CSS file with `customCSS.entrypoint`.

Note: The new resource doesn't include `fromId()` instantiation, since it cannot be referenced from other resources.

A dashboard was added to the E2E fixtures.

Small fix for `maintenance-windows` constructor, disallowing to use it through the Config.

> Resolves #734

## New Dependency Submission

> Depends on https://github.com/checkly/checkly-backend/pull/4268


packages/cli/src/constructs/dashboard.ts
```diff
@@ -0,0 +1,212 @@
...
+    if (props.customCSS) {
+      if ('entrypoint' in props.customCSS) {
+        if (!fs.existsSync(props.customCSS.entrypoint)) {
+          throw new Error(`Unrecognized CSS code for 'customCSS' property in dashboard '${logicalId}'.`)
+        }
+        this.customCSS = String(fs.readFileSync(props.customCSS.entrypoint))
+      } else if ('content' in props.customCSS) {
+        this.customCSS = props.customCSS.content
+      }
+    }
```
umutuzgur, line 177: Could you check how we do this in `browser-check.ts` with relative paths and implement it the same way so there is no difference?

packages/cli/src/constructs/dashboard.ts
```diff
@@ -0,0 +1,212 @@
...
+
+    Session.registerConstruct(this)
+  }
+
+  allowInChecklyConfig () {
```
umutuzgur, line 182: This should be false btw or just removed given it is false in `construct.ts`. The same thing applies to maintenance windows. I see that it is also true there

packages/cli/src/constructs/dashboard.ts
nahuelon: @umutuzgur , I checked it but then I saw the setup/tearDown script are different than the browser checks. So, I followed the same pattern for scripts. I'll use the browser checks files way now.

packages/cli/src/constructs/dashboard.ts
nahuelon: Right! Thanks for find that.

packages/cli/src/constructs/dashboard.ts
umutuzgur: You must have the older version. `api-check.ts` does also something similar atm

umutuzgur (approved):
LGTM

umutuzgur (approved):

PR #767, opened 2023-06-20
feat: propagate repo info on checkly deploy

clample:

## Affected Components
* [x] CLI
* [ ] Create CLI
* [ ] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer
This PR adds git info the `checkly deploy` command. This additional git info will then be displayed in the resource history log. The same method is used to fetch the git info that we're already using for `checkly test`.

I thought that the `CHECKLY_TEST_` naming format is now inaccurate for the git info env vars. This PR also updates these env var names, in a backwards compatible way.


umutuzgur (approved):
LGTM

PR #769, opened 2023-06-22
feat: automatic versioning when release published [gh-516]

nahuelon:

## Affected Components
* [ ] CLI
* [ ] Create CLI
* [ ] Test
* [ ] Docs
* [ ] Examples
* [x] Other

## Notes for the Reviewer
This PR modifies the release Github Action workflow adding:
- automatic versioning using the tag (`v#.#.#`) from the published release
- split `prerelease` and `release` jobs using `production` environment, so we can review the version before continuing with the latest version publish action
~- add `auto-changelog` NPM package dependency to automatically generate/update `CHANGELOG.md`~
- release&publish `create-cli` on every `cli` release using the same version number (from the release tag)
- remove `create-cli` version fetching from the NPM registry. Package version will be synced with the `cli` package, so we can use the `oclif` version.
- update `README.md` and `CONTRIBUTING.md` release docs.

> TODO 1: create the `production` environment in the Github repository

> ~TODO 2: test this workflow in a dummy forked project using a dummy NPM package name~

~> TODO 3: create Personal Access Token for push back changes to `main` (`package.json,` `package-lock.json` and `CHANGELOG.md` files). Need to discuss if we let the user to `--force` the push.~

> TODO 4: define release description format (file `.github/release.yml`)

~> TODO 5: decide if we are gonna add an automatic workflow for the `create-cli` package too.~


.github/workflows/release.yml
```diff
@@ -1,23 +1,90 @@
...
+      - name: Append changes to changelog
+        run: |
+          npx auto-changelog --latest-version ${{ github.event.release.tag_name }}
+          git add .
+      # Commit and push all changes to main
+      - name: Git push new package version
+        run: |
+          git commit -m "feat: release CLI ${{ github.event.release.tag_name }}"
+          git push origin main --follow-tags --force
+      - name: Output version change after push
```
umutuzgur, line 42: Why are we creating another tag here? We already created one using the Github UI anyways at this point, or?

nahuelon: I'm gonna use the release tag without pushing back the `package.json` to the main branch. The version will be something like `0.0.1` in the `package.json`, or a un-synced version number comparing with the NPM version.

packages/cli/src/commands/baseCommand.ts
```diff
@@ -9,9 +19,9 @@ export abstract class BaseCommand extends Command {
   static coreCommand = false
   static hidden = true
 
-  protected init (): Promise<void> {
+  protected async init (): Promise<void> {
     // TODO: find a better way to mock the CLI version for E2E tests.
-    api.defaults.headers['x-checkly-cli-version'] = process.env.CHECKLY_CLI_VERSION ?? this.config.version
+    api.defaults.headers['x-checkly-cli-version'] = process.env.CHECKLY_CLI_VERSION ?? await getLatestVersion()
     return super.init()
```
umutuzgur, line 25: This is gonna cause every cli version to report the wrong version. Just pass the version as an env var during the tests please

nahuelon: Good catch! Thanks.

packages/cli/e2e/__tests__/deploy.spec.ts
```diff
@@ -119,7 +119,6 @@ describe('deploy', () => {
       cliVersion: '4.0.9',
     })
     expect(status).toBe(0)
-    expect(stderr).toBe('')
```
umutuzgur, line 122: Please put this back and change the cli version as we discussed to something like `0.0.1-dev` to disable the warning. We will otherwise get this update warning each time we run the cli locally which we don't want

packages/cli/src/commands/baseCommand.ts
nahuelon: Reverted. `oclif` version (`this.config.version`) is used again. @umutuzgur , my concern is that if you are using you local env without configuring correct version in the `CHECKLY_CLI_VERSION` environment variable, the CLI will use an 'old' version and the API request header will be outdated. Currently, we have two different behaviors between `4.0.8` and `>4.0.8` for private-locations assignments. This could result in some annoying results (i.e. PLs assigments will be removed if you are using `>= 4.0.9` source code without `CHECKLY_CLI_VERSION` set). I just updated the `README.md` and `CONTRIBUTING.md`.

umutuzgur (approved):
LGTM

PR #770, opened 2023-06-22
fix: properly close scheduling timer to avoid checkly test hanging

clample:

## Affected Components
* [x] CLI
* [ ] Create CLI
* [ ] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer
With the current canary release from `main`, test runs with `checkly test` always takes 20 seconds. This occurs even when the checks are shown as finished after only several seconds. This can be reproduced with `time checkly test` on the boilerplate example. 

This issue is because we aren't closing the timeout for `MAX_SCHEDULING_DELAY_EXCEEDED`, so the node process hangs until after it runs (in 20 seconds). This issue doesn't affect any public releases since it was merged after v4.0.8.

This PR fixes the issue by closing the timeout after all of the check runs have been scheduled. The code is similar to what we already have for `AbstractCheckRunner.allChecksFinished`.

#### Testing:

Tested `time checkly test` on the boilerplate example. It finishes after ~7 seconds rather than 20.

Also tested that the scheduling delay message is working. To test, I used the following code snippet with `checkly test --location eu-south-1`:
```
const checkTime = 30000
const sleepScript = `await new Promise(resolve => setTimeout(resolve, ${checkTime}))`

for (let i = 0; i < 10; i++) {
  new BrowserCheck(`check-${i}`, {
    name: `Check ${i}`,
    code: {
      content: sleepScript,
    }
  })
}
```

### DIFF
--- a/packages/cli/src/services/abstract-check-runner.ts
+++ b/packages/cli/src/services/abstract-check-runner.ts
@@ -43,6 +43,7 @@ export default abstract class AbstractCheckRunner extends EventEmitter {
   // If there's an error in the backend and no check result is sent, the check run could block indefinitely.
   // To avoid this case, we set a per-check timeout.
   timeouts: Map<CheckRunId, NodeJS.Timeout>
+  schedulingDelayExceededTimeout?: NodeJS.Timeout
   accountId: string
   timeout: number
   verbose: boolean
@@ -198,6 +199,11 @@ export default abstract class AbstractCheckRunner extends EventEmitter {
       return
     }
     Array.from(this.checks.entries()).forEach(([checkRunId]) => this.disableTimeout(checkRunId))
+
+    if (this.schedulingDelayExceededTimeout) {
+      clearTimeout(this.schedulingDelayExceededTimeout)
+      this.schedulingDelayExceededTimeout = undefined
+    }
   }
 
   private startSchedulingDelayTimeout () {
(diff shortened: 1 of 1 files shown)


packages/cli/src/services/abstract-check-runner.ts
```diff
@@ -203,9 +201,19 @@ export default abstract class AbstractCheckRunner extends EventEmitter {
...
-    ))
+    let scheduledCheckCount = 0
+    const numChecks = this.checks.size
+    if (numChecks === 0) {
+      return
+    }
+    const schedulingDelayExceededTimeout = setTimeout(
+      () => this.emit(Events.MAX_SCHEDULING_DELAY_EXCEEDED),
+      DEFAULT_SCHEDULING_DELAY_EXCEEDED_MS,
+    )
```
umutuzgur, line 212: We should also clear this in the exception try catch

clample: Good catch, I'll update it accordingly

packages/cli/src/services/abstract-check-runner.ts
```diff
@@ -37,14 +37,13 @@ export const DEFAULT_CHECK_RUN_TIMEOUT_SECONDS = 300
...
 
-const SCHEDULING_DELAY_EXCEEDED_TIMEOUT_KEY = 'SCHEDULING_DELAY_EXCEEDED'
-
 export default abstract class AbstractCheckRunner extends EventEmitter {
   checks: Map<CheckRunId, { check: any, testResultId?: string }>
   testSessionId?: string
   // If there's an error in the backend and no check result is sent, the check run could block indefinitely.
   // To avoid this case, we set a per-check timeout.
   timeouts: Map<CheckRunId, NodeJS.Timeout>
+  schedulingDelayExceededTimeout?: NodeJS.Timeout
```
clample, line 46: We could put it in `timeouts`, but I prefer having a separate property

umutuzgur (approved):

PR #774, opened 2023-06-28
chore: refactor create-cli package [gh-761]

### LINKED ISSUE #761: story: add tests for `npm create checkly` prompts
### Story description

The package `create-cli` has one command with seven prompts a user could fill. There are lot of 
 combinations that should be tested every time a change is applied. Adding unit and integration tests will help us during the development.

### Shortcut link

_No response_

### Additional resources

At the moment, [prompts](https://www.npmjs.com/package/prompts) is used for user interaction. The package has the `inject()` [function](https://github.com/terkelg/prompts#injectvalues) that let us inject user answers.

nahuelon:

## Affected Components
* [ ] CLI
* [x] Create CLI
* [x] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer
This PR is the first iteration of a `create-cli` package refactor. It structures the source code separating `prompts.ts` and moving installation steps to `installations.ts`. The goal is to keep the source code in small functions that can be well tested using unit tests.

New unit tests and fixtures are included testing packages and directory related functions.

This PR DOES NOT modify any message or functionality from the two possible scenarios: `1` install within a initiated project or `2` creating a project using a template.

All `process.cwd()` are removed and absolute directory path is used in the complete flow: the path is asked at the beginning. It's important to note that users could specify absolute or relative paths (i.e. `../../../foo/bar` or just `.` to indicate the `CWD`)

The creation/installation flow is modified as:

image

> TODO: a second iteration will remove ECMAScript module support for the `create-cli` project because `jest` doesn't work well and it doesn't support 'chaining' dependencies with `.js` extensions and I wasn't able to add tests for the `installation.ts`. Before removing ESM, we have to find alternatives for `which-pm-run` and `fullname` package dependencies. The `execa` package could be replaced with the `node:child_process.spawnSync` and execute commands like we are already doing in `cli` package.

> TODO: avoid repeating `dirPath` argument on multiple functions, using some smarter solution.

> Resolves #761


tnolet (changes requested):
Code looks fine, but we want the old flow, but then without the bugs.

1. User calls `npm create checkly` in a directory that has either a `.git` file or `package.json` file.

2. We recognize that and DO NOT ask the user for the directory they want to install in. It makes no sense as a user now has to remember or know they have to provide a `.` as the current dir.

nahuelon:
@tnolet , done. Now, before asking the user to enter the project directory, we check if there is a `package.json` in the `CWD` (as we do in the actual flow).
In your comment you named the `.git` directory too, but we are NOT checking that in flow.
In both flows, the directory has to be totally empty if there is no `package.json` file.
Do you want me to add the `.git` to check it's a initiated project? In that case, we have to take care about creating the full `package.json` rather than just add the `devDependencies` to the file.

tnolet:
> @tnolet , done. Now, before asking the user to enter the project directory, we check if there is a `package.json` in the `CWD` (as we do in the actual flow). In your comment you named the `.git` directory too, but we are NOT checking that in flow. In both flows, the directory has to be totally empty if there is no `package.json` file. Do you want me to add the `.git` to check it's a initiated project? In that case, we have to take care about creating the full `package.json` rather than just add the `devDependencies` to the file.

I think the package.json check is enough for now actually.

tnolet (approved):
LGTM

tnolet (approved):
LGTM

PR #792, opened 2023-07-05
chore: create-cli cleanup

danielpaulus:

## Affected Components
* [ ] CLI
* [x] Create CLI
* [x] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer
Changes are described in https://github.com/checkly/checkly-cli/pull/803

- After merging https://github.com/checkly/checkly-cli/pull/803 the `prepare` script renaming was reverted (`TODO`: investigate why tests fails using `build` script).

This PR adds a retry/fallback when the template download fails using the version tag/branch. If template couldn't be found, `main` branch is used to download the latest version. This is helpful when we change git tag format or versioning naming.

**IMPORTANT**: after merging this PR and releasing the new `create-cli`, the templates will be downloaded using the tag/branch without the `v` prefix. So, release's tags MUST be like `4.0.15`.

> `TODO` in a separated PR: try to rename `prepare` script with `build`
> `TODO` in https://github.com/checkly/checkly-cli/pull/814: add `login` and `switch` CLI commands E2E tests (using `prompts.inject()`)


nahuelon:
@umutuzgur , I opened https://github.com/checkly/checkly-cli/pull/803 with some changes to be merged here, into this branch. I didn't commit into this branch because I would like to see how this branch is working in your local env. I tried a lot without success. The only way I found to make it work was removing all ESM references/packages. That's why I pushed all changes in a separated branch.

umutuzgur (dismissed):
I assume you tested it thoroughly

umutuzgur:
> @umutuzgur , I added a retrial mechanism when downloading the template from a specific version tag/branch. If the download fails (i.e. tag doesn't exists) the `main` branch is used to download the template.

@nahuelon is this for testing? If not, we don't need this. We shouldn't have a fallback for cases that are basically our mistakes

nahuelon:
> > @umutuzgur , I added a retrial mechanism when downloading the template from a specific version tag/branch. If the download fails (i.e. tag doesn't exists) the `main` branch is used to download the template.
> 
> @nahuelon is this for testing? If not, we don't need this. We shouldn't have a fallback for cases that are basically our mistakes

@umutuzgur , I removed the 'fallback to `main`' retrial. Now, if we detects the `create-cli` version is `0.0.1-dev` (which means it's a local env or a E2E) we replace it with the latest from NPM. This is useful mainly for the local development if you forgot to configure the `CHECKLY_CLI_VERSION` env var.
Also, I added a new E2E test forcing to use an older version (`4.0.13` in this case).

umutuzgur (approved):

PR #802, opened 2023-07-25
feat(heartbeat): add heartbeat check construct [sc-16788]

shiini2:

## Affected Components
* [x] CLI
* [ ] Create CLI
* [x] Test
* [ ] Docs
* [x] Examples
* [ ] Other

## Notes for the Reviewer

Adds the new `Heartbeat` check type.

This PR requires [this BE branch](https://github.com/checkly/checkly-backend/pull/4532).


packages/cli/src/constructs/heartbeat-check.ts
```diff
@@ -0,0 +1,68 @@
...
+
+export class HeartbeatCheck extends Check {
+  heartbeat: Heartbeat
+
+  constructor (logicalId: string, props: HeartbeatCheckProps) {
+    super(logicalId, props)
+
+    _customPeriodGraceValidation(props.period, props.periodUnit, props.grace, props.graceUnit)
+
+    this.heartbeat = {
```
maxigimenez, line 51: Do we need this extra `heartbeat` object?

maxigimenez (changes requested):

packages/cli/src/constructs/heartbeat-check.ts
```diff
@@ -0,0 +1,68 @@
+import { Check, CheckProps } from './check'
+import { Session } from './project'
+import { DateTime } from 'luxon'
+
+type unitType = 'seconds' | 'minutes' | 'hours' | 'days'
```
clample, line 5: Maybe `type TimeUnits` for the name is more specific

clample:
Looks good!

Could you please add the heartbeat check to e2e tests for `checkly test` and `checkly deploy` too? It should work to add heartbeat checks to some of the fixtures we have. The e2e tests are [here](https://github.com/checkly/checkly-cli/tree/main/packages/cli/e2e/__tests__).

How will `checkly test` work when there are heartbeat checks? Are they already skipped , or do we need to add some code for filtering for them?

Antoine-C (commented):
We should initialize the pingToken to an uuid.v4() if it's not set, it'll be required in the public API because of TF diffs

packages/cli/src/constructs/heartbeat-check.ts
shiini2: Yes we have validation where we need the normal check properties + a heartbeat object when a check is of type `HEARTBEAT`.

packages/cli/src/constructs/heartbeat-check.ts
shiini2: Good call!

packages/cli/src/constructs/heartbeat-check.ts
shiini2: Yes we have validation where we need the normal check properties + a heartbeat object when a check is of type HEARTBEAT.

packages/cli/src/constructs/heartbeat-check.ts
```diff
@@ -0,0 +1,65 @@
...
+  if (
+    addedTimePeriod.diff(now, 'days').days > MAX_PERIOD_GRACE_DAYS ||
+    addedTimePeriod.diff(now, 'seconds').seconds < MIN_PERIOD_SECONDS
+  ) {
+    throw new Error('Period must be between 30 seconds and 365 days.')
+  }
+
+  if (addedGracePeriod.diff(now, 'days').days > MAX_PERIOD_GRACE_DAYS) {
+    throw new Error('Grace must be less than 365 days.')
+  }
```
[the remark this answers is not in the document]
shiini2: The condition above this one is checking for `addedTimePeriod.diff()` while this one checks for `addedGracePeriod.diff()`, as far as I tested, both seem to work as expected

shiini2:
This PR should be tested with [this BE branch](https://github.com/checkly/checkly-backend/pull/4532) for the tests to pass.

packages/cli/src/constructs/heartbeat-check.ts
```diff
@@ -0,0 +1,66 @@
...
+}
+
+export class HeartbeatCheck extends Check {
+  heartbeat: Heartbeat
+
+  constructor (logicalId: string, props: HeartbeatCheckProps) {
+    super(logicalId, props)
+
+    _customPeriodGraceValidation(props.period, props.periodUnit, props.grace, props.graceUnit)
```
nahuelon, line 48: nit: you could replace the parameters with to an `object` and call just `_customPeriodGraceValidation(props)`

nahuelon (approved):
LGTM. (just one nit comment)

Antoine-C (approved):
LGTM 🚀

shiini2:
> Looks good!
> 
> Could you please add the heartbeat check to e2e tests for `checkly test` and `checkly deploy` too? It should work to add heartbeat checks to some of the fixtures we have. The e2e tests are [here](https://github.com/checkly/checkly-cli/tree/main/packages/cli/e2e/__tests__).
> 
> How will `checkly test` work when there are heartbeat checks? Are they already skipped , or do we need to add some code for filtering for them?

@clample Heartbeat checks are excluded from `trigger` and `test` commands, so the rest of the API and Browser checks should run as expected.

maxigimenez (approved):
LGTM 👍 🥳

nahuelon (approved):
LGTM. 🚀

PR #803, opened 2023-07-26
chore: remove all ESM packages/references [gh-0]

nahuelon:

## Affected Components
* [ ] CLI
* [x] Create CLI
* [x] Test
* [x] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer
The code in https://github.com/checkly/checkly-cli/pull/792 doesn't work in my local environment. That's why I open this PR to merge changes to https://github.com/checkly/checkly-cli/pull/792 instead of `main`. Changes here make the `create-cli` package to compile and run in Linux and Windows. I didn't test it in macOS.

This PR removes all references and usage of packages that require ESM support or `esModuleInterop: true`. Changes are:
- `execa` is downgraded to `5.1.0` version which supports `commonjs`
- `which-pm-runs` is replaced with `getPackageManager()` function which use same logic from `which-pm-runs`
- `fullname` is replaced with `getFullName()` function which uses same logic `fullname` NPM packages

This PR also includes new E2E tests for `create-cli` package using `prompt.inject()` values passed through `CHECKLY_E2E_PROMPTS_INJECTIONS` environment variable. Also, the `oclif` version is now working as expected and can be override using `CHECKLY_CLI_VERSION` (as we could so in `cli` package)

~> `TODO`: rename `package.json` `prepare` script with `build` and adjust CI/CD workflow.~
~> `TODO`: add e2e tests~
~> `TODO`: add E2E for template override using `npx create checkly --template ...`~
~> `TODO`: test `create-cli` E2E on Windows~
> `TODO`: test and review CI/CD workflows for `create-cli` package (ensure `test:e2e` is triggered when PR are created)
> `TODO`: install `execa@5.1.1` which it's not working from my local env: when I try it, the `execa@1.0.0` is added to the `package.json` 

> *IMPORTANT*: templates are downloaded using git tag without the `v` prefix, i.e.: `4.0.13`


packages/create-cli/package.json
```diff
@@ -40,19 +40,18 @@
...
   "homepage": "https://github.com/checkly/checkly-cli#readme",
   "dependencies": {
+    "@esm2cjs/execa": "6.1.1-cjs.1",
     "@oclif/core": "2.8.11",
     "@oclif/plugin-help": "5.1.20",
     "@oclif/plugin-plugins": "2.3.0",
     "axios": "1.3.1",
     "chalk": "4.1.2",
     "debug": "4.3.4",
-    "execa": "5.1.1",
```
clample, line 49: I think `execa 5.1.1` should still be Common JS. According to the release log, they bumped the version in v6.0.0. Do we still need to use `@esm2cjs2/execa`?

https://github.com/sindresorhus/execa/releases

clample:
@umutuzgur or @nahuelon - do you have any idea why https://github.com/checkly/checkly-cli/pull/792 would work on Umut's machine but not on Nahuel's? I understand you're on different OS's, but that would surprise me if it causes any differences with respect to ES Modules / CommonJS.

I see that umutuzgur/gh-0/create-cli-cleanup is a draft PR. @umutuzgur since it's your PR, I guess it's at your discretion whether or not to add this to the PR.

packages/create-cli/package.json
nahuelon: @clample , You are right. But `execa@5.1.1` compiles but when it runs returns that it need ESM.

umutuzgur:
> This PR removes all references and usage of packages that require ESM support or esModuleInterop: true.

Why are we doing this btw? What does it actually fix?

packages/create-cli/src/actions/dependencies.ts
```diff
@@ -1,7 +1,7 @@
 import * as fs from 'fs'
 import * as path from 'path'
-import detectPackageManager from 'which-pm-runs'
-import execa from 'execa'
+import * as whichpm from 'which-pm'
```
umutuzgur, line 3: This seems to be doing something different than `which-pm-runs`. This detects the installation tool and `which-pm-runs` detects which tool was used to run the code

packages/create-cli/package.json
```diff
@@ -40,19 +40,19 @@
...
-    "fullname": "4.0.1",
     "giget": "1.1.2",
     "ora": "5.4.1",
+    "passwd-user": "3.0.0",
```
umutuzgur, line 52: The package readme says it doesn't work on windows. Have you checked that out?

nahuelon: @umutuzgur , yes yes, it worked. That `passwd-user` package is one of the multiple alternatives to get the username. I tried the `create-cli` on Windows and Linux and it can get the fullname:
image

packages/create-cli/package.json
umutuzgur: The older versions of execa should work with commonjs. Let's not use a fork nobody knows

nahuelon: Agree with the fork usage. I'll try `execa@5.1.1` again. I wasn't able to get it work: it compiles but when it runs it throw a 'ESM required' error. I'll share the issue here.

packages/create-cli/src/actions/dependencies.ts
nahuelon: Fixed. I simply created a function `getPackageManager()` that uses the same logic used in `which-pm-runs`.

packages/create-cli/package.json
```diff
@@ -46,13 +45,12 @@
     "axios": "1.3.1",
     "chalk": "4.1.2",
     "debug": "4.3.4",
-    "execa": "5.1.1",
-    "fullname": "4.0.1",
+    "execa": "1.0.0",
```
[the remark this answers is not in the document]
nahuelon: I don't know what's happening but if I try to install `npm i execa@5.1.1`, the `1.0.0` is added to the `package.json`. I already cleaned the NPM cache, but it's still happening. Just to unblock, I used the `5.1.0` which works OK.

umutuzgur (approved):
LGTM

PR #818, opened 2023-08-08
feat(heartbeat): add Heartbeat checks to example projects [sc-17054]

shiini2:

## Affected Components
* [ ] CLI
* [ ] Create CLI
* [ ] Test
* [ ] Docs
* [x] Examples
* [ ] Other

## Notes for the Reviewer

We only had a Heartbeat check on one example project. 
Add the check to the rest of the projects.


ndom91 (approved):

stefanjudis:
If we move forward (don't have a solution for the free plan either), could we add the `mac` tag to the two included heartbeat checks? Than it's consistent with the other example checks. :)

shiini2:
@stefanjudis all the checks created from the 4 example projects should automatically include the `mac` tag for all checks as it's set within the `checkly.config.ts` [here](https://github.com/checkly/checkly-cli/blob/main/examples/boilerplate-project/checkly.config.ts#L22). Did it not add them when you created checks? 🤔

shiini2:
> Heartbeat checks aren't enabled for free plan users. One downside will be that free plan users start seeing errors when trying out the CLI. I don't have a good alternative, but just thought I'd bring up the tradeoff.

Good point! We could silently not create heartbeat checks if the account doesn't have the feature 🤔 
Although the current error message users get acts as a CTA, but you're right that it's not nice for a first time user to get errors. @maxigimenez what do you think?

stefanjudis:
@shiini2 Re: tags I looked at the wrong thing. Ignore me. All good. 🤦 

Re: the heartbeat checks, I thought about it a little bit and would probably keep the file to show folks what's possible, comment out `new HeartbeatCheck` (or whatever it is), and place comments above.

Very sloppy phrased: 👇 

> This is a feature on a paid plan, learn more about it here. Uncomment it to Heartbeats use it.

stefanjudis:
Looks very good to me. 💯

PR #819, opened 2023-08-08
chore: add sslCheckDomain property [gh-0]

nahuelon:

## Affected Components
* [x] CLI
* [ ] Create CLI
* [x] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer
This PR adds `sslCheckDomain` into `BrowserCheck` construct.

> Depends on https://github.com/checkly/checkly-backend/pull/4591

### DIFF
--- a/packages/cli/src/constructs/browser-check.ts
+++ b/packages/cli/src/constructs/browser-check.ts
@@ -133,7 +133,7 @@ export class BrowserCheck extends Check {
       script: this.script,
       scriptPath: this.scriptPath,
       dependencies: this.dependencies,
-      sslCheckDomain: this.sslCheckDomain ?? null,
+      sslCheckDomain: this.sslCheckDomain || null, // empty string is converted to null
     }
   }
 }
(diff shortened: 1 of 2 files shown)


packages/cli/src/constructs/browser-check.ts
```diff
@@ -17,6 +17,10 @@ export interface BrowserCheckProps extends CheckProps {
    * with the Puppeteer or Playwright frameworks.
    */
   code: Content|Entrypoint
+  /**
+   * A valid fully qualified domain name (FQDN) to check its SSL certificate.
```
clample, line 21:
```suggestion
   * A valid fully qualified domain name (FQDN) to check for SSL certificate expiration. For example, 'app.checklyhq.com'.
```

clample: I think users might be a bit confused whether or not to add `https://`, etc. I think that adding a little hint could make sense.

clample (dismissed):

packages/cli/src/constructs/browser-check.ts
nahuelon: Totally agree. Thanks!

clample (approved):

PR #820, opened 2023-08-08
feat(heartbeat): print `pingUrl` in the console after deploy for heartbeat checks [sc-17085]

shiini2:

## Affected Components
* [x] CLI
* [ ] Create CLI
* [ ] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer

When creating or updating heartbeat checks from `npx checkly deploy`, we want to print in the console the ping URL of each heartbeat check.
iTerm2:
image

Terminal with default theme:
image

Note: needs [deploy response adjustment PR](https://github.com/checkly/checkly-backend/pull/4593) on the BE when testing.

### DIFF
--- a/packages/cli/src/commands/deploy.ts
+++ b/packages/cli/src/commands/deploy.ts
@@ -116,6 +116,14 @@ export default class Deploy extends AuthCommand {
       if (!preview) {
         await ux.wait(500)
         this.log(`Successfully deployed project "${project.name}" to account "${account.name}".`)
+
+        // Print the ping URL for heartbeat checks.
+        data.diff.forEach(async (check) => {
+          if (check.checkId) {
+            const { data: { pingUrl, name } } = await api.heartbeatCheck.get(check.checkId)
+            this.log(`The ping URL for the heartbeat check ${chalk.greenBright(name)} is available at ${chalk.italic.underline.cyanBright(pingUrl)}.`)
+          }
+        })
       }
     } catch (err: any) {
       if (err?.response?.status === 400) {
--- a/packages/cli/src/commands/deploy.ts
+++ b/packages/cli/src/commands/deploy.ts
@@ -118,10 +118,14 @@ export default class Deploy extends AuthCommand {
         this.log(`Successfully deployed project "${project.name}" to account "${account.name}".`)
 
         // Print the ping URL for heartbeat checks.
-        data.diff.forEach(async (check) => {
-          if (check.checkId) {
-            const { data: { pingUrl, name } } = await api.heartbeatCheck.get(check.checkId)
-            this.log(`The ping URL for the heartbeat check ${chalk.greenBright(name)} is available at ${chalk.italic.underline.cyanBright(pingUrl)}.`)
+        const heartbeatLogicalIds = project.getHeartbeatLogicalIds()
+        const heartbeatCheckIds = data.diff.filter((check) => heartbeatLogicalIds.includes(check.logicalId))
+          .map(check => check?.physicalId)
+
+        heartbeatCheckIds.forEach(async (id) => {
+          if (id) {
+            const { data: { pingUrl, name } } = await api.heartbeatCheck.get(id)
+            this.log(`Ping URL of heartbeat check ${chalk.green(name)} is ${chalk.italic.underline.blue(pingUrl)}.`)
           }
         })
       }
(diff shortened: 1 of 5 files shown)


packages/cli/src/commands/deploy.ts
```diff
@@ -116,6 +116,14 @@ export default class Deploy extends AuthCommand {
       if (!preview) {
         await ux.wait(500)
         this.log(`Successfully deployed project "${project.name}" to account "${account.name}".`)
+
+        // Print the ping URL for heartbeat checks.
+        data.diff.forEach(async (check) => {
```
maxigimenez, line 121: Will `data.diff` only contains heartbeat checks?

maxigimenez (changes requested):

packages/cli/src/commands/deploy.ts
shiini2: On the current implementation for the BE PR [here](https://github.com/checkly/checkly-backend/pull/4593), yes. But will most likely change the logic for this according to Chris's suggestion.

ndom91 (dismissed):

packages/cli/src/commands/deploy.ts
```diff
@@ -116,6 +116,18 @@ export default class Deploy extends AuthCommand {
...
         await ux.wait(500)
         this.log(`Successfully deployed project "${project.name}" to account "${account.name}".`)
+
+        // Print the ping URL for heartbeat checks.
+        const heartbeatLogicalIds = project.getHeartbeatLogicalIds()
+        const heartbeatCheckIds = data.diff.filter((check) => heartbeatLogicalIds.includes(check.logicalId))
+          .map(check => check?.physicalId)
+
+        heartbeatCheckIds.forEach(async (id) => {
+          if (id && typeof id === 'string') {
```
umutuzgur, line 126: Which case is this btw?

packages/cli/src/constructs/project.ts
```diff
@@ -111,6 +111,13 @@ export class Project extends Construct {
           .filter((construct: Construct) => construct instanceof Check && construct.testOnly))
   }
 
+  getHeartbeatLogicalIds (): string[] {
+    return Object
+      .values(this.data.check)
+      .filter((construct: Construct) => construct instanceof Check && construct.constructor.name === 'HeartbeatCheck')
```
umutuzgur, line 117: We should have a HeartbeatCheck class as well, right? We can do a instance of directly on that

packages/cli/src/rest/projects.ts
```diff
@@ -11,7 +11,8 @@ type ProjectResponse = Project & { id: string, created_at: string }
 
 export interface Change {
   logicalId: string,
-  physicalId?: string,
+  physicalId?: string|number,
+  checkId?: string,
```
umutuzgur, line 15: This is a mistake I guess?

packages/cli/src/commands/deploy.ts
shiini2: This is when the `physicalId` is the `checkId` (in this case of a heartbeat check), this is always going to be a string since it's a uuid

packages/cli/src/rest/projects.ts
shiini2: The `physicalId` is [now](https://github.com/checkly/checkly-backend/pull/4593/files#diff-d3940f49c21abfbcd392c2edc807229e39ff31e9227f2ca682bea0c26572599fR99) filled either with the `physicalId` if it's available, otherwise it's the resource ID (either checkId, dashboardId etc..), those can either be a string or a number, right?

shiini2: Or do you mean the `checkId`? Because that one is indeed a mistake 😆

packages/cli/src/commands/deploy.ts
umutuzgur: Do we need this if then?

shiini2: Typescript was otherwise complaining on the usage of the id right below this line

umutuzgur: Ah, gotcha. You can cast it to string using `as` btw. It will be always string in heartbeats so we can force it here

ndom91 (approved):
Nice, is much more readable after those updates too :+1:

PR #825, opened 2023-08-14
chore: handle 408 error and show message [gh-824]

### LINKED ISSUE #824: story: better error message when a user recieves 408
### Story description

Given that the cli relies on a user's network, they might hit timeouts if they for example have a slower internet connection. Our BE returns 408 in this case. We should show a useful error message to the user to indicate that it is related to the internet connection

### Shortcut link

_No response_

### Additional resources

_No response_

nahuelon:

## Affected Components
* [x] CLI
* [ ] Create CLI
* [x] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer
This PRs makes use of a response `axios` interceptor to catch `408` errors and print a user-friendly message.

image

I tried to add unit-tests for `axios` requests and interceptors [here](https://github.com/checkly/checkly-cli/pull/825/commits/e1c70a68459104dacb09cf968394c4b05c8e16ec), but I had to `__mocks__` axios resulting in issues with the e2e tests. That's why I moved the interceptors handler to separated functions to see if I could `spyOn` or `mock` them (without success).

`TODO`: try to test interceptors. FYI, mocking `axios` worked to test request but it require some E2E tests adjustments.

> Resolves #824


clample:
The interceptor changes look good to me 👍 

I'm not sure about the runtimes e2e test. We'll need to update the snapshot whenever we update runtimes. Could you move it out of this PR since it doesn't seem needed to handle the 408 issue?

packages/cli/src/rest/api.ts
```diff
@@ -46,26 +46,42 @@ export async function validateAuthentication (): Promise<void> {
...
+  if (accountId && config.headers) {
+    config.headers['x-checkly-account'] = accountId
+  }
+
+  return config
+}
+
+export function responseErrorInterceptor (error: any) {
+  if (error.response?.status === 408) {
+    throw new Error('Encountered an error connecting to Checkly. Your Internet connection is too slow.')
```
umutuzgur, line 64: We should improve this message. It is not an error while connecting to Checkly but during transferring the data. We should also paraphrase the reason. "This can be triggered by a slow internet connection or a network with high packet loss"

nahuelon: @umutuzgur , I changed the error reaso. I also kept "Encountered an error connecting to Checkly.". Is it OK for you, or you want me to remove it and just show the message you sent above.

umutuzgur (approved):
LGTM

PR #827, opened 2023-08-14
fix: use ipv4 for local setup

clample:

## Affected Components
* [x] CLI
* [ ] Create CLI
* [ ] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer
On Mac, I receive the following error when running the CLI with `CHECKLY_ENV=local`:
```
 Error: Encountered an error connecting to Checkly. Please check that the internet connection is working. connect ECONNREFUSED ::1:3000
```

The CLI tries to connect to the IPv6 loopback address. The local setup isn't bound to the IPv6 interface, though, so this fails. To fix the issue, we can make sure that the CLI always uses the IPv4 address.

### DIFF
--- a/packages/cli/src/services/config.ts
+++ b/packages/cli/src/services/config.ts
@@ -79,7 +79,7 @@ class ChecklyConfig {
 
   getApiUrl (): string {
     const environments = {
-      local: 'http://localhost:3000',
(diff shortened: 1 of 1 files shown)


nahuelon (approved):
LGTM.

Issue #800, opened 2023-07-20
bug: referencing a file above the cwd / checkly config location fails


jan-osch:
### Node.js version

16.19.1

### NPM version

8.19.3

### @checkly/cli version

4.0.12

### Steps to reproduce

Repro repo:
https://github.com/checkly/cli-playground-jan/tree/issue-file-outside-temp-dir

### What is expected?

The check should pass

### What is actually happening?

Check is failing because it cannot create a dependency file outside of the temp directory

nahuelon:
@umutuzgur , as far I could investigate we should adjust dependencies path and make all to start from the path of the spec file. If now, the lambda-runner won't be able to handle `../../` path from the `/tmp/...` where the check runs.
If we adjust the dependency path we have to modify the `import` path (in the spec file) adjusting it  using a relative path starting from the spec file folder. I can create a draft PR to illustrate what I mean, but I would like to know if we agree in support it.

sorccu:
I know this is very old but I don't think this is relevant anymore. We've had workspace support for a while and as long as the file you are referring to is within the workspace (even if above the checkly config file), it will work just fine. However if it's outside the workspace root, then that still won't work.

PR #829, opened 2023-08-17
chore: add bad reference error [gh-800]

nahuelon:

## Affected Components
* [x] CLI
* [ ] Create CLI
* [x] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer
This PRs catches dependency file paths that are above the project CWD and show a user-friendly error.
The check is done for API and Browser checks, because both checks could have not permitted paths for their dependencies.

image

> Resolves #[issue-number]


umutuzgur:
Please fix the test so I can review it

packages/cli/src/constructs/api-check.ts
```diff
@@ -356,8 +356,12 @@ export class ApiCheck extends Check {
 
     const deps: ScriptDependency[] = []
     for (const { filePath, content } of parsed.dependencies) {
+      const relativePath = pathToPosix(path.relative(Session.basePath!, filePath))
+      if (relativePath.startsWith('..')) {
+        throw new Error(`You cannot reference a dependency file above the Checkly config location folder. Please, modify "${relativePath}" path.`)
```
clample, line 361: For the error message, what do you think about:

> Failed to parse dependencies for check ${checkLogicalId} "${checkName}". All check dependencies must be contained in the project directory, which is the directory containing the Checkly config file. Checks cannot reference files which are outside of this directory.
>
> Project directory: ${Session.basePath}
> Dependency: ${filePath}

I tried to add the check logicalId/name to give a hint where the issue is coming from. I also think logging the project directory could let the user know where they're allowed to put files.

I think that it could make sense to use the absolute path rather than `relativePath`. I'm not sure if we need to call `path.resolve` on `filePath`.

packages/cli/src/constructs/browser-check.ts
```diff
@@ -111,8 +111,12 @@ export class BrowserCheck extends Check {
 
     const deps: CheckDependency[] = []
     for (const { filePath, content } of parsed.dependencies) {
+      const relativePath = pathToPosix(path.relative(Session.basePath!, filePath))
```
clample, line 114: What happens if the main browser check script or setup/teardown scripts are outside of the `Session.basePath`? I'm not sure if checkly-runners supports that, or if we should add an error in that case as well.

tnolet:
@umutuzgur @clample can we update this issue?

clample:
I'm going to close this one for now

PR #830, opened 2023-08-25
feat: add support for retry strategies

clample:

## Affected Components
* [x] CLI
* [ ] Create CLI
* [ ] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer
This PR adds support for retry strategies to checks and check groups. The feature is still in beta, so the examples aren't updated. Users can now configure retry strategies as follows:

```
new ApiCheck('homepage-api-check-3', {
    name: 'Test',
    retryStrategy: RetryStrategyBuilder.linearStrategy({ baseBackoffSeconds: 10, maxAttempts: 10, maxDurationSeconds: 600, sameRegion: false }),
    request: {
       url: 'https://danube-web.shop/api/books',
       method: 'GET',
   }
})
```

It's possible to configure retry strategies on both checks and groups.

Defaults for retry strategies are implemented in the CLI side. This should give us a bit more flexibility in changing defaults between CLI releases.


packages/cli/src/constructs/retry-strategy.ts
```diff
@@ -0,0 +1,68 @@
...
+   * The delay between retries is calculated using `baseBackoffSeconds ^ attempt`.
+   * For example, retries will be run with backoff 10s, 100s, 1000s, and so on.
+   */
+  static exponentialStrategy (options: RetryStrategyOptions): RetryStrategy {
+    return RetryStrategyBuilder.retryStrategy('EXPONENTIAL', options)
+  }
+
+  private static retryStrategy (type: RetryStrategyType, options: RetryStrategyOptions): RetryStrategy {
+    return {
+      ...options,
```
umutuzgur, line 60: This is not actually needed given the possible options are repeated below, right?

clample: Good catch, I'll remove it

PR #854, opened 2023-09-20
chore: update module to node16 [gh-0]

umutuzgur:

## Affected Components
* [x] CLI
* [ ] Create CLI
* [ ] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer
The new ts parser requires node 16 minimum so we need to change the module


packages/cli/tsconfig.json
```diff
@@ -3,7 +3,8 @@
     "incremental": true,
     "declaration": true,
     "importHelpers": false,
-    "module": "commonjs",
+    "module": "node16",
+    "esModuleInterop": true,
```
clample, line 7: Do these settings apply to `ts-node` when we're picking up the `checkly.config.ts` and `*.check.ts` files? Is there any user facing impact?

clample (approved):
Look

packages/cli/tsconfig.json
umutuzgur: They shouldn't. We don't put the tsconfig into our build folder

clample: I think you're right - good point

umutuzgur:
I will try to migrate create-cli as well later. This was a big headache

PR #864, opened 2023-09-29
Antoinecoutellier/sc 17815/add new multistepcheck constructor to the

Antoine-C:
I hereby confirm that I followed the code guidelines found at [engineering guidelines](https://www.notion.so/checkly/Engineering-Guidelines-a3a165a203a04dc1a112f0e26b2f2d3f)

## Affected Components
* [ ] CLI
* [ ] Create CLI
* [ ] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer

> Resolves #[issue-number]

PR #868, opened 2023-10-12
feat: add multistep construct and support for running test session [sc-17814]

ndom91:

## Affected Components
* [x] CLI
* [ ] Create CLI
* [ ] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer

- Completes [sc-17814](https://app.shortcut.com/checkly/story/17814/update-public-api-to-allow-to-run-test-multi-steps-through-the-cli) and [sc-17815](https://app.shortcut.com/checkly/story/17815/add-new-multistepcheck-constructor-to-the-cli)
	- Regarding `17814`, we already supported `MULTI_STEP` check runs via test-session due to `checkly-backend/api/src/modules/public-api/test-sessions/PublicTestSessionsController.js:93` and the following code paths submitting the check run to the correct queue based off of its `checkType`.
- Adds multistep check construct, based off of browser construct
- Supercedes: https://github.com/checkly/checkly-cli/pull/864
	- Reused test and example files from there :pray: 

Manually tested with `npm create checkly` simple example and the following files:
```typescript
// multistep.check.ts
import { MultiStepCheck } from 'checkly/constructs'

new MultiStepCheck('multi-step-check-1', {
  name: 'SpaceX Template',
  frequency: 15,
  code: { entrypoint: 'homepage.spec.ts' },
})
```
```typescript
// homepage.spec.ts
import { test, expect } from '@playwright/test'

const baseUrl = 'https://api.spacexdata.com/v3'

test('space-x dragon capsules', async ({ request }) => {
  /**
   * Get all SpaceX Dragon Capsules
   */
  const response = await test.step('get all capsules', async () => {
    return request.get(`${baseUrl}/dragons`)
  })

  expect(response).toBeOK()

  const data = await response.json()
  expect(data.length).toBeGreaterThan(0)

  const [first] = data

  /**
   * Get a single Dragon Capsule
   */
  const getSingleResponse = await test.step('get single dragon capsule', async () => {
    return request.get(`${baseUrl}/dragons/${first.id}`)
  })

  expect(getSingleResponse).toBeOK()

  const dragonCapsule = await getSingleResponse.json()
  expect(dragonCapsule.name).toEqual(first.name)
})
```

Adding playwrights `context` to `homepage.spec.ts:5` test callback arguments correctly returns `cannot use 'context' in MULTI_STEP checks`, confirming its been submitted and handled by `2023-09-multistep` runner.


packages/cli/src/constructs/multi-step-check.ts
```diff
@@ -0,0 +1,141 @@
...
+}
+
+/**
+ * Creates a multi-step Check
+ *
+ * @remarks
+ *
+ * This class make use of the multi-step checks endpoints.
+ */
+export class MultiStepCheck extends Check {
```
Antoine-C, line 31: In the PR description, you mention `Adds multistep check construct, based off of browser construct` but it extends Check and not BrowserCheck. Is it wanted?

ndom91: Yeah it was just based off of the browsercheck construct, meaning they both extend `Check`

tnolet: @ndom91 @Antoine-C we should not add the `slDomain` property

ndom91: Haha we had this discussion elsewhere recently. They're still included in the multistep UI as well. So should we remove them completely from multistep?

Antoine-C: Do we need also to update the API?

Antoine-C (approved):

packages/cli/src/constructs/multi-step-check.ts
ndom91: Yeah good question, that might be the reason for the failing tests. Doing some more digging here today..

PR #876, opened 2023-10-31
feat: add alpha support for PWT snapshot testing

clample:

## Affected Components
* [x] CLI
* [ ] Create CLI
* [ ] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer
This PR adds initial support for PWT snapshot testing in the CLI . `npx checkly test --update-snapshots` updates any snapshots using the actual result of this test run. `npx checkly deploy` can then be used to actually push the snapshots: it updates the snapshots used by the check on Checkly.

TODO:
* [ ] Update examples
* [ ] Add integration test

One limitation in the current implementation is that we push the files to storage with every `npx checkly deploy`. We could probably improve this by adding some change detection using hashing. Such an implementation would be more complicated, though, so maybe we save it for later.

Integration tests are failing since the backend hasn't been deployed with the new schema yet.

### DIFF
--- a/packages/cli/src/commands/test.ts
+++ b/packages/cli/src/commands/test.ts
@@ -95,6 +95,7 @@ export default class Test extends AuthCommand {
       description: 'A name to use when storing results in Checkly with --record.',
     }),
     'update-snapshots': Flags.boolean({
+      char: 'u',
       description: 'Update any snapshots using the actual result of this test run.',
       default: false,
       // Mark --update-snapshots as hidden until we're ready for GA
--- a/packages/cli/src/commands/test.ts
+++ b/packages/cli/src/commands/test.ts
@@ -98,6 +98,8 @@ export default class Test extends AuthCommand {
     'update-snapshots': Flags.boolean({
       description: 'Update any snapshots using the actual result of this test run.',
       default: false,
+      // Mark --update-snapshots as hidden until we're ready for GA
+      hidden: true,
     }),
   }
 
--- a/packages/cli/src/commands/test.ts
+++ b/packages/cli/src/commands/test.ts
@@ -19,7 +19,6 @@ import { AuthCommand } from './authCommand'
 import { BrowserCheck, Check, HeartbeatCheck, Session } from '../constructs'
 import type { Region } from '..'
 import { splitConfigFilePath, getGitInformation, getCiInformation, getEnvs } from '../services/util'
-import { pullSnapshots } from '../services/snapshot-service'
 import { createReporters, ReporterType } from '../reporters/reporter'
 import commonMessages from '../messages/common-messages'
 import { TestResultsShortLinks } from '../rest/test-sessions'
@@ -227,6 +226,7 @@ export default class Test extends AuthCommand {
       repoInfo,
       ciInfo.environment,
       updateSnapshots,
+      configDirectory,
     )
 
     runner.on(Events.RUN_STARTED,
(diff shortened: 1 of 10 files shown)


packages/cli/src/services/snapshot-service.ts
```diff
@@ -0,0 +1,129 @@
...
+  // https://playwright.dev/docs/api/class-testconfig#test-config-snapshot-path-template
+  const snapshotFiles = findFilesRecursively(`${scriptFilePath}-snapshots`)
+  return snapshotFiles.map(absolutePath => ({
+    absolutePath,
+    path: pathToPosix(path.relative(projectBasePath, absolutePath)),
+  }))
+}
+
+export async function uploadSnapshots (project: Project) {
+  // TODO: Processing the whole project at ocne let's us reuse an httpsAgent and make a single
```
clample, line 85: I might have done some premature optimization here at the expense of making the code more complicated. Will take a closer look.

packages/cli/src/services/snapshot-service.ts
```diff
@@ -0,0 +1,65 @@
...
+      const { data: contentStream } = await checklyStorage.download(snapshot.key)
+      contentStream.pipe(fileStream)
+      await stream.finished(contentStream)
+    }
+  } catch (err: any) {
+    throw new Error(`Error downloading snapshots: ${err.message}`)
+  }
+}
+
+export function detectSnapshots (projectBasePath: string, scriptFilePath: string) {
```
clample, line 37: TODO:

Playwright is able to maintain different snapshot files for different platforms (f.ex `*-chromium-linux.png` for Chromium Linux, which is what we use in our runtimes).  

This current implementation will detect all of the snapshots, though, rather than just pushing the Chromium Linux ones.

packages/cli/src/commands/test.ts
```diff
@@ -94,6 +95,12 @@ export default class Test extends AuthCommand {
       char: 'n',
       description: 'A name to use when storing results in Checkly with --record.',
     }),
+    'update-snapshots': Flags.boolean({
```
umutuzgur, line 98: We can also add the char `u` option

packages/cli/src/commands/test.ts
```diff
@@ -233,10 +242,15 @@ export default class Test extends AuthCommand {
       reporters.forEach(r => r.onSchedulingDelayExceeded())
     })
 
-    runner.on(Events.CHECK_SUCCESSFUL, (checkRunId, check, result, links?: TestResultsShortLinks) => {
+    runner.on(Events.CHECK_SUCCESSFUL, async (checkRunId, check, result, links?: TestResultsShortLinks) => {
```
umutuzgur, line 245: The even emitter doesn't support async operations. This will lead to race conditions

packages/cli/src/commands/test.ts
```diff
@@ -233,10 +242,15 @@ export default class Test extends AuthCommand {
...
       if (result.hasFailures) {
         process.exitCode = 1
       }
+
+      if (updateSnapshots) {
+        await pullSnapshots(configDirectory, result.assets?.snapshots)
+      }
```
umutuzgur, line 252: Given we will write/update stuff in the local env, maybe we should ask a confirmation from the user at end of the test session

packages/cli/src/commands/test.ts
clample: Good idea, I added it

packages/cli/src/commands/test.ts
clample: I'm not so sure. I think that the whole point of `--update-snapshots` is to pull down the snapshots. I think that the user passing `--update-snapshots` is already enough of a confirmation to do this. I double checked and PWT also doesn't ask for additional confirmation.

packages/cli/src/services/util.ts
```diff
@@ -43,6 +43,35 @@ export async function walkDirectory (
   }
 }
 
+export function findFilesRecursively (directory: string, ignoredPaths: Array<string> = []) {
```
umutuzgur, line 46: Don't we have a directory walker somewhere? I think we need a more flexible solution to support multiple scenarios. This is already the case at the runners IMO

clample: We have `walkDirectory` here in `util.ts`. It's unused, though, and calls a callback rather than giving a list of matching files. I'll probably delete it.

`findFilesRecursively` should be basically the same as the one we have in runners, it's just missing the `filePredicate` arg. Should I go ahead and proactively add the `filePredicate` arg? Is there anything else that I should add to this one?

umutuzgur (approved):

maxigimenez (approved):

PR #883, opened 2023-11-07
feat: enable the default snapshot support [gh-0]

umutuzgur:

## Affected Components
* [x] CLI
* [ ] Create CLI
* [ ] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer
Enable snapshot support by default for test and the deploy flows

### DIFF
--- a/packages/cli/src/commands/test.ts
+++ b/packages/cli/src/commands/test.ts
@@ -23,6 +23,7 @@ import { createReporters, ReporterType } from '../reporters/reporter'
 import commonMessages from '../messages/common-messages'
 import { TestResultsShortLinks } from '../rest/test-sessions'
 import { printLn, formatCheckTitle, CheckStatus } from '../reporters/util'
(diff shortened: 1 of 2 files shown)


packages/cli/src/commands/test.ts
```diff
@@ -200,6 +199,13 @@ export default class Test extends AuthCommand {
         return check
       })
 
+    for (const check of checks) {
```
clample, line 202: 👍 Thanks for fixing this

clample (approved):
Looks good!

Tim made a ticket by the way: https://app.shortcut.com/checkly/story/18187/make-update-snapshots-flags-visible-on-cli-test-and-deploy-commands

packages/cli/src/commands/test.ts
umutuzgur: I got you!

PR #893, opened 2023-11-22
feat: add ability to filter by group tag

### LINKED ISSUE #878: feat: support group tags when filtering `checkly test --tags`

Currently the `--tags` flag for `npx checkly test` only applies to check tags. We should make the filtering also apply to group tags. If a group matches the tags filter, then all checks in the group should be run as well.

This is particularly important for users using glob patterns to create browser checks in a group, since there's no way to set tags on the automatically generated browser checks.

ferrandiaz:

## Affected Components
* [x] CLI
* [ ] Create CLI
* [x] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer

The check only provides groupId information, hence the part of trying to get the group from the check ref

> Resolves #878

### DIFF
--- a/packages/cli/src/commands/test.ts
+++ b/packages/cli/src/commands/test.ts
@@ -16,7 +16,7 @@ import { loadChecklyConfig } from '../services/checkly-config-loader'
 import { filterByFileNamePattern, filterByCheckNamePattern, filterByTags } from '../services/test-filters'
 import type { Runtime } from '../rest/runtimes'
 import { AuthCommand } from './authCommand'
-import { BrowserCheck, Check, HeartbeatCheck, Session } from '../constructs'
+import { BrowserCheck, Check, HeartbeatCheck, Project, Session } from '../constructs'
 import type { Region } from '..'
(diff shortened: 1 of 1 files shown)


packages/cli/src/commands/test.ts
```diff
@@ -181,7 +181,13 @@ export default class Test extends AuthCommand {
         return filterByCheckNamePattern(grep, check.name)
       })
       .filter(([, check]) => {
-        return filterByTags(targetTags?.map((tags: string) => tags.split(',')) ?? [], check.tags)
+        const tags = check.tags ??= []
```
umutuzgur, line 184: Why did you go with `??=`? This assigns an empty array to check.tags if it is empty first and then assigns the same value to the `tags` variable. I'm curious if there was a special reason

umutuzgur: Then I would suggest just using `??` to not modify the check and groups data, even though it wouldn't be a big problem 🤷‍♂️

ferrandiaz: yup, did that, thanks for the tip!

umutuzgur (approved):
LGTM

Issue #901, opened 2023-12-06
feat: support configuring alert escalation policies


clample:
### What problem does this feature solve?

We should add support for configuring a checks alert escalation policy:

![image](https://github.com/checkly/checkly-cli/assets/10483186/830ca081-47e1-49ca-be0e-dadfdeec553f)

PR #922, opened 2024-01-12
feat: add alert escalation policy as part of the check/config

ferrandiaz:

## Affected Components
* [x] CLI
* [ ] Create CLI
* [ ] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer
Alert Escalation policy can be defined either on the check, the check group or the checkly config file

Resolves https://github.com/checkly/checkly-cli/issues/901


ferrandiaz:
I see one test failing due to `Unable to run checks: \"checkRunJobs[0].useGlobalAlertSettings\" is not allowed` , I suspect this is going to be fixed the moment I merge this other PR https://github.com/checkly/checkly-backend/pull/5075

packages/cli/src/constructs/check-group.ts
```diff
@@ -166,7 +170,8 @@ export class CheckGroup extends Construct {
     // `frequency` is not a CheckGroup resource property. Not present in synthesize()
     this.frequency = props.frequency
     this.apiCheckDefaults = { ...defaultApiCheckDefaults, ...props.apiCheckDefaults }
-
+    this.alertSettings = props.alertEscalationPolicy
+    this.useGlobalAlertSettings = !this.alertSettings
```
umutuzgur, line 174: @tnolet what do you think about this? Basically we set `useGlobalAlertSettings` to `true` if `alertSettings` doesn't exist and vice versa

tnolet: @umutuzgur @ferrandiaz hmmmm, feels a bit iffy / hidden magic. Not sure if we have a great alternative.

clample: Just my 2 cents, but I think it's nice. As a user, I think that it would be confusing if I passed `alertSettings`, but they weren't actually applied because I need to also explicitly set `useGlobalAlertSettings`. I'm not sure there's a case in the CLI were I would pass `alertSettings` but not want it to actually be used.

tnolet: @clample yes, I agree. This seems to be the best UX. @umutuzgur @ferrandiaz let's go with this.

umutuzgur (approved):

PR #928, opened 2024-01-24
feat: move playwright config at check level

ferrandiaz:

## Affected Components
* [x] CLI
* [x] Create CLI
* [ ] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer
1. Move the playwright configuration on our checkly configuration at a check level
2. Made it so that multistep checks can also use the playwright config

### DIFF
--- a/packages/cli/src/services/checkly-config-loader.ts
+++ b/packages/cli/src/services/checkly-config-loader.ts
@@ -6,16 +6,12 @@ import { Session } from '../constructs'
 import { Construct } from '../constructs/construct'
 import type { Region } from '..'
 import { ReporterType } from '../reporters/reporter'
-import { BrowserPlaywrightDefaults } from '../constructs/browser-defaults'
 import * as fs from 'fs'
+import { PlaywrightConfig } from '../constructs/playwright-config'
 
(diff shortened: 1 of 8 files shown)


packages/cli/src/services/checkly-config-loader.ts
```diff
@@ -6,17 +6,13 @@ import { Session } from '../constructs'
 import { Construct } from '../constructs/construct'
 import type { Region } from '..'
 import { ReporterType } from '../reporters/reporter'
-import { BrowserPlaywrightDefaults } from '../constructs/browser-defaults'
 import * as fs from 'fs'
+import { PlaywrightConfig } from '../constructs/playwright-config'
 
 export type CheckConfigDefaults = Pick<CheckProps, 'activated' | 'muted' | 'doubleCheck'
```
umutuzgur, line 12: I think `playwrightConfig` should be here, right?

ferrandiaz: Playwright config is not part of the CheckProps to avoid setting it on all the checks.

umutuzgur (approved):
LGTM

PR #940, opened 2024-03-06
feat: support multistep glob pattern [sc-18837]

maxigimenez:

## Affected Components
* [x] CLI
* [ ] Create CLI
* [ ] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer

> Resolves #[issue-number]

### DIFF
--- a/packages/cli/src/constructs/check-group.ts
+++ b/packages/cli/src/constructs/check-group.ts
@@ -16,6 +16,8 @@ import type { Region } from '..'
 import type { Frequency } from './frequency'
 import type { RetryStrategy } from './retry-strategy'
 import { AlertEscalation } from './alert-escalation-policy'
+import { MultiStepCheck } from './multi-step-check'
+import CheckTypes from '../constants'
 
 const defaultApiCheckDefaults: ApiCheckDefaultConfig = {
   headers: [],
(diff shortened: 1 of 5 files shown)


packages/cli/src/constructs/check-group.ts
```diff
@@ -189,14 +191,21 @@ export class CheckGroup extends Construct {
     this.browserChecks = props.browserChecks
     const fileAbsolutePath = Session.checkFileAbsolutePath!
     if (props.browserChecks?.testMatch) {
-      this.__addChecks(fileAbsolutePath, props.browserChecks.testMatch)
+      this.__addChecks(fileAbsolutePath, props.browserChecks.testMatch, CheckTypes.BROWSER)
+    }
+    if (props.multiStepChecks?.testMatch) {
```
clample, line 196: Just checking - this property was already added in this previous PR, but it didn't do anything?

https://github.com/checkly/checkly-cli/pull/868/files#diff-5679022c8ca8164d11865b9bd6417b09260e3c4bbbf9386c468c3f53af16ccd8R36-R41

clample, line 196: Looking at that PR I also noticed that we're using the check group's browser check defaults as multistep check defaults. I think it makes sense to fix that: https://github.com/checkly/checkly-cli/pull/868/files#diff-5679022c8ca8164d11865b9bd6417b09260e3c4bbbf9386c468c3f53af16ccd8R250-R254

packages/cli/src/services/checkly-config-loader.ts
```diff
@@ -50,6 +50,15 @@ export type ChecklyConfig = {
        */
       testMatch?: string | string[],
     },
+    /**
+     * Multistep checks default configuration properties.
+     */
+    multistepChecks?: CheckConfigDefaults & {
```
clample, line 56: For the naming convention - It looks like we're already exposing `multiStep` arguments and `MultiStepCheck` as the construct

I think we should make sure to capitalize `Step` everywhere then in order to stay consistent.

https://github.com/checkly/checkly-cli/blob/bfc68daeeacc92cfe427e20305a2a36dc0fac043/packages/cli/src/constructs/check-group.ts#L93

https://github.com/checkly/checkly-cli/blob/bfc68daeeacc92cfe427e20305a2a36dc0fac043/packages/cli/src/constructs/multi-step-check.ts#L28

packages/cli/src/constructs/check-group.ts
maxigimenez: Good catch, looks like was not in use

clample (approved):
Looks good 👍

PR #944, opened 2024-03-25
fix: allow disabling smart retries [gh-943]

### LINKED ISSUE #943: bug: default retry strategy is added without having any strategy defined
### Node.js version

16.17.0

### NPM version

8.15.0

### @checkly/cli version

4.6.2

### Steps to reproduce

Create/update a check or group without defining a `retryStrategy` at any level

### What is expected?

The check or group will not have any retry strategy

A fixed default strategy is added and can not be removed.

### Any additional comments?

_No response_

clample:

## Affected Components
* [x] CLI
* [ ] Create CLI
* [ ] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer
Resolves #943 
 
Currently it's not straightforward to disable smart retries. Setting `retryStrategy: null` for a check gives a type error when using TypeScript (`type 'null' is not assignable to type 'RetryStrategy | undefined'`). When using JS, the [backend Joi default of `doubleCheck: true`](https://github.com/checkly/checkly-backend/blob/cd20f50db575b121052eb016d57b894f9a67f77b/api/src/modules/public-api/checkly-cli-schema.js#L317-L320) shows in the UI as a linear retry strategy. Currently to disable all retries you need to set both `retryStrategy: null` and `doubleCheck: false`.

This PR makes it simpler to disable smart retries. To disable smart retries, the user can just set `retryStrategy: null` on a check or a group - there's no need to also set `doubleCheck: false`. When a user sets `retryStrategy: null` and doesn't set `doubleCheck`, the CLI will set `doubleCheck: false` to avoid the Joi default.

This approach of fixing the issue on the CLI side should minimize the breaking changes for users. The only breaking change is that any users that had `retryStrategy: null` and that weren't setting explicitly setting `doubleCheck` will have their retries disabled when upgrading to the new CLI version (since the `doubleCheck: true` default is removed). This shouldn't really have an impact, though.

The PR also adds a `RetryStrategyBuilder.noRetries()` method to make it more clear/explicit how to disable retries.

### DIFF
--- a/packages/cli/src/constructs/retry-strategy.ts
+++ b/packages/cli/src/constructs/retry-strategy.ts
@@ -55,6 +55,10 @@ export class RetryStrategyBuilder {
     return RetryStrategyBuilder.retryStrategy('EXPONENTIAL', options)
   }
 
+  static noRetries (): null {
+    return null
+  }
+
   private static retryStrategy (type: RetryStrategyType, options?: RetryStrategyOptions): RetryStrategy {
     return {
       type,
--- a/packages/cli/src/constructs/retry-strategy.ts
+++ b/packages/cli/src/constructs/retry-strategy.ts
@@ -55,6 +55,9 @@ export class RetryStrategyBuilder {
     return RetryStrategyBuilder.retryStrategy('EXPONENTIAL', options)
   }
 
+  /**
+   * No retries are performed.
+   */
   static noRetries (): null {
     return null
   }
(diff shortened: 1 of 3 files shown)


packages/cli/src/constructs/retry-strategy.ts
```diff
@@ -55,6 +55,13 @@ export class RetryStrategyBuilder {
     return RetryStrategyBuilder.retryStrategy('EXPONENTIAL', options)
   }
 
+  /**
+   * No retries are performed.
+   */
+  static noRetries (): null {
```
clample, line 61: I couldn't think of a good name that matches the `<something>Strategy` pattern that we have.

PR #948, opened 2024-05-06
fix typo in example repos readmes

MariadeAnton:
Replace `npx check deploy` with `npx checkly deploy` in all examples' READMEs. 

## Affected Components
* [x] Examples


tnolet (approved):
@MariadeAnton thanks for catching that amazing set of errors.

PR #952, opened 2024-07-05
feat: add support for test retries [sc-20570]

clample:

## Affected Components
* [ ] CLI
* [ ] Create CLI
* [ ] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer
This PR adds support for retrying failed tests with `npx checkly test` and `npx checkly trigger`. Users can configure retries by either passing the `--retries=<numRetries>` flag or by setting `retries` in their Checkly config file.

### Check state changes

Previously the CLI tracked the state of a check run by using the `CheckRunId`. When MQTT/WebSocket updates were received, the CLI would look up which check it was for based on the `CheckRunId`. 
 
 Since we now have retries, the `CheckRunId` can be different with each retry of a check. In order to track the check state, this PR switches the CLI to use `SequenceId`. This is stable across all retries. `CheckRunId` can then be used to track the progress of a particular run.
 
 In `abstract-check-runner.js` this means that the PR switches to looking up the `sequenceId` for incoming MQTT messages and using this for tracking the check state. The `abstract-list.ts` reporter is also updated to track check state using `sequenceId`.
 
This PR also introduces a new check state `CheckStatus.RETRIED` for indicating that a check is being retried. When a check is retried, we leave it in the `CheckStatus.RETRIED` state rather than switching it back to `CheckStatus.SCHEDULING`/`CheckStatus.RUNNING`. The lifecycle of a check that's retried will then look like: `SCHEDULING` -> `RUNNING` -> `RETRIED` -> `FAILED`/`SUCCESSFUL`.
 
 ### Reporters
 
 #### GitHub reporter
No changes are made to the GitHub reporter. It will simply show the same pass/fail data that it normally does. We could include the number of retries, but there isn't so much horizontal screen space in the GitHub UI to add another column.
 
 #### Dot reporter
 No changes. Will just indicate passed/failed after all retries are finished.
 
 #### Json reporter
 Includes the results for the last check run. Will also include the number of retries. 
 
 #### CI + List Reporter
Prints out info from any retry attempts and shows retrying checks in the summary. Basically following the Notion doc spec. Will add a video to the PR.
 
 ### Running
 
The retry support depends on https://github.com/checkly/checkly-runners/pull/1877. For now, it's only possible to run locally. When running, you also need to manually set `CHECKLY_CLI_VERSION=4.8.0` (the planned next version), since the backend relies on this to use the new MQTT topic format.

🟢  It's expected that the tests are failing until the runners PR is released

### DIFF
--- a/packages/cli/src/commands/test.ts
+++ b/packages/cli/src/commands/test.ts
@@ -282,7 +282,8 @@ export default class Test extends AuthCommand {
           sourceFile: check.getSourceFile(),
           ...result,
         }, testResultId, links))
-    })
+      })
+
     runner.on(Events.CHECK_FAILED, (sequenceId: SequenceId, check, message: string) => {
       reporters.forEach(r => r.onCheckEnd(sequenceId, {
         ...check,
--- a/packages/cli/src/commands/test.ts
+++ b/packages/cli/src/commands/test.ts
@@ -8,7 +8,7 @@ import {
   Events,
   RunLocation,
   PrivateRunLocation,
-  CheckRunId,
+  SequenceId,
   DEFAULT_CHECK_RUN_TIMEOUT_SECONDS,
 } from '../services/abstract-check-runner'
 import TestRunner from '../services/test-runner'
@@ -16,7 +16,7 @@ import { loadChecklyConfig } from '../services/checkly-config-loader'
 import { filterByFileNamePattern, filterByCheckNamePattern, filterByTags } from '../services/test-filters'
 import type { Runtime } from '../rest/runtimes'
 import { AuthCommand } from './authCommand'
-import { BrowserCheck, Check, HeartbeatCheck, MultiStepCheck, Project, Session } from '../constructs'
+import { BrowserCheck, Check, HeartbeatCheck, MultiStepCheck, Project, RetryStrategyBuilder, Session } from '../constructs'
 import type { Region } from '..'
 import { splitConfigFilePath, getGitInformation, getCiInformation, getEnvs } from '../services/util'
 import { createReporters, ReporterType } from '../reporters/reporter'
@@ -26,6 +26,7 @@ import { printLn, formatCheckTitle, CheckStatus } from '../reporters/util'
 import { uploadSnapshots } from '../services/snapshot-service'
 
 const DEFAULT_REGION = 'eu-central-1'
+const MAX_RETRIES = 3
 
 export default class Test extends AuthCommand {
   static coreCommand = true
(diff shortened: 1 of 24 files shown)


examples/boilerplate-project/checkly.config.ts
```diff
@@ -23,7 +23,7 @@ const config = defineConfig({
     /** The Checkly Runtime identifier, determining npm packages and the Node.js version available at runtime.
      * See https://www.checklyhq.com/docs/cli/npm-packages/
      */
-    runtimeId: '2024.02',
+    runtimeId: '2023.09',
```
clample, line 26: For local testing. TODO - reset.

packages/cli/src/services/abstract-check-runner.ts
```diff
@@ -116,38 +118,47 @@ export default abstract class AbstractCheckRunner extends EventEmitter {
...
+      this.emit(Events.CHECK_INPROGRESS, check, sequenceId)
+    } else if (subtopic === 'result') {
+      const { result, testResultId, resultType } = message
       await this.processCheckResult(result)
       const links = testResultId && result.hasFailures && await this.getShortLinks(testResultId)
-      this.emit(Events.CHECK_SUCCESSFUL, checkRunId, check, result, links)
-      this.emit(Events.CHECK_FINISHED, check)
+      if (resultType === 'FINAL') {
+        this.disableTimeout(sequenceId)
+        this.emit(Events.CHECK_SUCCESSFUL, sequenceId, check, result, testResultId, links)
```
clample, line 154: Before, we would pass the `testResultId` to the reporters at the start of the test session. 

With retries, though, we don't actually know the `testResultId` of the final test result until all of the retries are finished. Need to switch to passing it on the `CHECK_SUCCESSFUL` event.

The `testResultId` is used to generate links to the test result in the reporters.

packages/cli/src/commands/test.ts
```diff
@@ -100,6 +101,9 @@ export default class Test extends AuthCommand {
       description: 'Update any snapshots using the actual result of this test run.',
       default: false,
     }),
+    retries: Flags.integer({
+      description: 'How many times to retry a failing test run.',
```
clample, line 105: TODO: We should document the max number of retries here

clample: Done ✅ 

@ferrandiaz as discussed, I added the max number of retries to the help output. Also added in the default there.

The brackets in front are to be consistent with the formatting we already have for defaults.

ferrandiaz (approved):

ferrandiaz (approved):

PR #954, opened 2024-07-19
feat: increase test session timeout

clample:

## Affected Components
* [ ] CLI
* [ ] Create CLI
* [ ] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer
Currently both `npx checkly test` and `npx checkly trigger` support a `--timeout` flag. This is a fallback timeout to make sure that if the Checkly backend doesn't respond with a check result for some reason, the user's CI doesn't become stuck indefinitely.

When it was originally added, there were few users of the CLI and checks ran on AWS Lambda which was able to run all checks concurrently. Now, users have very large test suites which can take longer, and have sometimes bumped into the default 300 second timeout. We'll also be adding test retries which will further increase the time tests take.

This PR simply bumps the default timeout to 10 minutes so that users are less likely to run into it.


ferrandiaz (approved):
⏲️

PR #964, opened 2024-08-28
feat: add environment secrets support [sc-21028]

maxigimenez:

## Affected Components
* [x] CLI
* [ ] Create CLI
* [ ] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer
- [x] Add `--secret` support while creating/updating environment variables with `npx checkly env add|update`
- [x] Add `secret: boolean` support for check/group constructs
- [x] Update `npx checkly env pull` to structure the file with regular and secret variables

> [!IMPORTANT]
> Important note: this feature is not available yet, if you have any questions feel free to reach out to our customer support

### DIFF
--- a/packages/cli/src/commands/env/add.ts
+++ b/packages/cli/src/commands/env/add.ts
@@ -14,7 +14,7 @@ export default class EnvAdd extends AuthCommand {
     }),
     secret: Flags.boolean({
       char: 's',
-      description: 'Indicate that the environment variable will be secret, the value will not be revealed anywhere.',
+      description: 'Indicate that the environment variable will be secret.',
       default: false,
       exclusive: ['locked'],
     }),
--- a/packages/cli/src/commands/env/add.ts
+++ b/packages/cli/src/commands/env/add.ts
@@ -12,6 +12,12 @@ export default class EnvAdd extends AuthCommand {
       description: 'Indicate that the environment variable will be locked.',
       default: false,
     }),
+    secret: Flags.boolean({
+      char: 's',
+      description: 'Indicate that the environment variable will be secret, the value will not be revealed anywhere.',
+      default: false,
+      exclusive: ['locked'],
+    }),
   }
 
   static args = {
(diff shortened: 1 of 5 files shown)


packages/cli/src/commands/env/add.ts
```diff
@@ -12,6 +12,12 @@ export default class EnvAdd extends AuthCommand {
       description: 'Indicate that the environment variable will be locked.',
       default: false,
     }),
+    secret: Flags.boolean({
+      char: 's',
+      description: 'Indicate that the environment variable will be secret.',
+      default: false,
+      exclusive: ['locked'],
```
drakirnosslin, line 19: Do you need `exclusive` on only one of the flags, or should it be on the locked one too?

maxigimenez: On one flag is enough

drakirnosslin: TIL, thanks!

Antoine-C:
Reviewing 👀

packages/cli/src/commands/env/add.ts
```diff
@@ -40,7 +46,7 @@ export default class EnvAdd extends AuthCommand {
       envValue = await ux.prompt(`What is the value of ${envVariableName}?`, { type: 'mask' })
     }
     try {
-      await api.environmentVariables.add(envVariableName, envValue, locked)
+      await api.environmentVariables.add(envVariableName, envValue, locked, secret)
       this.log(`Environment variable ${envVariableName} added.`)
```
Antoine-C, line 50: Do we want to tweak this a bit to say something like `Secret environment variable added` in case the `s` flag is used?

Antoine-C (approved):
Just a nit, otherwise LGTM

packages/cli/src/commands/env/add.ts
maxigimenez: Updated

Antoine-C (approved):

PR #980, opened 2024-11-13
feat: compress large JSON requests [sc-22310]

sorccu:
When you deploy or test a project, a potentially very large JSON payload is sent to the API server. In large projects, the time to upload the payload can be quite significant (depending on connection speed). This PR introduces gzip compression to payloads sent to relevant endpoints, which should make the payload easier to upload.

Additionally, those large payloads were stringified fully in memory, causing unnecessarily high memory usage. This PR changes the payload to be stringified and streamed with json-stream-stringify instead, which should decrease memory usage. This change only applies to requests that are now also getting compressed.

## Affected Components
* [x] CLI
* [ ] Create CLI
* [ ] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer

> Resolves #[issue-number]

## New Dependency Submission

- `json-stream-stringify`: This package is dependency-free, and perhaps the only package providing the needed functionality to stream JSON payloads.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10927,6 +10927,14 @@
       "dev": true,
       "license": "MIT"
     },
+    "node_modules/json-stream-stringify": {
+      "version": "3.1.6",
+      "resolved": "https://registry.npmjs.org/json-stream-stringify/-/json-stream-stringify-3.1.6.tgz",
+      "integrity": "sha512-x7fpwxOkbhFCaJDJ8vb1fBY3DdSa4AlITaz+HHILQJzdPMnHEFjxPwVUi1ALIbcIxDE0PNe/0i7frnY8QnBQog==",
+      "engines": {
+        "node": ">=7.10.1"
+      }
+    },
     "node_modules/json-stringify-nice": {
       "version": "1.1.4",
       "dev": true,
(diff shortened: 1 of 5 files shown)


umutuzgur:
@sorccu nice PR. Out of curiosity, do you have any numbers in terms of the asset size reduction?

clample:
Great PR ➕, looks good to merge to me.

One possible side effect - we protect checkly-api from high memory usage by setting a limit on the payload size: `checkly-api public-api/test-sessions/routes.js#L24-L25`. The limit applies to the raw payload size - so users can now send more data and put more memory pressure on checkly-api. If the compression ratio is really high, maybe we should keep it in mind and add a limit on the uncompressed payload size somewhere.

sorccu:
A reasonable estimate would be that compression should reduce the payload to about 25% or less. I do not currently have a large example project to try with, but since the code is essentially a combination of JSON and code snippets, we should be able to roughly estimate the performance by comparing a JavaScript or TypeScript codebase as plain text vs compressed. Using the CLI repository as an example:

```
➜ fd -e ts -e js | xargs cat | wc
   14026   47822  478588
   
➜ fd -e ts -e js | xargs cat | gzip | wc
     390    2960   99972
```

Results in a reduction to about ~21% of original size.

Note that in this example, each file is present only once. In checks, that may not be the case, so actual performance may be even better.

Note that enabling compression does not automatically make existing check suites larger. Yes, it may allow larger test suites in the future, but it's not like load will suddenly increase. Anyway, if this is a concern, the backend can also be updated to process the payload in streaming fashion.

umutuzgur (approved):
LGTM

PR #984, opened 2024-11-26
feat: allow configuring the CLI to error on no matching tests

imbstack:

## Affected Components
* [x] CLI
* [ ] Create CLI
* [ ] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer

This will be useful for us to assert results from this in CI and such but I would understand if you'd prefer to not take it. Also happy to change the name, etc.

### DIFF
--- a/packages/cli/e2e/__tests__/trigger.spec.ts
+++ b/packages/cli/e2e/__tests__/trigger.spec.ts
@@ -65,4 +65,20 @@ describe('trigger', () => {
     expect(result.stdout).toContain('No matching checks were found.')
     expect(result.status).toBe(0)
   })
+
+  test('Should return code 1 when no checks match and the fail-on-no-match flag is set', async () => {
+    const result = await runChecklyCli({
+      args: [
+        'trigger',
+        '--tags',
+        'no-checks-match-this-tag',
+        '--fail-on-no-matching',
+      ],
+      apiKey: config.get('apiKey'),
+      accountId: config.get('accountId'),
+    })
+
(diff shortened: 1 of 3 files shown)


sorccu:
Thanks @imbstack, I'll be taking a look at this soon!

packages/cli/e2e/__tests__/trigger.spec.ts
```diff
@@ -65,4 +65,20 @@ describe('trigger', () => {
     expect(result.stdout).toContain('No matching checks were found.')
     expect(result.status).toBe(0)
   })
+
+  test('Should return code 1 when no checks match and the fail-on-no-match flag is set', async () => {
+    const result = await runChecklyCli({
+      args: [
+        'trigger',
+        '--tags',
```
sorccu, line 73: It may be trying to treat the `--fail-on-no-matching` flag as a tag right now because `--tags` is not given a value.

```suggestion
```

sorccu (changes requested):
Please check my comment.

packages/cli/e2e/__tests__/trigger.spec.ts
imbstack: Ah yah, good call. I have fixed the test I think!

sorccu (approved):

sorccu:
Workflows will fail due to secrets not getting passed to PRs from forks. I will merge later regardless.

PR #1001, opened 2024-12-10
chore!: update node engine requirement to >=18 [sc-23291]

sorccu:
The reason for this change is that new versions of many of our dependencies now require Node 18 or later, meaning we cannot update them without breaking compatibility. Bumping our Node version in a major release makes that possible again.

This is a breaking change (despite probably no one using Node 16 anymore), so a new major version is required.

## Affected Components
* [x] CLI
* [x] Create CLI
* [ ] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer

> Resolves #[issue-number]

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -18788,7 +18788,7 @@
         "typescript": "5.3.3"
       },
       "engines": {
-        "node": ">=16.0.0"
+        "node": ">=18.0.0"
       }
     },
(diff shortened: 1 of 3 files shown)


hdf1996:
Additionally, looks like #991 already took care of updating the github actions to use 18.x, so this should be fine

PR #1003, opened 2024-12-13
feat: support wildcard exports when parsing dependency tree [sc-22632]

sorccu:
This PR makes the project parser compatible with wildcard exports such as `export * from './file'`. Previously, only named exports such as `export { Thing } from './file'` were supported. The end result was that the file being referred to in the export would not be added to the dependency tree.

## Affected Components
* [x] CLI
* [ ] Create CLI
* [x] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer

> Resolves #[issue-number]


umutuzgur (approved):
LGTM

PR #1007, opened 2025-01-12
feat: check constructs against account default runtime if not given [sc-22843]

sorccu:
The CLI had a hardcoded default runtime value of `2024.02`, which became the effective value if no runtime was set at check or project level. Now, the account's default runtime is used instead. Additionally, if the account's default runtime is used, then we leave `runtimeId` undefined when synthesizing resources. This allows the user to have their checks always use the current account default runtime if they wish. If they do not wish to have such behavior, they should set a default runtime at project level as usual.

Technically speaking this is a potentially breaking change if a user has something other than `2024.02` set as their default account runtime, and have never set a runtime at project or check level. When they next deploy their checks (or run `checkly test`), a different runtime will be used. However, this kind of usage should be rare, and the user may simply change their default account runtime or define the runtime in the CLI project to restore the previous runtime, if needed.

## Affected Components
* [x] CLI
* [ ] Create CLI
* [x] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer

> Resolves #[issue-number]

### DIFF
--- a/packages/cli/src/services/project-parser.ts
+++ b/packages/cli/src/services/project-parser.ts
@@ -23,12 +23,12 @@ type ProjectParseOpts = {
   checkDefaults?: CheckConfigDefaults,
   browserCheckDefaults?: CheckConfigDefaults,
   availableRuntimes: Record<string, Runtime>,
+  defaultRuntimeId: string,
   verifyRuntimeDependencies?: boolean,
   checklyConfigConstructs?: Construct[],
 }
 
 const BASE_CHECK_DEFAULTS = {
-  runtimeId: '2024.02',
 }
 
 export async function parseProject (opts: ProjectParseOpts): Promise<Project> {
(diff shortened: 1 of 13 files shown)


maxigimenez (approved):
LGTM 👍

packages/cli/src/services/project-parser.ts
```diff
@@ -23,12 +23,12 @@ type ProjectParseOpts = {
   checkDefaults?: CheckConfigDefaults,
   browserCheckDefaults?: CheckConfigDefaults,
   availableRuntimes: Record<string, Runtime>,
+  defaultRuntimeId: string,
   verifyRuntimeDependencies?: boolean,
   checklyConfigConstructs?: Construct[],
 }
 
 const BASE_CHECK_DEFAULTS = {
```
maxigimenez, line 31: nit. this can be removed

sorccu: Yeah I know but I decided to leave it in order to keep the PR a bit simpler, and it could potentially be useful for something.

PR #1019, opened 2025-02-17
refactor: add missing `mts` ext. support to checkly config

kiroushi:

## Affected Components
* [x] CLI
* [ ] Create CLI
* [ ] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer
Currently the CLI is able to read both JS/TS files, or `.mjs` when the app is declared as `commonjs` and needs to be interpreted as a ESM; but is missing the ability to do so when the config is written in TypeScript.

This PR adds the `.mts` extension to the list of candidates to read the config from to allow certain settings like https://www.typescriptlang.org/tsconfig/#verbatimModuleSyntax to work while in commonjs apps.

### DIFF
--- a/packages/cli/src/services/__tests__/checkly-config-loader.spec.ts
+++ b/packages/cli/src/services/__tests__/checkly-config-loader.spec.ts
@@ -23,7 +23,7 @@ describe('loadChecklyConfig()', () => {
       await loadChecklyConfig(configDir)
     } catch (e: any) {
       expect(e.message).toContain(`Unable to locate a config at ${configDir} with ${
-        ['checkly.config.ts', 'checkly.config.js', 'checkly.config.mjs'].join(', ')}.`)
+        ['checkly.config.ts', 'checkly.config.js', 'checkly.config.mts', 'checkly.config.mjs'].join(', ')}.`)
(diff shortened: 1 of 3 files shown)


sorccu:
Are you sure that the current changes are enough? I don't think this works without changing `loadFile` too:

https://github.com/checkly/checkly-cli/blob/90724777fa27727f5c1e5fb122afc991f9999ed5/packages/cli/src/services/checkly-config-loader.ts#L77-L97

Without adding `mts` there, you'll just get an error saying `checkly.config.mts` is unsupported which makes no sense.

Additionally `loadTsFile` may potentially need some changes for modules (perhaps similar to `loadJsFile`), not completely sure about that without trying it out.

Either way I wouldn't really call this a refactor. I would like you to change this to feat instead because that's what it is, and it should therefore come with at least 1 test to prove that it works, or at minimum you must provide some proof that it worked for you locally.

sorccu (changes requested):
See my comment

sorccu:
Also I would like to know the motivation for this change. I agree that it's good but just wondering what made you think of doing it.

PR #1021, opened 2025-02-18
feat: update inline hints and tips

### LINKED ISSUE #973: feat: Show hint to update CLI when there's a new version available
Saves mental energy from users, improves DX :)

When running the CLI with an older version it says:

```bash
›   Warning: checkly update available from 4.7.0 to 4.9.1.
```

It'd be nice if it were to have a hint to remind folks how to update it. 

i.e.

```bash
 ›   Warning: checkly update available from 4.7.0 to 4.9.1.
 ›   use: npm install checkly@latest

### LINKED ISSUE #1018: feat: update `npx checkly test` to `--record` by default

The question is, is this a major version change? How do we make this happen nicely?

Ideally people opt-out to places.

:)

tnolet:

## Affected Components
* [x] CLI
* [x] Create CLI
* [ ] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer
-  adds a hint on relevant reporters to use `--record` if the user did not use it when running `test` or `trigger`
- adds a copy & pastable hint on updating the CLI when there is a new version.
- small wording updates to other parts.

CleanShot 2025-02-18 at 14 32 17@2x

CleanShot 2025-02-18 at 14 51 50@2x

Resolves #1018 
Resolves #973

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,7 +74,7 @@ First, make sure you sign up for a [free Checkly account](https://app.checklyhq.
 Then, the **easiest** way to get started is to install the CLI using the following command:
 
 ```bash
-npm create checkly
+npm create checkly@latest
 ```
 This command will guide you through some simple steps and set up a fully working example project for you. Should take 
 ~1 minute.
--- a/examples/advanced-project-js/README.md
+++ b/examples/advanced-project-js/README.md
@@ -8,7 +8,7 @@ https://checklyhq.com website as a monitoring target.
 3. Test -> Deploy: now you have your app monitored around the clock. All from your code base.
 
 ```
-npm create checkly -- --template advanced-project
+npm create checkly@latest -- --template advanced-project
 ```
 
 ## Project Structure
--- a/examples/advanced-project/README.md
+++ b/examples/advanced-project/README.md
@@ -8,7 +8,7 @@ https://checklyhq.com website as a monitoring target.
 3. Test -> Deploy: now you have your app monitored around the clock. All from your code base.
 
 ```
-npm create checkly -- --template advanced-project
+npm create checkly@latest -- --template advanced-project
 ```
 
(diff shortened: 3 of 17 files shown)


sorccu:
Tangentially related but we should really add that update warning for the create-cli. After you've installed it once it's basically guaranteed to stay the same version unless you lose your package cache somehow (by e.g. updating node). I guess we as the developers are more exposed to that rather than real users who might only ever run it once, but there may be some who try us out, leave, and then come back months later. If they just follow our docs (`npm create checkly`) it'll use whatever old version of the create-cli they already had installed and the user is not informed of that either. Worst case the version they had (and still have) is buggy, they see no improvement and they leave again. It would also be nice if our docs used `npm create checkly@latest` instead which would be another way to solve the problem.

tnolet:
> Tangentially related but we should really add that update warning for the create-cli. After you've installed it once it's basically guaranteed to stay the same version unless you lose your package cache somehow (by e.g. updating node). I guess we as the developers are more exposed to that rather than real users who might only ever run it once, but there may be some who try us out, leave, and then come back months later. If they just follow our docs (`npm create checkly`) it'll use whatever old version of the create-cli they already had installed and the user is not informed of that either. Worst case the version they had (and still have) is buggy, they see no improvement and they leave again. It would also be nice if our docs used `npm create checkly@latest` instead which would be another way to solve the problem.

I think we should indeed update all the links for to `npm create checkly@latest`. I will mark that on my list. Can you add update logic to the `create-cli` package?

tnolet:
@sorccu I've updated some parts to `npm create checkly@latest` where relevant. I will proceed to also update our docs site and the marketing site. Can you review when convenient?

sorccu (approved):
Looks good to me.

PR #1024, opened 2025-02-20
chore: update oclif [sc-23314]

sorccu:
Replaces removed ux.wait() with stdlib, ux.prompt() with prompts.

Minor changes to bin files were also required as otherwise errors were printed with stack traces. I had not realized that the create-cli bin was checking Node version - it has now been updated to Node 18 too. The new oclif also requires Node 18.

The `default` command which was used for the `create-cli` no longer exists and was replaced with the `single` command strategy.

Custom help output had some type changes but is basically the same.

Note: for whatever reason we now MUST set the oclif `topicSeparator`. This makes no sense but if topicSeparator is not present or is set to `":"`, CLI errors such as attempting to run `bin/run foo` in the packages/create-cli folder will error with a stack trace, with the error being thrown and handled (incorrectly) by oclif itself.

## Affected Components
* [x] CLI
* [x] Create CLI
* [ ] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer

> Resolves #[issue-number]


tnolet:
@sorccu tests are still failing. I will hold off on review till they pass.

tnolet:
@sorccu I actually would vote to introduce full stack traces. I find it very annoying to debug our current CLI as it does not print usable runtime exceptions due to the missing stack traces. Maybe we can switch it based on a `DEBUG` env variable?

tnolet (approved):
Smoke tested locally. LGTM

PR #1041, opened 2025-03-10
feat: full ESM config support using Jiti

### LINKED ISSUE #1015: bug: The 'import.meta' meta-property is only allowed when the...
### Node.js version

22

### NPM version

10

### @checkly/cli version

0.4.14

### Steps to reproduce

1. Use `import.meta.dirname`
2. Run `pnpx checkly test`

### What is expected?

For tests to run.

> The 'import.meta' meta-property is only allowed when the '--module' option is 'es2020', 'es2022', 'esnext', 'system', 'node16', or 'nodenext'.
> entrypoint: path.join(import.meta.dirname, 'homepage.spec.ts')

Our `tsconfig.json` already uses `nodenext`, so it is not clear where it is picking up the other value.

### Any additional comments?

_No response_

henrist:
I hereby confirm that I followed the code guidelines found at [engineering guidelines](https://www.notion.so/checkly/Engineering-Guidelines-a3a165a203a04dc1a112f0e26b2f2d3f) <-- I don't have access to this

## Affected Components
* [x] CLI
* [x] Create CLI
* [x] Test
* [ ] Docs
* [x] Examples
* [ ] Other

## Notes for the Reviewer

> Resolves #1015

The current use of `ts-node` is problematic:

- Only CommonJS is supported, so loading ESM doesn't work as expected
- `ts-node` is currently not maintained and haven't received updates for 1+ year. It lacks newer TypeScript features. For instance it does not support multiple extends in `tsconfig.json` that came with TypeScript 5.

This has been a large pain as most of the code I work with is in ESM. It has required non-trivial tsconfig files to workaround this.

This PR introduces [Jiti](https://github.com/unjs/jiti) as an alternative to `ts-node`. I could have picked tsx as well, but I don't find the API of tsx as intuitive/simple as Jiti. This is very similar as the implementation of eslint that uses Jiti for its TypeScript support.

Additional file extensions are also supported: `.mts`, `.cts`, `.cjs`

To avoid a breaking change it supports both `ts-node` and `jiti`. Existing users with only `ts-node` should not be affected, while new users will be recommended to add `jiti`. Jiti does not depend on a `tsconfig.json` file (and does not do type checking), so `typescript` is not requested as a dependency. Existing users can add `jiti` to get better (and ESM) support, and we also recommend this on compile issues. If you're open for a breaking change we could remove `ts-node` support and avoid dealing with both.

### DIFF
--- a/examples/advanced-project/package.json
+++ b/examples/advanced-project/package.json
@@ -11,7 +11,6 @@
   "license": "ISC",
   "devDependencies": {
     "checkly": "latest",
-    "ts-node": "10.9.1",
-    "typescript": "4.9.5"
+    "jiti": "^2"
   }
 }
--- a/examples/boilerplate-project/package.json
+++ b/examples/boilerplate-project/package.json
@@ -11,7 +11,6 @@
   "license": "ISC",
   "devDependencies": {
     "checkly": "latest",
-    "ts-node": "10.9.1",
-    "typescript": "4.9.5"
+    "jiti": "^2"
   }
 }
--- a/package-lock.json
+++ b/package-lock.json
@@ -10537,6 +10537,16 @@
         "url": "https://github.com/chalk/supports-color?sponsor=1"
       }
     },
+    "node_modules/jiti": {
+      "version": "2.4.2",
+      "resolved": "https://registry.npmjs.org/jiti/-/jiti-2.4.2.tgz",
+      "integrity": "sha512-rg9zJN+G4n2nfJl5MW3BMygZX56zKPNVEYYqq7adpmMh4Jn2QNEwhvQlFy6jPVdcod7txZtKHWnyZiA3a0zP7A==",
+      "dev": true,
+      "license": "MIT",
+      "bin": {
+        "jiti": "lib/jiti-cli.mjs"
+      }
+    },
     "node_modules/jmespath": {
       "version": "0.16.0",
       "dev": true,
@@ -18877,6 +18887,7 @@
         "config": "3.3.9",
         "cross-env": "7.0.3",
         "jest": "29.7.0",
+        "jiti": "2.4.2",
         "nanoid": "3.3.4",
         "oclif": "3.7.3",
         "simple-git-hooks": "2.11.1",
@@ -18886,6 +18897,14 @@
       },
       "engines": {
         "node": ">=18.0.0"
+      },
+      "peerDependencies": {
+        "jiti": ">=2"
+      },
+      "peerDependenciesMeta": {
+        "jiti": {
+          "optional": true
+        }
       }
     },
     "packages/cli/node_modules/@oclif/core": {
@@ -19463,6 +19482,7 @@
         "@types/uuid": "9.0.1",
         "config": "3.3.9",
         "jest": "29.7.0",
+        "jiti": "2.4.2",
         "rimraf": "5.0.1",
         "ts-jest": "29.2.4",
         "ts-node": "10.9.1",
@@ -19470,6 +19490,14 @@
       },
       "engines": {
         "node": ">=18.0.0"
+      },
+      "peerDependencies": {
+        "jiti": ">=2"
+      },
+      "peerDependenciesMeta": {
+        "jiti": {
+          "optional": true
+        }
       }
     },
     "packages/create-cli/node_modules/@oclif/core": {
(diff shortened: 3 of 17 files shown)


sorccu:
Hi @henrist,

I really like the change, but I will need some time to consider any possible backwards compatibility implications caused it. I may not be able to properly review this until next week (it's the end of tax season over here which limits my extra time), but based on a quick scroll through I think it looks great overall. Backwards compatibility and potential breakage is my main concern and I'm grateful that you seem to have kept that in mind. For now, we are not ready to fully remove `ts-node` but that may be a possibility in a future major version.

Do note that while this PR may add full ESM support on the CLI side, our runtimes do not yet support ESM due to complexities of the current architecture, so unfortunately this PR does not fully enable ESM on the entire Checkly platform. However, we have a project in progress to allow ESM to work on our runners, so ESM everywhere may soon be a possibility too. For now though it's a limitation that you have to keep in mind.

Thanks a lot!

henrist:
@sorccu Thanks for your quick feedback! Looking forward to your review. I have put some effort into avoiding this to become a breaking change. I've tested it an original template project, new template project and on a couple of our repos. But useful to have someone else double check.

Great to hear that you're working on ESM support on the runner side as well, although due to the limitation of dependencies and the fact that it parses module syntax just fine, this haven't been a real problem for us. I did use the term "config" in the PR title in an attempt to make it clear that this is for config of checks, not the check scripts/specs themself.

Would it be possible to trigger the canary/experimental release for this PR?

sorccu:
I would be happy to trigger a build, but unfortunately that will also have to wait till next week as our workflows do not currently give access to secrets for PRs from forks, which means that both integration tests and the build would fail. I will massage the workflows a little next week to make that work.

sorccu:
Unfortunately I was extra busy last week and did not have bandwidth to test your PR, but I will handle this within the next 2 days. Super excited about this feature.

sorccu:
Unfortunately I had to come up with a temporary janky solution to run the workflows by creating a non-fork PR with your changes in #1048, but since the commit hash is the same, the test results can be seen in this original PR too. I am happy to keep merging any changes to the build PR and make any other changes as needed.

An experimental build is also available: `checkly@0.0.0-pr.1048.4426d48`

However, as can be seen from the result of the test run, there seems to be some sort of an incompatibility. I will look into this shortly.

henrist:
That's great! I'll test out the experimental build on my end tomorrow or Thursday, and also see if I can understand the reason for the failure. I have 30+ repos here with Checkly CLI, however they are pretty homogeneous.

`npm run test` runs fine here on same Node.js version (but I'm on MacOS). Could it be platform-specific problem (the failures were on Windows)? I see the Linux job never ran. Would be nice to set `strategy.fail-fast: false` on the job.

henrist:
Indeed looks like a platform problem as it runs fine on ubuntu-latest. I wonder if this is a Jest issue and a caching bug in Jiti.

[Running it with debug mode](https://github.com/henrist/checkly-cli/actions/runs/14074511276/job/39414961169) shows that under Windows some additional file are being transpiled. It seems under Linux those same files exists in Node.js module cache, but on Windows I think it becomes a mix of path styles so the cache is not used.

Even if we set the `--experimental-vm-modules` option [it won't work](https://github.com/henrist/checkly-cli/actions/runs/14074887913/job/39416017734), since it would load a fresh constructs file so the global state in Session will be different/not so global.

This problem is limited to testing here though, as regular users won't have this behavior since the Checkly CLI code is compiled to JS. A bit sure what would be the best fix. I'll anyways see if I can create a repro for the potential caching bug.

sorccu:
I am setting up a Windows VM to verify that this won't affect Windows users. If there is no issue, then I will attempt to set the proper `NODE_OPTIONS` when running jest and that should be it. And if there is an issue, then we will need some workaround for it as we do have active Windows users.

sorccu:
I have confirmed that there is no issue when using the CLI. I will make a few final changes and merge.

Thanks a lot for this PR, it is massively appreciated. I'm sorry that it has taken so long to get it merged. There has been a lot of stuff going on.

sorccu:
Due to internal needs, I have merged the PR and testing on Windows has been disabled for now. I will rework the jest setup shortly which should help with the issue.

Thanks again!

PR #1074, opened 2025-05-21
feat: parse playwright config dependencies

ferrandiaz:
## Affected Components
* [x] CLI
* [ ] Create CLI
* [ ] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer

We are currently adding the playwright config file but not it's dependency, by parsing it with the other project files we make sure the config dependencies are also added as part of the code bundle

### DIFF
--- a/packages/cli/src/services/check-parser/parser.ts
+++ b/packages/cli/src/services/check-parser/parser.ts
@@ -133,9 +133,9 @@ export class Parser {
     }
   }
 
-  async getFilesAndDependencies (playwrightConfig: PlaywrightConfig):
+  async getFilesAndDependencies (playwrightConfig: PlaywrightConfig, playwrightConfigFilePath: string):
     Promise<{ files: string[], errors: string[] }> {
-    const files = new Set(await this.getFilesFromPaths(playwrightConfig))
+    const files = new Set(await this.getFilesFromPaths(playwrightConfig)).add(playwrightConfigFilePath)
     const errors = new Set<string>()
     const missingFiles = new Set<string>()
     const resultFileSet = new Set<string>()
--- a/packages/cli/src/services/util.ts
+++ b/packages/cli/src/services/util.ts
@@ -194,10 +194,9 @@ Promise<{outputFile: string, browsers: string[], relativePlaywrightConfigPath: s
 
   const [cacheHash] = await Promise.all([
     getCacheHash(dir),
-    loadPlaywrightProjectFiles(dir, pwConfigParsed, include, archive)
+    loadPlaywrightProjectFiles(dir, filePath, pwConfigParsed, include, archive)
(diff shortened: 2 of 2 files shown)


packages/cli/src/services/check-parser/parser.ts
```diff
@@ -133,9 +133,9 @@ export class Parser {
     }
   }
 
-  async getFilesAndDependencies (playwrightConfig: PlaywrightConfig):
+  async getFilesAndDependencies (playwrightConfig: PlaywrightConfig, playwrightConfigFilePath: string):
```
umutuzgur, line 136: Could you instead make `playwrightConfig` aware of its location? I already knows about the `dir`

packages/cli/src/services/playwright-config.ts
```diff
@@ -46,7 +47,7 @@ export class PlaywrightConfig {
   projects?: PlaywrightProject[]
   files: Set<string>
 
-  constructor(dir: string, playwrightConfig: any) {
+  constructor(dir: string, playwrightConfig: any, configFileName: string) {
     this.projectName = ''
```
umutuzgur, line 51: playwrightConfigPath: string

this.dir = path.dirname(playwrightConfigPath)
this.filePath = playwrightConfigPath

packages/cli/src/services/__tests__/playwright-config.spec.ts
```diff
@@ -9,13 +9,13 @@ const fixturesPath = path.join(__dirname, 'fixtures', 'playwright-configs')
...
     const pwConfig = await Session.loadFile(path.join(fixturesPath, 'simple-config.ts'))
-    const config = new PlaywrightConfig(fixturesPath, pwConfig)
+    const config = new PlaywrightConfig(fixturesPath, pwConfig, 'simple-config.ts')
     expect(Array.from(config.testMatch)).toEqual(['**/*.@(spec|test).?(c|m)[jt]s?(x)'])
     expect(config.getBrowsers()).toEqual(['chromium', 'webkit', 'msedge', 'chrome'])
   })
   it('it should load simple config correctly', async () => {
     const pwConfig = await Session.loadFile(path.join(fixturesPath,'simple-config-no-browsers.ts'))
-    const config = new PlaywrightConfig(fixturesPath, pwConfig)
+    const config = new PlaywrightConfig(fixturesPath, pwConfig, 'simple-config-no-browsers.ts')
```
umutuzgur, line 18: How are these tests passing now?

umutuzgur: I guess I get why they are passing but can we fix them as well?

umutuzgur (approved):
LGTM

PR #1080, opened 2025-06-04
feat: introduce CheckGroupV2 which avoids unintuitive implicit defaults [sc-24260]

sorccu:
The current CheckGroup continues to work the same as before, however a warning is output recommending the user to upgrade to CheckGroupV2.

## Affected Components
* [x] CLI
* [ ] Create CLI
* [ ] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer

> Resolves #[issue-number]


packages/cli/src/constructs/check-group-v1.ts
```diff
@@ -0,0 +1,536 @@
...
+ *
+ * The original CheckGroup v1 comes with implicit defaults that are not
+ * intuitive and make it impossible to keep check-level behavior for
+ * certain properties.
+ *
+ * The following properties will always use the group's values if set, or
+ * implicit defaults if not set:
+ *
+ *   - {@link CheckGroupV1Props.alertEscalationPolicy}
+ *   - {@link CheckGroupV1Props.retryStrategy}
```
sujaya-sys, line 272: are we missing "locations" here?

packages/cli/src/constructs/check-group-v1.ts
```diff
@@ -0,0 +1,536 @@
...
+    Session.registerConstruct(this)
+    this.__addSubscriptions()
+    this.__addPrivateLocationGroupAssignments()
+  }
+
+  protected async onBeforeValidate (diagnostics: Diagnostics): Promise<void> {
+    diagnostics.add(new DeprecatedConstructDiagnostic(
+      'CheckGroup',
+      new Error(
+        `Please update to CheckGroupV2 which has more intuitive behavior.`
```
sujaya-sys, line 354: Will provide a link here shortly to add

packages/cli/src/constructs/check-group-v1.ts
sorccu: No because the backend has already been changed in a way that makes it apply to all groups. There is no difference in behavior between `CheckGroup` and `CheckGroupV2`.

PR #1091, opened 2025-06-19
fix: allow all files with the include option [sc-0]

umutuzgur:
## Affected Components
* [ ] CLI
* [ ] Create CLI
* [ ] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer
If the user wants to have specific files, we should just allow what they want


ferrandiaz (approved):
LGTM

PR #1092, opened 2025-06-19
feat: create pw-test command

ferrandiaz:
## Affected Components
* [x] CLI
* [ ] Create CLI
* [ ] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer
- Add a new command `pw-test`
- New command allows you to run whatever playwright command you would like
```bash
npx checkly pw-test -l us-east-1 -- --project foo --grep @bar
```
- following industry standards the commands before the separator `--` are related to checkly, the ones after the separator are related to playwright
```bash
npx checkly pw-test  { checkly flags } -- { playwright flags }
```
- command allows you to run the test on checkly without having a `checkly.config.ts` file nor the monitor defined
- If the user wants to create a basic checkly config or wants to add the test they just ran as a monitor they can do that by passing the `--create-check` flag, if this flag is passed no test/check will run, it will just add the monitor to the config.
```bash
npx checkly pw-test -l us-east-1 --create-check -- --project foo --grep @bar
```
This will create something that will look like:
```typescript
import { defineConfig } from 'checkly'

const config = defineConfig({
  logicalId: 'logical-id',
  projectName: 'project-name',
  checks: {
    playwrightConfigPath: './playwright.config.ts',
    playwrightChecks: [
      {
        logicalId: 'playwright-check---project-foo---grep--bar',
        name: 'Playwright Test: --project foo --grep @bar',
        testCommand: 'npx playwright test --project foo --grep @bar',
        locations: [
          'us-east-1',
        ],
        frequency: 10,
      },
    ],
    frequency: 10,
    locations: [
      'us-east-1',
    ],
  },
  cli: {
    runLocation: 'us-east-1',
  },
})

export default config
```

### DIFF
--- a/packages/cli/src/commands/pw-test.ts
+++ b/packages/cli/src/commands/pw-test.ts
@@ -12,7 +12,7 @@ import * as api from '../rest/api'
 import config from '../services/config'
 import { parseProject } from '../services/project-parser'
 import type { Runtime } from '../rest/runtimes'
-import { Diagnostics, Session } from '../constructs'
+import { Diagnostics, RuntimeCheck, Session } from '../constructs'
 import { Flags, ux } from '@oclif/core'
 import { createReporters, ReporterType } from '../reporters/reporter'
 import TestRunner from '../services/test-runner'
@@ -150,15 +150,17 @@ export default class PwTestCommand extends AuthCommand {
       include: checklyConfig.checks?.include,
       playwrightChecks: [playwrightCheck],
       checkFilter: check => {
-        if (Object.keys(testEnvVars).length) {
-          check.environmentVariables = check.environmentVariables
-            ?.filter((envVar: any) => !testEnvVars[envVar.key]) || []
-          for (const [key, value] of Object.entries(testEnvVars)) {
-            check.environmentVariables.push({
-              key,
-              value,
-              locked: true,
-            })
+        if (check instanceof RuntimeCheck) {
+          if (Object.keys(testEnvVars).length) {
+            check.environmentVariables = check.environmentVariables
+              ?.filter((envVar: any) => !testEnvVars[envVar.key]) || []
+            for (const [key, value] of Object.entries(testEnvVars)) {
+              check.environmentVariables.push({
+                key,
+                value,
+                locked: true,
+              })
+            }
           }
         }
         return true


packages/cli/src/helpers/test-helper.ts
```diff
@@ -0,0 +1,71 @@
+import { PrivateRunLocation, RunLocation } from '../services/abstract-check-runner'
+import { Session } from '../constructs'
+import type { Region } from '..'
+import { ReporterType } from '../reporters/reporter'
+import { isCI } from 'ci-info'
+
+const DEFAULT_REGION = 'eu-central-1'
```
sorccu, line 7: Why not export this and use it everywhere instead of having a hardcoded `DEFAULT_REGION` in all of these files:

```
packages/cli/src/commands/trigger.ts
packages/cli/src/commands/test.ts
packages/cli/src/commands/pw-test.ts
```

packages/cli/src/commands/pw-test.ts
```diff
@@ -0,0 +1,380 @@
...
+        throw err
+      }
+    })()
+
+    const checkBundles = Object.values(projectBundle.data.check)
+
+    if (this.fancy) {
+      ux.action.stop()
+    }
+
```
sorccu, line 210: Can you remove this? I know the `test` command also has it (by mistake it looks like) and you've just copied it, but there is no corresponding start for this stop action, so it should be removed. Also, it will potentially mean removing the `ux` import.

```suggestion
```

sorccu (changes requested):
I've left some comments about things that should be addressed.

sorccu:
Also, the Windows tests are flaky because the GH Windows runner is ultra slow - rerun after implementing the changes above if it still doesn't succeed.

packages/cli/src/commands/pw-test.ts
```diff
@@ -0,0 +1,365 @@
...
+    this.style.actionSuccess()
+    return
+
+  }
+
+  private static async getTestCommand(directoryPath: string, input: string): Promise<string| undefined> {
+    const packageManager = await detectPackageManager(directoryPath)
+    if (!packageManager) {
+      throw new Error('Unable to detect package manager. Please ensure you are in a valid Node.js project directory.')
+    }
```
sorccu, line 361: You can remove this check - `detectPackageManager()` always returns npm as the final fallback, so it will never return an undefined value.

packages/cli/src/commands/pw-test.ts
```diff
@@ -0,0 +1,365 @@
...
+
+    static async createPlaywrightCheck(args: string[], runLocation: keyof Region, dir: string): Promise<PlaywrightSlimmedProp> {
+      const parseArgs = args.map(arg => {
+        if (arg.includes(' ')) {
+          arg = `"${arg}"`
+        }
+        return arg
+      })
+      const input = parseArgs.join(' ') || ''
+      const inputLogicalId = input.replace(/[^a-zA-Z0-9]/g, '-').toLowerCase().substring(0, 50)
```
sorccu, line 295: You can use `cased(input, 'kebab-case')` from `packages/cli/src/sourcegen/case.ts` instead.

packages/cli/src/commands/pw-test.ts
```diff
@@ -0,0 +1,365 @@
...
+    }
+
+    static async createPlaywrightCheck(args: string[], runLocation: keyof Region, dir: string): Promise<PlaywrightSlimmedProp> {
+      const parseArgs = args.map(arg => {
+        if (arg.includes(' ')) {
+          arg = `"${arg}"`
+        }
+        return arg
+      })
+      const input = parseArgs.join(' ') || ''
```
sorccu, line 294: It would be nice to not do this here, because the `unsafeDisplayCommand` (used in getTestCommand) quotes args for you. Instead, the input args should just be passed to `execCommand()` as array items. However, that has some implications for the test name below which also uses the manually quoted `input` string.

packages/cli/src/commands/pw-test.ts
sorccu: Or `kebabCase(input)` from the same place if you prefer. It's essentially the same thing.

packages/cli/src/commands/pw-test.ts
```diff
@@ -0,0 +1,363 @@
...
+  splitConfigFilePath, writeChecklyConfigFile
+} from '../services/util'
+import { getChecklyConfigFile, loadChecklyConfig, PlaywrightSlimmedProp } from '../services/checkly-config-loader'
+import { prepareReportersTypes, prepareRunLocation, splitChecklyAndPlaywrightFlags } from '../helpers/test-helper'
+import * as api from '../rest/api'
+import config from '../services/config'
+import { parseProject } from '../services/project-parser'
+import type { Runtime } from '../rest/runtimes'
+import { Diagnostics, Session } from '../constructs'
+import { Flags, ux } from '@oclif/core'
```
sorccu, line 16: You can remove the unused ux import

sorccu (dismissed):
Added a minor comment, but approving for now.

PR #1093, opened 2025-06-20
feat: add support for UrlMonitors [sc-24607]

sorccu:
This PR introduces support for UrlMonitors, which are essentially lightweight API checks.

This change separates monitors from runtime checks, changing the existing `TcpCheck` and `HeartbeatCheck` to `TcpMonitor` and `HeartbeatMonitor`, respectively. Other check types require a runtime and are therefore runtime checks. Aliases are provided for backwards compatibility (`TcpCheck` and `HeartbeatCheck` are still usable).

Also included is a minor fix for an unrelated issue: If you have a check that belongs to a group, and then want to ungroup it by removing the `group:` property, that doesn't work because it results in `group: undefined` which is not included in the payload sent to the backend, and the backend then keeps the current value.

## Affected Components
* [x] CLI
* [ ] Create CLI
* [x] Test
* [ ] Docs
* [x] Examples
* [ ] Other

## Notes for the Reviewer

> Resolves #[issue-number]

## New Dependency Submission

# Test Evidence `(against prod)`
Added by @ejanusevicius 

`test`, `test --record`, `deploy`, `destroy`:

https://github.com/user-attachments/assets/77c7103e-e80b-4e11-979b-2c4067c34ec6

`trigger`:

https://github.com/user-attachments/assets/f7c8d510-f19a-4fca-8912-782a5ea9b7fd


packages/cli/src/constructs/http-check.ts
```diff
@@ -0,0 +1,58 @@
+import { Check, CheckProps } from './check'
+import { HttpRequest } from './http-request'
+import { Session } from './project'
+
+export interface HttpCheckProps extends Omit<CheckProps, 'doubleCheck' | 'shouldFail'> {
```
sbezludny, line 5: everything related to scripting needs to be omitted as well: tearDownScript and so on

sbezludny (changes requested):

packages/cli/src/constructs/http-assertion.ts
```diff
@@ -0,0 +1,49 @@
...
+    return new NumericAssertionBuilder<HttpAssertionSource>('RESPONSE_TIME')
+  }
+}
+
+// This class serves as an alias for backwards compatibility while also
+// removing the deprecated method from HttpAssertionBuilder by isolating
+// it to the AssertionBuilder.
+export class AssertionBuilder extends HttpAssertionBuilder {
+  /** @deprecated Use responseTime() instead */
+  static responseTme () {
```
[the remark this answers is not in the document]
sorccu: It's deprecated because of the typo. For now we keep it as is until it can be removed entirely.

packages/cli/src/constructs/http-check.ts
sorccu: How about environment variables? I guess they will still be usable in headers and such?

PR #1098, opened 2025-07-04
JSDoc enhancements

hlenke:
I created some basic additional JSDocs based on our official documentation. Please review. 

## Affected Components
* [X] CLI
* [ ] Create CLI
* [ ] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

### DIFF
--- a/packages/cli/src/constructs/alert-escalation-policy.ts
+++ b/packages/cli/src/constructs/alert-escalation-policy.ts
@@ -1,33 +1,80 @@
+/** Types of alert escalation strategies */
 // eslint-disable-next-line no-restricted-syntax
 enum AlertEscalationType {
   RUN = 'RUN_BASED',
   TIME = 'TIME_BASED'
 }
 
+/**
+ * Configuration for alert reminders.
+ * Defines how often to send reminder notifications after initial alert.
+ */
 export type Reminders = {
+  /** Number of reminder notifications to send (0 to disable) */
   amount?: number,
+  /** Interval between reminder notifications in minutes */
   interval?: number
 }
 
+/**
+ * Configuration for parallel run failure threshold.
+ * Determines when to alert based on percentage of failed parallel runs.
+ */
 export type ParallelRunFailureThreshold = {
+  /** Whether parallel run failure threshold is enabled */
   enabled?: boolean,
+  /** Percentage of runs that must fail to trigger alert (1-100) */
   percentage?: number,
 }
 
+/**
+ * Configuration for alert escalation policies.
+ * Defines when and how to escalate alerts based on check failures.
+ */
 export interface AlertEscalation {
+  /** The type of escalation strategy to use */
   escalationType?: AlertEscalationType,
+  /** Configuration for run-based escalation */
   runBasedEscalation?: {
+    /** Number of consecutive failed runs before escalating */
     failedRunThreshold?: number
   },
+  /** Configuration for time-based escalation */
   timeBasedEscalation?: {
+    /** Minutes that check must be failing before escalating */
     minutesFailingThreshold?: number
   },
+  /** Configuration for reminder notifications */
   reminders?: Reminders
+  /** Configuration for parallel run failure threshold */
   parallelRunFailureThreshold?: ParallelRunFailureThreshold
 }
 
+/**
+ * Options for configuring alert escalation behavior.
+ * These options can be used with any escalation strategy.
+ */
 export type AlertEscalationOptions = Pick<AlertEscalation, 'runBasedEscalation' | 'timeBasedEscalation' | 'reminders' | 'parallelRunFailureThreshold'>
 
+/**
+ * Builder class for creating alert escalation policies.
+ * Provides convenient methods to create different types of escalation strategies.
+ * 
+ * @example
+ * ```typescript
+ * // Run-based escalation - alert after 3 consecutive failures
+ * const runBased = AlertEscalationBuilder.runBasedEscalation(3, {
+ *   amount: 2,
+ *   interval: 10
+ * })
+ * 
+ * // Time-based escalation - alert after 10 minutes of failure
+ * const timeBased = AlertEscalationBuilder.timeBasedEscalation(10, {
+ *   amount: 1,
+ *   interval: 30
+ * })
+ * ```
+ */
 export class AlertEscalationBuilder {
   private static DEFAULT_RUN_BASED_ESCALATION = { failedRunThreshold: 1 }
   private static DEFAULT_TIME_BASED_ESCALATION = { minutesFailingThreshold: 5 }
--- a/packages/cli/src/constructs/api-check.ts
+++ b/packages/cli/src/constructs/api-check.ts
@@ -10,6 +10,7 @@ import { Diagnostics } from './diagnostics'
 import { DeprecatedPropertyDiagnostic, InvalidPropertyValueDiagnostic } from './construct-diagnostics'
 import { ApiCheckBundle, ApiCheckBundleProps } from './api-check-bundle'
 
+/** Sources that can be used for API check assertions */
 type AssertionSource =
   | 'STATUS_CODE'
   | 'JSON_BODY'
@@ -17,21 +18,76 @@ type AssertionSource =
   | 'TEXT_BODY'
   | 'RESPONSE_TIME'
 
+/**
+ * Assertion configuration for API checks.
+ * Allows validation of different aspects of HTTP responses.
+ */
 export type Assertion = CoreAssertion<AssertionSource>
 
+/**
+ * Builder class for creating API check assertions.
+ * Provides convenient methods to create assertions for different parts of HTTP responses.
+ * 
+ * @example
+ * ```typescript
+ * // Status code assertions
+ * AssertionBuilder.statusCode().equals(200)
+ * AssertionBuilder.statusCode().isIn([200, 201, 202])
+ * 
+ * // JSON body assertions using JSONPath
+ * AssertionBuilder.jsonBody('$.user.name').equals('John')
+ * AssertionBuilder.jsonBody('$.users').hasLength(5)
+ * AssertionBuilder.jsonBody('$.data[0].id').isNotNull()
+ * 
+ * // Header assertions
+ * AssertionBuilder.headers('content-type').contains('application/json')
+ * AssertionBuilder.headers('x-rate-limit-remaining').greaterThan(0)
+ * 
+ * // Text body assertions
+ * AssertionBuilder.textBody().contains('Welcome to our API')
+ * AssertionBuilder.textBody().matches(/success|ok/i)
+ * 
+ * // Response time assertions
+ * AssertionBuilder.responseTime().lessThan(1000)
+ * AssertionBuilder.responseTime().between(100, 2000)
+ * ```
+ * 
+ * @see {@link https://jsonpath.com/ | JSONPath Online Evaluator}
+ * @see {@link https://www.checklyhq.com/docs/monitoring/api-checks/assertions/ | API Check Assertions}
+ */
 export class AssertionBuilder {
+  /**
+   * Creates an assertion builder for HTTP status codes.
+   * @returns A numeric assertion builder for status codes
+   */
   static statusCode () {
     return new NumericAssertionBuilder<AssertionSource>('STATUS_CODE')
   }
 
+  /**
+   * Creates an assertion builder for JSON response body.
+   * @param property Optional JSON path to specific property (e.g., 'user.name')
+   * @returns A general assertion builder for JSON body content
+   */
   static jsonBody (property?: string) {
     return new GeneralAssertionBuilder<AssertionSource>('JSON_BODY', property)
   }
 
+  /**
+   * Creates an assertion builder for HTTP headers.
+   * @param property Optional header name to target specific header
+   * @param regex Optional regex pattern for header matching
+   * @returns A general assertion builder for headers
+   */
   static headers (property?: string, regex?: string) {
     return new GeneralAssertionBuilder<AssertionSource>('HEADERS', property, regex)
   }
 
+  /**
+   * Creates an assertion builder for text response body.
+   * @param property Optional property path for text content
+   * @returns A general assertion builder for text body content
+   */
   static textBody (property?: string) {
     return new GeneralAssertionBuilder<AssertionSource>('TEXT_BODY', property)
   }
@@ -41,13 +97,19 @@ export class AssertionBuilder {
     return new NumericAssertionBuilder<AssertionSource>('RESPONSE_TIME')
   }
 
+  /**
+   * Creates an assertion builder for response time.
+   * @returns A numeric assertion builder for response time in milliseconds
+   */
   static responseTime () {
     return new NumericAssertionBuilder<AssertionSource>('RESPONSE_TIME')
   }
 }
 
+/** HTTP request body types supported by API checks */
 export type BodyType = 'JSON' | 'FORM' | 'RAW' | 'GRAPHQL' | 'NONE'
 
+/** HTTP request methods supported by API checks */
 export type HttpRequestMethod =
   | 'get' | 'GET'
   | 'post' | 'POST'
@@ -57,35 +119,66 @@ export type HttpRequestMethod =
   | 'delete' | 'DELETE'
   | 'options' | 'OPTIONS'
 
+/** IP family versions for network requests */
 export type IPFamily = 'IPv4' | 'IPv6'
+
+/**
+ * Basic authentication credentials for HTTP requests.
+ */
 export interface BasicAuth {
+  /** Username for basic authentication */
   username: string
+  /** Password for basic authentication */
   password: string
 }
 
+/**
+ * Default configuration that can be applied to API checks.
+ * Used for setting common defaults across multiple checks.
+ */
 export type ApiCheckDefaultConfig = {
+  /** Default URL for API requests */
   url?: string,
+  /** Default HTTP headers to include */
   headers?: Array<HttpHeader>
+  /** Default query parameters to include */
   queryParameters?: Array<QueryParam>
+  /** Default basic authentication credentials */
   basicAuth?: BasicAuth
+  /** Default assertions to apply */
   assertions?: Array<Assertion>
 }
 
+/**
+ * Configuration for an HTTP request in an API check.
+ * Defines all aspects of the HTTP request to be made.
+ */
 export interface Request {
+  /** The URL to make the request to */
   url: string,
+  /** The HTTP method to use */
   method: HttpRequestMethod,
+  /** IP family version to use (IPv4 or IPv6) */
   ipFamily?: IPFamily,
+  /** Whether to follow HTTP redirects automatically */
   followRedirects?: boolean,
+  /** Whether to skip SSL certificate verification */
   skipSSL?: boolean,
   /**
+   * Assertions to validate the HTTP response.
    * Check the main Checkly documentation on assertions for specific values like regular expressions
    * and JSON path descriptors you can use in the "property" field.
    */
   assertions?: Array<Assertion>
+  /** The request body content */
   body?: string
+  /** The type of the request body */
   bodyType?: BodyType
+  /** HTTP headers to include in the request */
   headers?: Array<HttpHeader>
+  /** Query parameters to include in the request URL */
   queryParameters?: Array<QueryParam>
+  /** Basic authentication credentials */
   basicAuth?: BasicAuth
 }
 
@@ -123,11 +216,45 @@ export interface ApiCheckProps extends CheckProps {
 }
 
 /**
- * Creates an API Check
+ * Creates an API Check to monitor HTTP endpoints and APIs.
+ * 
+ * API checks allow you to monitor REST APIs, GraphQL endpoints, and any HTTP-based service.
+ * You can validate response status codes, response times, headers, and response body content.
  *
- * @remarks
- *
- * This class make use of the API Checks endpoints.
+ * @example
+ * ```typescript
+ * // Basic API check
+ * new ApiCheck('hello-api', {
+ *   name: 'Hello API',
+ *   request: {
+ *     method: 'GET',
+ *     url: 'https://api.example.com/hello'
+ *   }
+ * })
+ * 
+ * // Advanced API check with assertions
+ * new ApiCheck('user-api', {
+ *   name: 'User API Check',
+ *   frequency: Frequency.EVERY_5M,
+ *   locations: ['us-east-1', 'eu-west-1'],
+ *   request: {
+ *     method: 'POST',
+ *     url: 'https://api.example.com/users',
+ *     headers: [{ key: 'Content-Type', value: 'application/json' }],
+ *     body: JSON.stringify({ name: 'test-user' }),
+ *     assertions: [
+ *       AssertionBuilder.statusCode().equals(201),
+ *       AssertionBuilder.jsonBody('$.id').isNotNull(),
+ *       AssertionBuilder.responseTime().lessThan(1000)
+ *     ]
+ *   },
+ *   maxResponseTime: 5000,
+ *   degradedResponseTime: 2000
+ * })
+ * ```
+ * 
+ * @see {@link https://www.checklyhq.com/docs/cli/constructs-reference/#apicheck | ApiCheck API Reference}
+ * @see {@link https://www.checklyhq.com/docs/monitoring/api-checks/ | API Checks Documentation}
  */
 export class ApiCheck extends Check {
   readonly request: Request
--- a/packages/cli/src/constructs/browser-check.ts
+++ b/packages/cli/src/constructs/browser-check.ts
@@ -31,11 +31,50 @@ export interface BrowserCheckProps extends CheckProps {
 }
 
 /**
- * Creates a Browser Check
+ * Creates a Browser Check to monitor web applications using Playwright.
+ * 
+ * Browser checks allow you to monitor complex user interactions, page performance,
+ * and visual regressions. They run real browser scripts using Playwright to simulate
+ * user behavior and validate web application functionality.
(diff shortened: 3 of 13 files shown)


packages/cli/src/constructs/api-check.ts
```diff
@@ -10,28 +10,84 @@ import { Diagnostics } from './diagnostics'
...
+ * 
+ * @example
+ * ```typescript
+ * // Status code assertions
+ * AssertionBuilder.statusCode().equals(200)
+ * AssertionBuilder.statusCode().isIn([200, 201, 202])
+ * 
+ * // JSON body assertions using JSONPath
+ * AssertionBuilder.jsonBody('$.user.name').equals('John')
+ * AssertionBuilder.jsonBody('$.users').hasLength(5)
```
sorccu, line 39: We don't have a `hasLength()` assertion.

packages/cli/src/constructs/api-check.ts
```diff
@@ -10,28 +10,84 @@ import { Diagnostics } from './diagnostics'
...
+ * AssertionBuilder.headers('content-type').contains('application/json')
+ * AssertionBuilder.headers('x-rate-limit-remaining').greaterThan(0)
+ * 
+ * // Text body assertions
+ * AssertionBuilder.textBody().contains('Welcome to our API')
+ * AssertionBuilder.textBody().matches(/success|ok/i)
+ * 
+ * // Response time assertions
+ * AssertionBuilder.responseTime().lessThan(1000)
+ * AssertionBuilder.responseTime().between(100, 2000)
```
sorccu, line 52: We don't have a `between()` assertion.

packages/cli/src/constructs/api-check.ts
```diff
@@ -10,28 +10,84 @@ import { Diagnostics } from './diagnostics'
...
 
+/**
+ * Builder class for creating API check assertions.
+ * Provides convenient methods to create assertions for different parts of HTTP responses.
+ * 
+ * @example
+ * ```typescript
+ * // Status code assertions
+ * AssertionBuilder.statusCode().equals(200)
+ * AssertionBuilder.statusCode().isIn([200, 201, 202])
```
sorccu, line 35: We don't have an `isIn()` assertion.

packages/cli/src/constructs/check.ts
```diff
@@ -14,95 +14,226 @@ import { RetryStrategy } from './retry-strategy'
...
+   *   assertions: [AssertionBuilder.statusCode().equals(404)]
+   * }
+   * ```
    */
   shouldFail?: boolean
+  
   /**
    * The runtime version, i.e. fixed set of runtime dependencies, used to execute this check.
+   * 
+   * @example "2023.09" | "2022.10"
```
sorccu, line 88: Maybe we should use a newer runtime as an example?

packages/cli/src/constructs/construct.ts
```diff
@@ -4,26 +4,75 @@ import { Diagnostics } from './diagnostics'
...
+ * class MyConstruct extends Construct {
+ *   constructor(logicalId: string) {
+ *     super('my-construct', logicalId);
+ *   }
+ *   
+ *   synthesize() {
+ *     return { name: 'my-construct' };
+ *   }
+ * }
+ * ```
```
sorccu, line 54: I don't think we want to provide an example for this. While technically a customer may be able to create their own constructs, they should instead extend one of the higher level types such as `ApiCheck`. If we openly support creating custom constructs there's no telling what will happen over time.

```suggestion
```

sorccu (changes requested):
Left some comments

packages/cli/src/constructs/alert-channel.ts
```diff
@@ -51,11 +83,38 @@ export class AlertChannelRef extends Construct {
...
 /**
- * Creates an Alert Channels
+ * Base class for creating alert channels that notify when checks fail or recover.
+ * 
+ * Alert channels define how and when to send notifications when monitoring checks
+ * fail, recover, or enter a degraded state. Checkly supports multiple alert channel types
+ * including email, SMS, Slack, webhooks, and third-party integrations.
  *
- * @remarks
+ * **Plan Limitations:**
```
stefanjudis, line 92: @hlenke I wouldn't have this here. Pricing in TSDocs is too much and most likely won't be updated when pricing changes.

packages/cli/src/constructs/api-assertion.ts
```diff
@@ -4,35 +4,96 @@ import {
...
+ * // Text body assertions
+ * AssertionBuilder.textBody().contains('Welcome to our API')
+ * AssertionBuilder.textBody().notContains('error')
+ * 
+ * // Response time assertions
+ * AssertionBuilder.responseTime().lessThan(1000)
+ * AssertionBuilder.responseTime().greaterThan(100)
+ * ```
+ * 
+ * @see {@link https://jsonpath.com/ | JSONPath Online Evaluator}
```
stefanjudis, line 51: Let's scratch this. It might be unnecessary.

packages/cli/src/constructs/api-assertion.ts
```diff
@@ -4,35 +4,96 @@ import {
...
+ * @see {@link https://www.checklyhq.com/docs/monitoring/api-checks/assertions/ | API Check Assertions}
```
stefanjudis, line 52:
```suggestion
 * @see {@link https://www.checklyhq.com/docs/api-checks/assertions/ | API Check Assertions}
```

packages/cli/src/constructs/browser-check.ts
```diff
@@ -31,11 +31,50 @@ export interface BrowserCheckProps extends RuntimeCheckProps {
...
+ *   playwrightConfig: {
+ *     use: {
+ *       viewport: { width: 1280, height: 720 }
+ *     }
+ *   }
+ * })
+ * ```
+ * 
+ * @see {@link https://www.checklyhq.com/docs/cli/constructs-reference/#browsercheck | BrowserCheck API Reference}
+ * @see {@link https://www.checklyhq.com/docs/monitoring/browser-checks/ | Browser Checks Documentation}
```
stefanjudis, line 76:
```suggestion
 * @see {@link https://www.checklyhq.com/docs/browser-checks/ | Browser Checks Documentation}
```

packages/cli/src/constructs/check.ts
```diff
@@ -16,90 +16,215 @@ import { ConfigDefaultsGetter, makeConfigDefaultsGetter } from './check-config'
 import { Diagnostics } from './diagnostics'
 import { validateDeprecatedDoubleCheck } from './internal/common-diagnostics'
 
+/**
```
stefanjudis, line 19: @sorccu Hmmm... I'm unsure of including these docs because people would never interact with `CheckProps` or am I wrong?

packages/cli/src/constructs/check.ts
```diff
@@ -16,90 +16,215 @@ import { ConfigDefaultsGetter, makeConfigDefaultsGetter } from './check-config'
...
   tags?: Array<string>
+  
   /**
-   * How often the check should run in minutes.
+   * How often the check should run. Can be specified in minutes or using Frequency constants.
+   * 
+   * @example
+   * ```typescript
+   * frequency: Frequency.EVERY_5M  // Using frequency constant
+   * frequency: 10                   // Every 10 minutes
```
stefanjudis, line 132: Let's remove this line and enfore `Frequency`?

packages/cli/src/constructs/check.ts
```diff
@@ -16,90 +16,215 @@ import { ConfigDefaultsGetter, makeConfigDefaultsGetter } from './check-config'
...
+   * ```
    */
   frequency?: number | Frequency
```
stefanjudis, line 135: Is this correct? Do we allow custom frequencies?

packages/cli/src/constructs/check.ts
```diff
@@ -197,6 +322,11 @@ export abstract class Check extends Construct {
     return config
   }
 
+  /**
+   * Creates alert channel subscriptions for this check.
+   * Links the check to its configured alert channels so notifications are sent.
```
stefanjudis, line 327:
```suggestion
   * Links the check to its configured alert channels to send notifications.
```

packages/cli/src/constructs/frequency.ts
```diff
@@ -1,24 +1,81 @@
...
+ * frequency: Frequency.EVERY_5M
+ * 
+ * // Daily health checks (every 24 hours)
+ * frequency: Frequency.EVERY_24H
+ * 
+ * // Custom frequency (every 45 minutes) - must be supported value
+ * frequency: 45  // Not supported - use closest: 30 or 60
+ * ```
+ * 
+ * @see {@link https://www.checklyhq.com/docs/monitoring/check-frequency/ | Check Frequency Documentation}
```
stefanjudis, line 31: This link 404s. Not sure where to link instead, though.

packages/cli/src/constructs/frequency.ts
```diff
@@ -1,24 +1,81 @@
...
+ * 
+ * // Standard API monitoring (every 5 minutes)
+ * frequency: Frequency.EVERY_5M
+ * 
+ * // Daily health checks (every 24 hours)
+ * frequency: Frequency.EVERY_24H
+ * 
+ * // Custom frequency (every 45 minutes) - must be supported value
+ * frequency: 45  // Not supported - use closest: 30 or 60
+ * ```
```
stefanjudis, line 29: This seems "a bit much"... Should we shorten it?

stefanjudis (dismissed):
I've left some niggles and double-checked all linked URLs. I can't give a statement in terms of correctness of default values and stuff, though.

But other than that — LGTM and it's def better than what we had before.

packages/cli/src/constructs/browser-check.ts
sorccu: You suggestion looks correct to me.

packages/cli/src/constructs/alert-escalation-policy.ts
```diff
@@ -1,33 +1,113 @@
...
+/**
+ * Configuration for alert reminders.
+ * Defines how often to send reminder notifications after initial alert.
+ */
 export type Reminders = {
+  /** 
+   * Number of reminder notifications to send (0 to disable).
+   * @defaultValue 0
+   * @minimum 0
+   * @maximum 5
```
sorccu, line 17: It looks like `100000` is also a supported value, probably to indicate practically unlimited reminders.

This could therefore perhaps be changed to:
```suggestion
   * @enum [0, 1, 2, 3, 4, 5, 100000]
```

Although ideally we'd put the "unlimited" value behind a constant. Do we want to expose this unlimited value in the docs?

packages/cli/src/constructs/alert-escalation-policy.ts
```diff
@@ -1,33 +1,113 @@
...
+ * Defines when and how to escalate alerts based on check failures.
+ */
 export interface AlertEscalation {
+  /** The type of escalation strategy to use */
   escalationType?: AlertEscalationType,
+  /** Configuration for run-based escalation */
   runBasedEscalation?: {
+    /** 
+     * Number of consecutive failed runs before escalating.
+     * @defaultValue 1
```
sorccu, line 63: I don't actually see a default value for this at the backend level, or at least at the schema level, when `runBasedEscalation` is not specified, or set to an empty object, or `alertSettings` is left out entirely. It is true however that if the user uses `AlertEscalationBuilder`, then the default value is indeed `1`. But if the user manually creates the `alertEscalationPolicy`, I'm not sure what the effective default will end up being. This is probably a bug that should be fixed in the backend tbh.

packages/cli/src/constructs/alert-escalation-policy.ts
```diff
@@ -1,33 +1,113 @@
...
+     * @maximum 5
+     * @example 3  // Escalate after 3 consecutive failures
+     */
     failedRunThreshold?: number
   },
+  /** Configuration for time-based escalation */
   timeBasedEscalation?: {
+    /** 
+     * Minutes that check must be failing before escalating.
+     * @defaultValue 5
```
sorccu, line 74: Similar to above, the backend doesn't seem to enforce a default value. If `AlertEscalationBuilder` is used, then the default is `5`.

packages/cli/src/constructs/api-assertion.ts
sorccu: Your suggestion looks good to me.

packages/cli/src/constructs/check.ts
sorccu: No, we can't introduce breaking changes in a docs PR.

packages/cli/src/constructs/check.ts
sorccu: Not usually no, but I think the example is valid - if you wanted to have maintain a common set of properties for your checks, this would be the way to do it in a type safe way. You'd then use it like this:

```
new ApiCheck('foo', {
  ...commonCheckProps,
  name: 'Foo',
})
```

The example could be expanded to elaborate on this use case though.

packages/cli/src/constructs/check.ts
```diff
@@ -16,90 +16,215 @@ import { ConfigDefaultsGetter, makeConfigDefaultsGetter } from './check-config'
...
   shouldFail?: boolean
+  
   /**
-   * An array of one or more data center locations where to run this check. The supported regions are:
+   * An array of one or more data center locations where to run this check.
+   * The supported regions are:
    * us-east-1, us-east-2, us-west-1, us-west-2, ca-central-1, sa-east-1,
    * eu-west-1, eu-central-1, eu-west-2, eu-west-3, eu-north-1, eu-south-1, me-south-1,
    * ap-southeast-1, ap-northeast-1, ap-east-1, ap-southeast-2, ap-southeast-3, ap-northeast-2, ap-northeast-3,
    * ap-south-1, af-south-1
```
sorccu, line 94: Not sure if it's worth specifying all of these here tbh. I doubt we'll remember to keep it up to date, should it change.

packages/cli/src/constructs/check.ts
sorccu: Mentioned elsewhere but no breaking changes in a doc PR.

packages/cli/src/constructs/check.ts
```diff
@@ -16,90 +16,215 @@ import { ConfigDefaultsGetter, makeConfigDefaultsGetter } from './check-config'
...
    * The CheckGroup that this check is part of.
+   * Groups allow you to organize related checks and apply common settings.
+   * 
+   * @example
+   * ```typescript
+   * // Create a new check group
+   * const apiGroup = new CheckGroupV2('api-group', {
+   *   name: 'API Checks',
+   *   activated: true,
+   *   locations: ['us-east-1']
```
sorccu, line 152: Well, this is correct, but given how we went through all the trouble to introduce "folder" groups, maybe our example should remove the `activated` and `locations` properties that can potentially influence checks in the group?

Or not, up to you.

packages/cli/src/constructs/check.ts
```diff
@@ -16,90 +16,215 @@ import { ConfigDefaultsGetter, makeConfigDefaultsGetter } from './check-config'
...
    * If you don't set at least one, we won't be able to alert you in case something goes wrong with your check.
-   * @link {https://www.checklyhq.com/docs/alerting-and-retries/alert-channels/#alert-channels Alert channels}
+   * 
+   * @example
+   * ```typescript
+   * alertChannels: [
+   *   new EmailAlertChannel('email-alerts', { address: 'team@example.com' }),
+   *   new SlackAlertChannel('slack-alerts', { url: 'https://hooks.slack.com/...' })
+   * ]
+   * ```
```
sorccu, line 174: While this is a valid way to use alert channels, I think the example is a bit dangerous... should a user copy this as is, they'd probably end up creating a whole bunch of duplicate alert channels if they try to apply the same to multiple checks. And/or, get confused why they're getting `logicalId` conflicts. Instead, the example should emphasize storing the AlertChannel instances in variables, and then using the alert channels via those variable references.

packages/cli/src/constructs/check.ts
```diff
@@ -197,6 +322,11 @@ export abstract class Check extends Construct {
     return config
   }
 
+  /**
+   * Creates alert channel subscriptions for this check.
+   * Links the check to its configured alert channels so notifications are sent.
+   * Only creates subscriptions if alert channels are configured and check is not test-only.
+   */
   addSubscriptions () {
```
sorccu, line 330: This should be a `protected` method tbh - it's not mean to be called by users

packages/cli/src/constructs/retry-strategy.ts
```diff
@@ -1,27 +1,101 @@
...
+   */
   baseBackoffSeconds?: number,
+  
   /**
-  * The maximum number of attempts to retry the check. Value must be between 1 and 10.
-  */
+   * The maximum number of attempts to retry the check.
+   * Value must be between 1 and 10.
+   * 
+   * @defaultValue 2
```
sorccu, line 38: The default value seems to be coming from the `RetryStrategyBuilder`. The backend has no default for this. A bit unclear how to best present the default value.

Regardless, the value has min/max, added those.

```suggestion
   * @defaultValue 2
   * @min 1
   * @max 10
```

packages/cli/src/constructs/retry-strategy.ts
```diff
@@ -1,27 +1,101 @@
...
+   */
   maxDurationSeconds?: number,
+  
   /**
-  * Whether retries should be run in the same region as the initial check run.
-  */
+   * Whether retries should be run in the same region as the initial check run.
+   * If false, retries use a randomly selected region for better fault tolerance.
+   * 
+   * @defaultValue true
```
sorccu, line 55: This is a bit tricky. The default is true when using the builder. However, the backend default is `false`, though the value is also required, so there's effectively no default value. Perhaps true is acceptable.

packages/cli/src/constructs/retry-strategy.ts
```diff
@@ -1,27 +1,101 @@
...
   /**
-  * The number of seconds to wait before the first retry attempt.
-  */
+   * The number of seconds to wait before the first retry attempt.
+   * This value is used differently based on the retry strategy type:
+   * - FIXED: Same delay for all retries
+   * - LINEAR: Base value that increases linearly (baseBackoffSeconds * attempt)
+   * - EXPONENTIAL: Base value that increases exponentially (baseBackoffSeconds ^ attempt)
+   * 
+   * @defaultValue 60
```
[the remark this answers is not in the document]
sorccu: Also, `@min` is 0, there's no `@max`.

packages/cli/src/constructs/tcp-monitor.ts
```diff
@@ -52,10 +69,29 @@ export interface TcpMonitorProps extends MonitorProps {
   request: TcpRequest
   /**
    * The response time in milliseconds where a check should be considered degraded.
+   * TCP checks have lower thresholds than HTTP checks due to protocol differences.
+   * 
+   * @defaultValue 4000
+   * @minimum 0
+   * @maximum 5000
```
sorccu, line 76: These are correct for the public API but it looks like the CLI endpoint does not enforce the same values. These are the effective values currently:

```
   * @defaultValue 10000
   * @minimum 0
   * @maximum 300000
```

This should probably be fixed in the backend.

packages/cli/src/constructs/tcp-monitor.ts
```diff
@@ -52,10 +69,29 @@ export interface TcpMonitorProps extends MonitorProps {
...
    */
   degradedResponseTime?: number
+  
   /**
    * The response time in milliseconds where a check should be considered failing.
+   * Maximum allowed value is lower for TCP checks compared to HTTP checks.
+   * 
+   * @defaultValue 5000
+   * @minimum 0
+   * @maximum 5000
```
sorccu, line 90: These are correct for the public API but it looks like the CLI endpoint does not enforce the same values. These are the effective values currently:

```
   * @defaultValue 20000
   * @minimum 0
   * @maximum 300000
```

This should probably be fixed in the backend.

packages/cli/src/constructs/tcp-monitor.ts
sorccu, line 1: No comments have been generated for `TcpAssertionBuilder`.

sorccu (changes requested):
Left some comments. Also, no comments have been generated for the new UrlMonitor and UrlAssertionBuilder.

packages/cli/src/constructs/tcp-monitor.ts
hlenke: @sorccu Ok, that means we keep it this way for this PR, right?

packages/cli/src/constructs/check.ts
sorccu: Let's keep this as is for now to avoid any potentially breaking changes in the PR, but might want to rethink in the next major release.

packages/cli/src/constructs/retry-strategy.ts
```diff
@@ -1,27 +1,101 @@
...
   maxRetries?: number,
+  
   /**
-  * The total amount of time to continue retrying the check (maximum 600 seconds).
-  */
+   * The total amount of time to continue retrying the check.
+   * Maximum value is 600 seconds (10 minutes).
+   * If retries would exceed this duration, they are stopped early.
+   * 
+   * @defaultValue 600
```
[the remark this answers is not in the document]
sorccu: I think it's reasonable to keep the default value, and we can try to fix it in the BE to apply even if the user does not use the builder. That's more consistent anyway.

packages/cli/src/constructs/retry-strategy.ts
sorccu: I think it's reasonable to keep the default value, and we can try to fix it in the BE to apply even if the user does not use the builder. That's more consistent anyway.

packages/cli/src/constructs/retry-strategy.ts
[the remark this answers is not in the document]
sorccu: I think it's reasonable to keep the default value, and we can try to fix it in the BE to apply even if the user does not use the builder. That's more consistent anyway.

packages/cli/src/constructs/retry-strategy.ts
sorccu: I think it's reasonable to keep the default value, and we can try to fix it in the BE to apply even if the user does not use the builder. That's more consistent anyway.

packages/cli/src/constructs/frequency.ts
sorccu: I think we could remove the `**Supported Frequencies:**` and `**Frequency Offset Validation:**` sections, and then the last `// Custom frequency` example which is kinda useless. Also, in the future we could add CLI-side validation for the number value, making the example even less useful.

packages/cli/src/constructs/alert-escalation-policy.ts
sorccu: I think it's reasonable to keep the default value, and we can try to fix it in the BE to apply even if the user does not use the builder. That's more consistent anyway.

packages/cli/src/constructs/api-assertion.ts
hlenke: I'll keep it since it seems like an important detail.

packages/cli/src/constructs/check.ts
hlenke: deleted.

packages/cli/src/constructs/check.ts
hlenke: I'll delete the JSDoc here.

packages/cli/src/constructs/retry-strategy.ts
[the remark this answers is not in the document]
hlenke: Will do. Thank you!

packages/cli/src/constructs/tcp-monitor.ts
hlenke: I added comments. We still need to re-review these.

packages/cli/src/constructs/check.ts
```diff
@@ -16,90 +16,200 @@ import { ConfigDefaultsGetter, makeConfigDefaultsGetter } from './check-config'
...
   tags?: Array<string>
+  
   /**
-   * How often the check should run in minutes.
+   * How often the check should run. Can be specified in minutes or using Frequency constants.
+   * 
+   * @example
+   * ```typescript
+   * frequency: Frequency.EVERY_5M  // Using frequency constant
+   * frequency: 10                   // Every 10 minutes
```
sorccu, line 113:
```suggestion
   * frequency: Frequency.EVERY_5M // Every 5 minutes
```

packages/cli/src/constructs/check.ts
sorccu: Updated

packages/cli/src/constructs/frequency.ts
```diff
@@ -1,24 +1,70 @@
...
+ * frequency: Frequency.EVERY_30S
+ * 
+ * // Standard API monitoring (every 5 minutes)
+ * frequency: Frequency.EVERY_5M
+ * 
+ * // Daily health checks (every 24 hours)
+ * frequency: Frequency.EVERY_24H
+ * 
+ * // Custom frequency (every 45 minutes) - must be supported value
+ * frequency: 45  // Not supported - use closest: 30 or 60
```
sorccu, line 17:
```suggestion
```

packages/cli/src/constructs/check.ts
sorccu: Removed

sorccu:
I think everything's good now, except that the new UrlMonitor/UrlRequest/UrlAssertionBuilder have not had documentation generated for them.

hlenke:
@sorccu I would prefer to merge this one and create another for the three constructs. Any objections?

sorccu:
Ok, approved

PR #1111, opened 2025-07-14
feat: add ability to stream logs

ferrandiaz:
## Affected Components
* [ ] CLI
* [ ] Create CLI
* [ ] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer

> Resolves #[issue-number]


packages/cli/src/commands/pw-test.ts
```diff
@@ -82,6 +82,10 @@ export default class PwTestCommand extends AuthCommand {
     'create-check': Flags.boolean({
       description: 'Create a Checkly check from the Playwright test.',
       default: false,
+    }),
+    'stream-logs': Flags.boolean({
+      description: 'Stream logs from the test run to the console.',
+      default: false,
```
umutuzgur, line 88: Anything against having this defaulted to true?

ferrandiaz: With our current setup I would say cost.

packages/cli/src/reporters/abstract-list.ts
```diff
@@ -96,6 +97,45 @@ export default abstract class AbstractListReporter implements Reporter {
...
+        name: check.name || check.logicalId || 'Unknown Test'
+      }
+      printLn(formatCheckTitle(CheckStatus.RUNNING, displayCheck, { includeSourceFile: true }))
+    }
+    
+    // Format and display each log with proper indentation and timestamp handling
+    logList.forEach((logEntry) => {
+      // Format timestamp from Unix timestamp to HH:mm:ss.SSS format
+      const date = new Date(logEntry.timestamp)
+      const timestamp = date.toISOString().substring(11, 23) // HH:mm:ss.SSS format
```
umutuzgur, line 123: Why substring when you can enforce the format to `HH:mm:ss.SSS`?

packages/cli/src/reporters/abstract-list.ts
```diff
@@ -101,6 +103,44 @@ export default abstract class AbstractListReporter implements Reporter {
...
+      const logLines = formattedLog.split('\n')
+      logLines.forEach((line, index) => {
+        if (line.trim()) { // Only print non-empty lines
+          printLn(indentString(line, 4), 1, 0) // Always add 1 newline after each line
+        }
+      })
+    })
+
+    // Store logs for later use if needed
+    checkFile.logs.push(...logList)
```
sorccu, line 141: Roughly how many logs can we expect? Is this going to consume a ton of memory?

packages/cli/src/reporters/abstract-list.ts
```diff
@@ -101,6 +103,44 @@ export default abstract class AbstractListReporter implements Reporter {
...
+    logList.forEach(logEntry => {
+      // Format timestamp from Unix timestamp to HH:mm:ss.SSS format
+      const timestamp = DateTime.fromMillis(logEntry.timestamp).toFormat('HH:mm:ss.SSS')
+      const formattedLog = `[${timestamp}] ${logEntry.message}`
+
+      // Handle logs that contain newlines by splitting and indenting each line
+      const logLines = formattedLog.split('\n')
+      logLines.forEach((line, index) => {
+        if (line.trim()) { // Only print non-empty lines
+          printLn(indentString(line, 4), 1, 0) // Always add 1 newline after each line
```
sorccu, line 135: I'm a bit confused by this. Won't a multi-line log line end up looking like this:

```
> console.log(require('indent-string')('[22:03:11.739] foo\nbar\nqux', 4))
    [22:03:11.739] foo
    bar
    qux
```

And even for single-line log lines, the line will still be indented with 4 spaces.

Is this as intended?

PR #1112, opened 2025-07-16
feat: add ability to get playwright version from lock file

ferrandiaz:
## Affected Components
* [x] CLI
* [ ] Create CLI
* [ ] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer
Add ability to parse lock files and get the playwright version from them so we can populate it as part of the playwright check


sorccu (changes requested):
This seems overly complicated. Looking at https://github.com/microsoft/playwright/blob/a6cb0093567e3d539f93978362a8fa784c73f2fe/packages/playwright-test/package.json, `@playwright/test` exports its own `package.json`. Is the following not enough?

```
const { version } = require('@playwright/test/package.json')
```

Or am I misunderstanding what exactly you need from there?

ferrandiaz:
@sorccu We are trying to get it from the lock file, so the exact version the user is using.

sorccu:
Do you mean exact version (`1.5.4`) or exact version range (e.g. `^1.5`)? Can you explain a little more? Just tell me exactly what you need and for what reason and I may be able to help. I would like to avoid this hackiness if there's any other alternative.

ferrandiaz:
@sorccu I mean exact version `1.5.4`

sorccu:
That's exactly what you get with the snippet I posted. Though it has one caveat: it only works as long as the CLI itself does not have `@playwright/test` as a dependency, but currently it does not.

Alternatively you could look up the closest `node_modules` folder and attempt to read `node_modules/@playwright/test/package.json` as a file, without require/import.

```
> (await import('@playwright/test/package.json', { with: { type: 'json' } })).default.version
'1.54.1'
> require('@playwright/test/package.json').version
'1.54.1'
```

packages/cli/src/services/util.ts
```diff
@@ -214,16 +229,30 @@ Promise<{outputFile: string, browsers: string[], relativePlaywrightConfigPath: s
...
   return hash.digest('hex');
+}
 
+export async function getPlaywrightVersion(projectDir: string): Promise<string | undefined> {
+  try {
+    const modulePath = path.join(projectDir, 'node_modules', '@playwright', 'test', 'package.json');
+    const packageJson = JSON.parse(await readFile(modulePath, 'utf-8'));
+    return normalizeVersion(packageJson.version);
+  } catch {
+    // If node_modules not found, fall back to checking the project's package.json
```
sorccu, line 245: What's the reason for the fallback? Like, did you actually encounter this issue, and if so, under what conditions (did they not run npm install?), or are you just adding this for good measure?

Since you have to run npm install (or equivalent) to even have access to the checkly CLI in a functional way, I'm a bit confused why this is needed.

sorccu: Hmm, ok. That could be solved by just using require or await import though, because then it would go through the normal resolution system. Or, at the very least `require.resolve` should help if you insist on loading the file manually.

Do you have a reason to not use require/import? Does it make testing harder or...?

ferrandiaz: I wasn't able to make testing work for `require.resolve` to resolve to a specific path (it was using default cli playwright version path) for some reason.

ferrandiaz: cc. @sorccu

packages/cli/src/commands/pw-test.ts
```diff
@@ -12,7 +12,7 @@ import * as api from '../rest/api'
 import config from '../services/config'
 import { parseProject } from '../services/project-parser'
 import type { Runtime } from '../rest/runtimes'
-import { Diagnostics, RuntimeCheck, Session } from '../constructs'
+import { Diagnostics, PlaywrightCheck, RuntimeCheck, Session } from '../constructs'
 import { Flags, ux } from '@oclif/core'
```
sorccu, line 16:
```suggestion
import { Flags } from '@oclif/core'
```

sorccu (approved):
I still think that require() or import() could have been used for this, but alright.

PR #1114, opened 2025-07-21
feat: single select in interactive mode and printing the rules to stdout in non interactive mode

aluedeke:
## Affected Components
* [x] CLI
* [ ] Create CLI
* [ ] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

image


tnolet (commented):
Two things:

- just make the menu single select and show all options. It advertises that we have this and might prompt a user to enable / install some AI LLM coding agent.
- fix the filename written on standard MD

PR #1115, opened 2025-07-21
feat: create non interactive mode

aluedeke:
## Affected Components
* [ ] CLI
* [x] Create CLI
* [ ] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

In non interactive mode no questions are being asked.


sorccu (changes requested):
Check my comments, also there are some unnecessary formatting changes too but I can take care of those after merging when we add ESLint Stylistic (no longer included by default in ESLint 9+).

PR #1124, opened 2025-07-23
fix: fallback to config snapshot path template if project has none [sc-0]

umutuzgur:
## Affected Components
* [x] CLI
* [ ] Create CLI
* [ ] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer
Adds a fallback for the snapshot path template. I forgot to add this fallback and snapshots are not bundled in certain config combinations


ferrandiaz (approved):
LGTM

PR #1144, opened 2025-09-22
chore: minor bug fixes/features for playwright checks

ferrandiaz:
## Affected Components
* [x] CLI
* [ ] Create CLI
* [ ] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer

Minor fixes:
1. Make current dir root when resolving lock files for playwright checks
2. Allow using non-root directory for playwright config file
3. Allow using playwright's own `--config` flag when running `pw-test`

### DIFF
--- a/packages/cli/src/services/util.ts
+++ b/packages/cli/src/services/util.ts
@@ -20,7 +20,12 @@ import { readFile } from 'fs/promises'
 import { createHash } from 'crypto'
 import { Session } from '../constructs'
 import semver from 'semver'
-import { detectNearestLockfile } from './check-parser/package-files/package-manager'
+import {
+  detectNearestLockfile,
+  NpmDetector,
+  PNpmDetector,
+  YarnDetector,
+} from './check-parser/package-files/package-manager'
 
 export interface GitInformation {
   commitId: string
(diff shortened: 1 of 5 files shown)


packages/cli/src/services/util.ts
```diff
@@ -275,10 +282,10 @@ export async function loadPlaywrightProjectFiles (
     const relativePath = path.relative(dir, file)
     archive.file(file, { name: relativePath, mode })
   }
-  // TODO: This code below should be a single glob
-  archive.glob('**/package*.json', { cwd: path.join(dir, '/'), ignore: ignoredFiles }, { mode })
```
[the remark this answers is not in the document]
ferrandiaz: Good point, I will keep the glob for the package.json

packages/cli/src/services/util.ts
```diff
@@ -275,10 +283,11 @@ export async function loadPlaywrightProjectFiles (
     const relativePath = path.relative(dir, file)
     archive.file(file, { name: relativePath, mode })
   }
-  // TODO: This code below should be a single glob
+  const lockFileDirName = path.dirname(lockFile)
+  archive.file(lockFile, { name: path.basename(lockFile), mode })
+  archive.file(path.join(lockFileDirName, 'package.json'), { name: 'package.json', mode })
+  // handle workspaces
   archive.glob('**/package*.json', { cwd: path.join(dir, '/'), ignore: ignoredFiles }, { mode })
```
sorccu, line 290: This will also include package-lock.json and package.json in the project root path. If the same file is added twice, it will still only show up once in the archive, right?

Also, maybe the pattern could be simplified to remove to wildcard from `package*.json` so that it will only match package.json files.

ferrandiaz: Yes, archiver is smart enough, not to duplicate files

sorccu (approved):

PR #1149, opened 2025-09-26
feat: support `group` and group defaults in PlaywrightCheck, deprecate `groupName`

ferrandiaz:
## Affected Components
* [x] CLI
* [ ] Create CLI
* [ ] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer
1. Allow group as parameter on a playwright check
2. Propagate group config to check

### DIFF
--- a/packages/cli/src/constructs/playwright-check.ts
+++ b/packages/cli/src/constructs/playwright-check.ts
@@ -7,7 +7,11 @@ import {
   bundlePlayWrightProject, cleanup,
 } from '../services/util'
 import { RuntimeCheck, RuntimeCheckProps } from './check'
-import { InvalidPropertyValueDiagnostic } from './construct-diagnostics'
+import {
+  ConflictingPropertyDiagnostic,
+  DeprecatedPropertyDiagnostic,
+  InvalidPropertyValueDiagnostic,
+} from './construct-diagnostics'
 import { Diagnostics } from './diagnostics'
 import { PlaywrightCheckBundle } from './playwright-check-bundle'
 import { Session } from './project'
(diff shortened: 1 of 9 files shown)


packages/cli/src/constructs/playwright-check.ts
```diff
@@ -76,6 +77,7 @@ export interface PlaywrightCheckProps extends RuntimeCheckProps {
    *
    * @example "E2E Tests"
    * @example "Critical User Flows"
+   * @deprecated Use groupId instead for better consistency
```
sorccu, line 80: This is not what we want. `groupId` is a deprecated property. What we want to support is the `group` property. Fixing this will also have some implications on the rest of the implementation.

Also, please try to use `{@link}` when making references. Like this:

```suggestion
   * @deprecated Use {@link group} instead.
```

packages/cli/src/constructs/playwright-check.ts
```diff
@@ -159,8 +197,22 @@ export class PlaywrightCheck extends RuntimeCheck {
...
+    this.#validateGroupReferences(diagnostics)
+  }
+
+  #validateGroupReferences (diagnostics: Diagnostics): void {
+    if (this.groupId) {
+      const checkGroup = PlaywrightCheck.#findGroupByRef(this.groupId)
+      if (!checkGroup) {
+        diagnostics.add(new InvalidPropertyValueDiagnostic(
+          'groupId',
+          new Error(`No such group with ID "${this.groupId.ref}".`),
```
sorccu, line 209: This can result in `No such group with ID "undefined".`  if `groupId` is a string. But again, groupId is not what we want (though it should be supported).

sorccu (approved):

PR #1154, opened 2025-10-02
feat: allow passing playwright config path at a check level

ferrandiaz:
## Affected Components
* [x] CLI
* [ ] Create CLI
* [ ] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer
1. Allow passing playwright config path at a check level
2. if no config path is provided at check level fallback to general config path
3. If no config path is provided, fail.

This also allows the user not to define a `playwrightConfigPath` at the top level and just define them at check level if they want.


sorccu (approved):
I think this needs a cleanup at some point but it'll do for now.

PR #1158, opened 2025-10-08
feat(playwright): remove retryStrategy from PlaywrightCheck construct [kit-2258]

maxigimenez:
## Affected Components
* [x] CLI
* [ ] Create CLI
* [ ] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer
Update Playwright constructs and remove `retryStrategy` from the supported properties, also include a validation error.

### DIFF
--- a/packages/cli/src/constructs/playwright-check.ts
+++ b/packages/cli/src/constructs/playwright-check.ts
@@ -19,7 +19,7 @@ import { Session } from './project'
 import { Ref } from './ref'
 import { ConfigDefaultsGetter, makeConfigDefaultsGetter } from './check-config'
 
-export interface PlaywrightCheckProps extends Omit<RuntimeCheckProps, 'retryStrategy'> {
+export interface PlaywrightCheckProps extends Omit<RuntimeCheckProps, 'retryStrategy' | 'doubleCheck'> {
   /**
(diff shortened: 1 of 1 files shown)


sorccu:
There may be a few potential issues with this:

1. `retryStrategy` can be inherited from the Checkly config file. It will trigger the validation error and the user doesn't have a good way to get rid of it, because now that the PW config props don't even include `retryStrategy`, they can't even do `retryStrategy: undefined` (which is already pretty crappy) without `// @ts-ignore`.
 2. The backend will set `doubleCheck: true` by default which then gets converted into an equivalent retry strategy in any API responses. While the backend should handle this, perhaps the easiest way to work around this right now would be to set `doubleCheck: false` in `PlaywrightCheck`'s `synthesize()`. I would also recommend adding validation for `doubleCheck`, basically the equivalent of `validateRemovedDoubleCheck` but with a different error message since `retryStrategy` (which the current error message recommends) is not supported.
 3. Another likely backend issue, but if you retrieve checks with group settings applied (e.g. via `/v1/checks?applyGroupSettings=true`), the group's retry strategy will likely be included in the response. These APIs are not used by the CLI but this is another indication that the backend may need a little more work (note: I did not check whether it already handles this).

PR #1166, opened 2025-10-16
feat: fix specs & skip copy PW config when using examples [kit-2292]

maxigimenez:
## Affected Components
* [ ] CLI
* [x] Create CLI
* [ ] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer
- [x] Fix examples with correct paths & copying playwright config
- [x] Skip "Copy playwright" step during creation when using examples to avoid confusion


ferrandiaz (approved):

sorccu (approved):
Looks good but it would be nice to use single quotes everywhere instead of mixing both. That being said it looks like some of the files are already mixed so maybe not worth handling in this PR.

packages/create-cli/src/commands/bootstrap.ts
```diff
@@ -115,7 +116,8 @@ export default class Bootstrap extends Command {
     await installDependenciesAndInitGit({ projectDirectory })
 
     const playwrightConfig = getPlaywrightConfig(projectDirectory)
-    if (playwrightConfig) {
+    // Only prompt playwright copy when not using exampels
```
miliberlin: @maxigimenez the config is not copied at all anymore then, right?
I just encountered this issue when using the advanced-project-js and prompting to copy the config, but that should go away with this change then?

Screenshot 2025-10-17 at 10 50 16

maxigimenez: > @maxigimenez the config is not copied at all anymore then, right? I just encountered this issue when using the advanced-project-js and prompting to copy the config, but that should go away with this change then?
> 
> Screenshot 2025-10-17 at 10 50 16

@miliberlin you need to install the dependencies in order to copy the config from the playwright file

maxigimenez: this will go away, with the new changes whenever using examples codes it will skip the copy playwrigth step

PR #1172, opened 2025-11-04
fix: use dynamic import for archiver CommonJs

ferrandiaz:
## Affected Components
* [x] CLI
* [ ] Create CLI
* [ ] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer
When using the checkly/playwright-reporter together with the CLI, since there are 2 versions of the same library using a different type (CommonJS, ESM) CLI breaks, by using a dynamic import we fix that issue.

The reporter uses a fork of archiver npm package https://github.com/checkly/node-archiver which allows us to know the byte offset (start/end) of the different files so then we can pick up exactly those files instead of unzipping everything.


hwartig (approved):
LGTM :shipit:

sorccu:
Could you mention in the description what the two versions of `archiver` being used are and how they differ?

PR #1174, opened 2025-11-07
fix: status summary for pw-test

hwartig:
The status summary was overwriting log messages instead of the previous status summary. Now the summary sticks to the bottom of the screen with logs steaming in above it.

Screenshot 2025-11-07 at 20 01 41

## Affected Components
* CLI `pw-test` command


ferrandiaz (approved):
:shipit:

PR #1175, opened 2025-11-10
fix: support IgnoreDirectoriesMatch

ferrandiaz:
## Affected Components
* [x] CLI
* [ ] Create CLI
* [ ] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer
We have a bug in which we are not ignoring the directories/files on this particular field when bundling the playwright check

### DIFF
--- a/packages/cli/src/services/util.ts
+++ b/packages/cli/src/services/util.ts
@@ -210,13 +210,18 @@ export async function bundlePlayWrightProject (
   const output = fsSync.createWriteStream(outputFile)
 
   // Dynamic import for CommonJs so it doesn't break when using checkly/playwright-reporter archiver
-  const { default: archiver } = await import('archiver')
-  const archive = archiver('tar', {
-    gzip: true,
-    gzipOptions: {
-      level: 9,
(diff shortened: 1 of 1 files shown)


packages/cli/src/services/util.ts
```diff
@@ -320,7 +320,10 @@ export async function loadPlaywrightProjectFiles (
     prefix,
   })
   for (const includePattern of include) {
-    archive.glob(includePattern, { cwd: dir }, {
+    archive.glob(includePattern, {
+      cwd: dir,
+      ignore: ignoredFiles,
```
sorccu: This change makes it so that our default ignore patterns (`['**/node_modules/**', '.git/**']`) now also apply to included files. I suppose that's OK though. I haven't tried if it works, but there's a chance that you may have been able to use `include` to add a private node module to the bundle before this change (by doing e.g. `include: ['./node_modules/some-dep/**']`). Even if you could have, I'm not sure if we'd want to support that, though. Thoughts?

PR #1183, opened 2025-11-26
fix: update reporter for changed check result format

sorccu:
The new check result format sometimes uses `null` instead of an empty array. It can cause the following error to manifest under certain circumstances:

    TypeError: Cannot read properties of null (reading 'length')

This change handles a null value too, preventing the error.

The error also broke our E2E tests.

## Affected Components
* [x] CLI
* [ ] Create CLI
* [ ] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer

> Resolves #[issue-number]


sorccu:
This is no longer needed because the backend is receiving a fix.

PR #1195, opened 2026-01-15
fix: shell quoting on playwright flags

ferrandiaz:
## Affected Components
* [x] CLI
* [ ] Create CLI
* [ ] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

### Summary
- Add shared shell quoting utility (`src/services/shell.ts`)
- Fix argument quoting in `pw-test`, `buildTestCommand`, and package manager
- Remove duplicated quoting logic
### Problem
Shell commands were constructed with inconsistent/incomplete quoting:
- `pw-test` only quoted args containing spaces, missing other metacharacters (`|`, `$`, `;`, `*`, etc.)
- `buildTestCommand` had manual quoting that missed edge cases
- `package-manager` had a local `unsafeQuoteArg` function not shared elsewhere
### Solution
Created a shared `shell.ts` utility with consistent quoting that handles all shell metacharacters. Safe characters (`%+,-./:=@_` and alphanumeric) pass through unquoted; everything else gets properly escaped.

### DIFF
--- a/packages/cli/src/services/shell.ts
+++ b/packages/cli/src/services/shell.ts
@@ -3,12 +3,12 @@ const SAFE_SHELL_CHARS = /^[%+,\-./:=@_0-9A-Za-z]+$/
 
 export function shellQuote (arg: string): string {
   if (arg === '') {
-    return '\'\''
+    return `''`
   }
   if (SAFE_SHELL_CHARS.test(arg)) {
     return arg
   }
-  return `'${arg.replace(/'/g, '\'"\'"\'')}'`
+  return `'${arg.replace(/'/g, `'"'"'`)}'`
 }
 
 export function shellJoin (args: string[]): string {
--- a/packages/cli/src/services/__tests__/shell.spec.ts
+++ b/packages/cli/src/services/__tests__/shell.spec.ts
@@ -4,7 +4,7 @@ import { shellQuote, shellJoin } from '../shell'
 describe('shell', () => {
   describe('shellQuote', () => {
     it('returns empty string as quoted', () => {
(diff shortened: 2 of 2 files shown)


packages/cli/src/services/shell.ts
```diff
@@ -0,0 +1,26 @@
...
+  if (arg === '') {
+    return '\'\''
+  }
+  if (SAFE_SHELL_CHARS.test(arg)) {
+    return arg
+  }
+  return `'${arg.replace(/'/g, '\'"\'"\'')}'`
+}
+
+export function shellQuoteDouble (arg: string): string {
```
sorccu: Are you sure we want a function that does double quotes? While these are not meant to be perfectly safe (as the previous name I used indicated), double quotes are more trouble because they allow things like variable interpolation. You'd have to cover at least `$` and possibly a few more characters to make it safer (though never perfectly safe). It's just a lot of trouble. Why not use single quotes everywhere?

sorccu (changes requested):
See my comment.

packages/cli/src/services/shell.ts
```diff
@@ -0,0 +1,26 @@
+// Safe shell characters: %+,-./:=@_ and alphanumeric
+const SAFE_SHELL_CHARS = /^[%+,\-./:=@_0-9A-Za-z]+$/
+
+export function shellQuote (arg: string): string {
+  if (arg === '') {
+    return '\'\''
```
sorccu: Alright but not sure why you changed these. I thought it was more readable with backticks instead of this forest of backslashes.

PR #1201, opened 2026-02-02
fix: remove double test line printed with pw-test [red-114]

pilimartinez:
## Affected Components
* [ ] CLI
* [ ] Create CLI
* [ ] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer

In pw-test we can avoid having the double line on the logs, check [here](https://linear.app/checklyhq/issue/RED-114/remove-double-line-printed-with-pw-test) and Before screenshot. 

Before

image

After

image


hwartig (approved):
👍

PR #1205, opened 2026-02-04
fix: validate logicalId via diagnostics instead of throwing at construction time [red-71]

miliberlin:
## Summary

Instead of throwing a hard error at parse time when a logicalId contains invalid characters, we now report it as a validation diagnostic. This means:

1. Parsing completes successfully even with invalid logicalIds
2. **All** invalid logicalIds are reported at once during validation
3. Users can fix everything in one go instead of fixing one, re-running, finding the next, etc.

No sanitization or interactive prompts — just clear error diagnostics.

## Example

```
Parsing your project... ✅

Validating project resources... ❌
✖ Invalid property value

  The value provided for property "logicalId" is not valid.

  Reason: The logicalId "boilerplate project (invalid)" contains invalid
  characters. Only A-Z, a-z, 0-9, _, -, /, #, and . are allowed.

✖ [ApiCheck:books api check (v1)] Invalid property value

  The value provided for property "logicalId" is not valid.

  Reason: The logicalId "books api check (v1)" contains invalid characters.
  Only A-Z, a-z, 0-9, _, -, /, #, and . are allowed.

✖ Your project is not valid.
```

## Affected Components
* [x] CLI
* [ ] Create CLI
* [ ] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Changes

- Move logicalId character validation from `Session.validateCreateConstruct` (constructor-time error) to `Construct.validate()` (diagnostic)
- Remove `sanitizeLogicalId` and interactive confirmation prompt approach
- Delete `helpers/sanitized-logical-id.ts`
- Extract `LOGICAL_ID_PATTERN` to `constants.ts` for reuse
- Convert `CheckTypes` from default export to named export
- Add tests for logicalId validation via diagnostics

### DIFF
--- a/packages/cli/src/constructs/incidentio-alert-channel.ts
+++ b/packages/cli/src/constructs/incidentio-alert-channel.ts
@@ -33,7 +33,7 @@ export interface IncidentioAlertChannelProps extends AlertChannelProps {
 export class IncidentioAlertChannel extends WebhookAlertChannel {
   static DEFAULT_PAYLOAD = `{
   "title": "{{ALERT_TITLE}}",
-  "description": "{{ALERT_TITLE}} at {{STARTED_AT}} in {{RUN_LOCATION}} {{RESPONSE_TIME}}ms",
+  "description": "{{ALERT_TITLE}} at {{STARTED_AT}} in {{RUN_LOCATION}} {{RESPONSE_TIME}}ms\\n\\n{{#if AI_ANALYSIS_CLASSIFICATION}}# AI Analysis\\n\\n{{AI_ANALYSIS_CLASSIFICATION}}\\n\\n{{{AI_ANALYSIS_ROOT_CAUSE}}}\\nRead full analysis: {{AI_ANALYSIS_LINK}}{{/if}}",
   "deduplication_key": "{{CHECK_ID}}",
   "metadata": {
     "alertType": "{{ALERT_TYPE}}",
--- a/packages/cli/src/constructs/msteams-alert-channel.ts
+++ b/packages/cli/src/constructs/msteams-alert-channel.ts
@@ -43,9 +43,53 @@ export class MSTeamsAlertChannel extends WebhookAlertChannel {
               {
                 "type": "TextBlock",
                 "text": "{{ALERT_TITLE}}",
-                "weight": "bolder",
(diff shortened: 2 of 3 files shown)


packages/cli/src/constructs/project.ts
```diff
@@ -293,15 +293,15 @@ export class Session {
     }
   }
 
+  static sanitizeLogicalId (logicalId: string): string {
+    return logicalId.replace(/[^A-Za-z0-9_\-/#.]/g, '')
```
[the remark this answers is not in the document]
miliberlin: I updated the PR to prompt confirmation and not just overwrite the logicalIds, but yes, the source code will remain unchanged.

sorccu:
I would want to remove the current validation from the constructor and use the validate（） method instead. Of course, there are some drawbacks if the construct can store a potentially unclean value, but as long as the validation is error level I think it should be good enough and far better than the current behavior we have.

I would rather not interactively ask the user.

sorccu:
If you use the validate method, you can use the hidden validate command to try it out.

npx checkly validate

packages/cli/src/constructs/project.ts
miliberlin: Nit: `Project.logicalId` re-declares and re-assigns the property already defined on the base `Construct` class. We could remove it to avoid the shadowing, but I kept it for now to minimize changes. Should we clean this up or keep for clarity?

miliberlin:
@sorccu I updated the approach, can you check?

packages/cli/src/constructs/__tests__/project.spec.ts
```diff
@@ -0,0 +1,65 @@
...
+    Session.project = { addResource: () => {} } as any
+  })
+
+  it('should pass validation for valid logicalIds', async () => {
+    const validIds = ['my-check', 'my_check', 'myCheck123', 'my/check', 'my#check', 'my.check']
+    for (const id of validIds) {
+      const construct = new TestConstruct(id)
+      const diagnostics = new Diagnostics()
+      await construct.validate(diagnostics)
+      expect(diagnostics.isFatal(), `Expected "${id}" to be valid`).toBe(false)
```
sorccu: What's that message about?

PR #1207, opened 2026-02-04
feat: add helpful message when pw-test finds no tests [red-125]

miliberlin:
## Summary
- When `pw-test` encounters "No tests found", display a helpful hint suggesting to check the Playwright configuration
- Helps users understand they need to verify their Playwright config targets the correct test files

## Test plan
- [x] Run `pw-test` with a Playwright config pointing to a non-existent test directory
- [x] Verify the helpful message appears after the "No tests found" error


hwartig (approved):
LGTM :shipit:

PR #1211, opened 2026-02-05
fix: exclude runtimeId from PlaywrightCheckProps

stefanjudis:
The runtimeId property should not be configurable for Playwright checks since they use a fixed runtime.

## Affected Components
* [x] CLI
* [ ] Create CLI
* [ ] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer

Just fixing the typings for `PlaywrightCheck` as it doesn't use a runtime

Context: 

https://checklyhq.slack.com/archives/C04PFSV5W3B/p1770309102867809

### DIFF
--- a/packages/cli/src/constructs/playwright-check.ts
+++ b/packages/cli/src/constructs/playwright-check.ts
@@ -21,7 +21,7 @@ import { Session } from './project'
 import { Ref } from './ref'
 import { ConfigDefaultsGetter, makeConfigDefaultsGetter } from './check-config'
 
-export interface PlaywrightCheckProps extends Omit<RuntimeCheckProps, 'retryStrategy' | 'doubleCheck'> {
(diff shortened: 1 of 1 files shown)


stefanjudis:
Closing because this fix change requires [more changes to do it properly](https://checklyhq.slack.com/archives/C04PFSV5W3B/p1770375471784899?thread_ts=1770309102.867809&cid=C04PFSV5W3B).

> I have a slight issue with this change which is that PlaywrightCheck inherits from RuntimeCheck which by definition brings in the runtime. I think just plucking it out of the type is a reasonable thing to do but at the same time maybe PlaywrightCheck should not be a RuntimeCheck (though there are some type checks in the code that would also need to be updated, so it wouldn't be super straightforward)

PR #1216, opened 2026-02-10
feat: add checks list command using versioned API

thebiglabasky:
## Summary
- Adds `checkly checks list` CLI command using the new versioned `/api/checks` endpoint
- Adds REST client with auto-pagination support and `X-Checkly-API-Version` header
- Supports filtering by check type, tag, muted, and activated status

## Companion PR
- monorepo: https://github.com/checkly/monorepo/pull/424 (backend versioned API surface)

## Note
This is an **exploration/discussion PR** — not intended for merge as-is.

## Test plan
- [ ] Run `npx checkly checks list` against a test account
- [ ] Verify filtering flags work (`--check-type`, `--tag`, `--activated`, `--muted`)
- [ ] Verify `--json` output format

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -64,6 +64,9 @@
       "message": "<%= config.name %> update available from <%= chalk.greenBright(config.version) %> to <%= chalk.greenBright(latest) %>. To update, run `npm install -D checkly@latest`"
     },
     "topics": {
+      "checks": {
+        "description": "Manage and inspect Checkly checks."
+      },
       "env": {
         "description": "Manage Checkly environment variables."
       },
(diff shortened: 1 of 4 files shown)


MichaelHogers:
@thebiglabasky hey hey, i would personally go for:

- publish contract from backend upon merge into main, public npm package for our CLI (because our CLI is a public repo)
- import this contract into CLI, so we don't hardcode any types/paths/wahtever here, all should just be a single source of truth coming from backend 

we should nail this tomorrow, Javier did this alrdy but next step is to move to https://trpc.io/ and allow to create domain specific sdks (@kiroushi)

PR #1223, opened 2026-02-16
fix: skip PlaywrightCheck constructs during pw-test [RED-155]

miliberlin:
## Background

Reported internally by Stefan: when a project has a `PlaywrightCheck` construct defined in a `.check.ts` file, `pw-test` incorrectly includes it alongside the vanilla Playwright tests. This causes `pw-test` to run **two checks** — the construct-based check and the ad-hoc one — when it should only run the ad-hoc one:

```
__checks__/critical-suite.check.ts
  ✔ Critical User Flows (8s)
playwright.config.ts
  ✖ Playwright Test:  (11s)
```

The construct-based check passes (self-contained config/deps), but the ad-hoc `playwright.config.ts` check fails because the construct's bundling context clobbers its dependency resolution.

## Root cause

`parseProject()` was not scoped to the `pw-test` command:

1. `loadAllCheckFiles()` ran for all commands including `pw-test`, picking up `.check.ts` files that contain `new PlaywrightCheck(...)` constructs
2. `checklyConfigConstructs` (from config file evaluation) were added to the project unfiltered, even for `pw-test`

This meant `pw-test` would end up with extra PlaywrightCheck instances in the project. Since the `checkFilter` keeps all `PlaywrightCheck` instances, both the construct-based check and the ad-hoc check made it into the bundle. The construct's dependency tree interfered with the ad-hoc check's bundling.

## Changes

Two changes in `parseProject()` when `currentCommand === 'pw-test'`:

1. **Filter PlaywrightCheck constructs** from `checklyConfigConstructs` before adding to the project
2. **Skip `loadAllCheckFiles()`** entirely

## Before/after (released CLI v7.0.1 vs this fix)

Before:
```
Running 2 checks in eu-central-1.

playwright.config.ts
  ✔ Playwright Test (15s)
src/__checks__/synthetics/07-pw-check-construct.check.ts
  ✔ File Based PW Check (should be ignored by pw-test) (13s)

2 passed, 2 total
```

After:
```
Running 1 checks in eu-central-1.

playwright.config.ts
  ✔ Playwright Test:  (15s)

1 passed, 1 total
```

## Test results

Each test project includes a `PlaywrightCheck` construct in a `.check.ts` file, a local project dependency (`src/utils/selectors.ts`), and an npm devDependency (`lodash`). The test files import both to verify dependency resolution works correctly.

| Issue   | Scenario                               | Before (v7.0.1) | After (fix) |
|---------|----------------------------------------|------------------|-------------|
| RED-122 | PlaywrightCheck in `.check.ts` file    | 2 checks run     | 1 check run |
| RED-119 | PW construct causes config to be ignored | 2 checks run   | 1 check run |
| RED-118 | PW construct blocks dependency resolution | 2 checks run  | 1 check run |

All three child issues verified fixed. The released CLI (v7.0.1) consistently runs 2 checks (the bug), while the fix correctly runs only 1 ad-hoc check. Both local and npm dependencies resolve correctly with the fix.

Before sessions: [RED-122](https://chkly.link/l/a3o6P) · [RED-119](https://chkly.link/l/suJdq5) · [RED-118](https://chkly.link/l/lh-H8)
After sessions: [RED-122](https://chkly.link/l/SRPMC) · [RED-119](https://chkly.link/l/WQYOx) · [RED-118](https://chkly.link/l/4vJs-)

### DIFF
--- a/packages/cli/src/constructs/api-check.ts
+++ b/packages/cli/src/constructs/api-check.ts
@@ -11,7 +11,6 @@ import { DeprecatedPropertyDiagnostic, InvalidPropertyValueDiagnostic } from './
 import { ApiCheckBundle, ApiCheckBundleProps } from './api-check-bundle'
 import { Assertion } from './api-assertion'
 import { validateResponseTimes } from './internal/common-diagnostics'
-import { Runtime } from '../runtimes'
 
 /**
  * Default configuration that can be applied to API checks.
--- a/packages/cli/src/services/__tests__/util.spec.ts
+++ b/packages/cli/src/services/__tests__/util.spec.ts
@@ -30,17 +30,17 @@ describe('util', () => {
   })
 
   describe('getPlaywrightVersionFromPackage()', () => {
-    it('should throw error when playwright package is not found', () => {
+    it('should throw error when playwright package is not found', async () => {
       // Use a directory that doesn't have playwright installed
       const nonExistentDir = '/tmp/non-existent-dir'
-      expect(() => getPlaywrightVersionFromPackage(nonExistentDir))
-        .toThrow('Could not find @playwright/test package. Make sure it is installed.')
+      await expect(getPlaywrightVersionFromPackage(nonExistentDir))
+        .rejects.toThrow('Could not find @playwright/test package. Make sure it is installed.')
     })
 
-    it('should get version from installed playwright package', () => {
+    it('should get version from installed playwright package', async () => {
       // Use the current working directory which should have playwright installed
       const currentDir = process.cwd()
-      const version = getPlaywrightVersionFromPackage(currentDir)
+      const version = await getPlaywrightVersionFromPackage(currentDir)
 
       // Should return a valid semver version
       expect(version).toMatch(/^\d+\.\d+\.\d+/)
--- a/packages/cli/src/services/util.ts
+++ b/packages/cli/src/services/util.ts
@@ -16,6 +16,7 @@ import { createHash } from 'crypto'
 import { Session } from '../constructs/project'
 import semver from 'semver'
 import { existsSync } from 'fs'
+import { detectNearestPackageJson } from './check-parser/package-files/package-manager'
 
 export interface GitInformation {
   commitId: string
@@ -184,7 +185,7 @@ export async function bundlePlayWrightProject (
 
   const pwConfigParsed = new PlaywrightConfig(filePath, pwtConfig)
 
-  const playwrightVersion = getPlaywrightVersionFromPackage(dir)
+  const playwrightVersion = await getPlaywrightVersionFromPackage(dir)
 
   const [cacheHash] = await Promise.all([
     getCacheHash(lockfile),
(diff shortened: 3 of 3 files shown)


packages/cli/src/constructs/__tests__/fixtures/playwright-check/test-cases/test-webServer/checkly.config.ts
miliberlin: Summary of why we need to tweak the config for this test below. I think this is correct since this is how a user would set up the project, but please double-check.

> Before our fix, the webServer test worked like this:
> 
> 1. parseProject() runs with --emulate-pw-test
> 2. loadAllCheckFiles() picks up test.check.ts, which calls new PlaywrightCheck(...) with playwrightConfigPath:  './playwright.config.ts'
> 3. The PlaywrightCheck constructor runs validateWebServerConfig(), sees webServer in the playwright config, and emits the warning
> 
> Our fix adds if (currentCommand !== 'pw-test') around loadAllCheckFiles(), so step 2 never happens — test.check.ts is never loaded, no PlaywrightCheck is created, no warning fires.
> 
> But there's a second code path that creates PlaywrightCheck instances: loadPlaywrightChecks(). This function still runs for pw-test. It creates an "ad-hoc" PlaywrightCheck when playwrightConfigPath is set in the checkly config — this is the same path that real users hit when they have checks: { playwrightConfigPath: './playwright.config.ts' } in their checkly.config.ts.
> 
> By adding playwrightConfigPath to the fixture's config, the ad-hoc check gets created via loadPlaywrightChecks() instead of via loadAllCheckFiles(). The webServer validation fires on this ad-hoc check, and the test passes.
> 
> This also makes the test more realistic — in practice, pw-test users always have playwrightConfigPath in their config (that's how pw-test knows which playwright config to use). The old fixture relied on a .check.ts file to create the PlaywrightCheck, which is exactly the scenario our fix is designed to skip.

packages/cli/src/services/project-parser.ts
```diff
@@ -186,7 +190,9 @@ export async function parseProject (opts: ProjectParseOpts): Promise<Project> {
   // TODO: Do we really need all of the ** globs, or could we just put node_modules?
   const ignoreDirectories = ['**/node_modules/**', '**/.git/**', ...ignoreDirectoriesMatch]
 
-  await loadAllCheckFiles(directory, checkMatch, ignoreDirectories)
+  if (currentCommand !== 'pw-test') {
```
sorccu: I don't like this `currentCommand` thing, especially now that it's being used more. It feels like we're just adding a whole bunch of special cases based on the identity of the caller, which is IMO unpredictable and just ass in general. Instead of the caller identifying itself, it should request specific behavior via a relevant flag.

Maybe we can keep this as is if you need this out soon but I do NOT want to see this become a standard.

miliberlin:
@sorccu is this good to go now?

packages/cli/src/services/project-parser.ts
```diff
@@ -166,7 +166,11 @@ export async function parseProject (opts: ProjectParseOpts): Promise<Project> {
     ignoreWorkspaces: !enableWorkspaces,
   })
 
-  checklyConfigConstructs?.forEach(
+  const filteredConstructs = loadPlaywrightChecksOnly
+    ? checklyConfigConstructs?.filter(c => !(c instanceof PlaywrightCheck))
+    : checklyConfigConstructs
```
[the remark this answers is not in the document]
miliberlin: @sorccu Ferran approved this, but I didn't merge it yet to give you time to clear your concerns. Are we good to go in your opinion?

PR #1229, opened 2026-02-18
feat: allow users to ignore existing pwcs dependency cache

hwartig:
Users can pass `--refresh-cache` to ignore existing dependency caches and rebuild the cache for future runs.

## Affected Components
* CLI


sorccu (approved):
Looks good to me. I assume the test failure is due to the backend not having been deployed yet.

PR #1248, opened 2026-03-06
fix(cli): add V2 asset support for logs and check-run-data retrieval [AI-61]

martzoukos:
## Summary

- Add V2 asset handling to `packages/cli/src/rest/assets.ts` so that `checkly test --verbose` displays logs and check-run-data for accounts using the `LEGACY_RUNNER_ASSETS_V2` feature flag
- Fix `config.get` crash in empty-account e2e tests by guarding with `config.has`

## Problem

[monorepo#596](https://github.com/checkly/monorepo/pull/596) (`feat: add support for v2 assets for BROWSER/MULTI_STEP checks`) changed how runners store check result assets. V1 uploaded each asset as a separate S3 file (`logPath`,  `checkRunDataPath`, etc.). V2 bundles everything into a single ZIP archive with byte-range offsets per entry.

The runner side was updated but the CLI was not. The `getAssets` method in `assets.ts` only handled V1 and V3 formats —  when it received a V2 asset, `key` stayed `undefined` and the method silently returned nothing. This meant `npx checkly test --verbose` showed no logs or check-run-data for BrowserCheck and MultiStepCheck runs on affected accounts.

The V2 types (`CheckResultAssetsV2`, `isCheckResultAssetsV2`) were already defined in the file but never wired into the retrieval logic.

## Fix

Added a `getAssetsV2` method that:
1. Looks up the requested entry by name in the `entries` array (`logs.txt` for logs, `check-run-data.json` for check-run-data)
2. Fetches a presigned S3 URL via the existing `/redirect?link=true` endpoint
3. Uses an HTTP `Range` request to fetch just the needed bytes from the ZIP
4. Parses and returns the JSON

### DIFF
--- a/packages/cli/src/rest/assets.ts
+++ b/packages/cli/src/rest/assets.ts
@@ -1,3 +1,4 @@
+import axios from 'axios'
 import type { AxiosInstance } from 'axios'
 
 // eslint-disable-next-line no-restricted-syntax
@@ -103,6 +104,8 @@ export default class Assets {
       case AssetType.LOG: {
         if (isCheckResultAssetsV1(asset) && asset.logPath) {
           key = asset.logPath
+        } else if (isCheckResultAssetsV2(asset)) {
+          return this.getAssetsV2(assetType, asset)
         } else if (isCheckResultAssetsV3(asset) && asset.logPath) {
           key = asset.clickHouseId
         }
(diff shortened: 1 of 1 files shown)


martzoukos:
sweating-nervous

Ok, let's wait for the tests to pass in the PR, first. 😅

PR #1269, opened 2026-03-24
refactor: consolidate confirmOrAbort across deploy, destroy, and env rm

thebiglabasky:
## Summary

Consolidates write-command confirmation around `confirmOrAbort` for `deploy`, `destroy`, and `env rm`, and fixes the follow-up issues found while manually validating the branch in both interactive and agent/CI-style flows.

- `authCommand.ts` — `confirmOrAbort` supports an optional `interactiveConfirm` callback for commands needing custom confirmation in interactive mode.
- `env rm` — uses `confirmOrAbort` instead of ad-hoc `prompts()` confirmation.
- `destroy` — uses `confirmOrAbort` plus `interactiveConfirm`, preserving the type-the-project-name safeguard for interactive users while returning structured confirmation JSON in agent/CI mode.
- `deploy` — uses `confirmOrAbort` and now exits for confirmation before parsing, bundling, or uploading anything in agent/CI mode. This keeps confirmation output machine-readable and avoids pre-confirm side effects.
- `baseCommand.ts` / `authCommand.ts` — local-dev version replacement notices are suppressed outside interactive mode, and `AuthCommand.init()` now awaits `BaseCommand.init()` so interactive notices appear in the correct order.
- `api.spec.ts` — CLI-mode detection tests now clear `CODEX_*` and related env vars so the suite is stable when run from Codex/Desktop environments.

### No breaking changes
- `--force` / `-f` preserved on all three commands
- interactive confirmation behavior preserved
- agent/CI confirmation still exits with code `2` and a structured `confirmation_required` payload
- `deploy --preview` remains a server-side dry run and does not use the confirmation gate

Closes AI-163

## Test plan

- [x] Unit tests for `interactiveConfirm` callback in `confirm-or-abort.spec.ts`
- [x] Confirmation flow tests for `env rm`, `destroy`, and `deploy`
- [x] Regression test that `deploy` exits for confirmation before parsing/bundling in agent mode
- [x] Regression test that `AuthCommand.init()` awaits `BaseCommand.init()`
- [x] Command metadata validation (readOnly/destructive/idempotent)
- [x] `npm run prepare` succeeds
- [x] AI context regenerated successfully
- [x] Full local test suite passes
- [x] `checkly` workspace: 709 passed, 2 skipped
- [x] `create-checkly` workspace: 19 passed

## Manual validation notes

- Reproduced the original issue where `CI=1 checkly deploy` emitted progress output before the confirmation JSON envelope.
- Verified the fix locally: confirmation now happens before deploy parsing/bundling work, and interactive notice ordering is restored.

### DIFF
--- a/packages/cli/src/commands/__tests__/auth-command-init.spec.ts
+++ b/packages/cli/src/commands/__tests__/auth-command-init.spec.ts
@@ -28,7 +28,10 @@ describe('AuthCommand.init', () => {
     expect(api.validateAuthentication).not.toHaveBeenCalled()
 
     releaseBaseInit()
-    await expect(initPromise).rejects.toThrow('Cannot write private member #account')
+    // This intentionally invokes AuthCommand.init with a fake `this`, so the
+    // eventual rejection is incidental. The behavior under test is that auth
+    // validation does not begin until BaseCommand.init has finished.
+    await expect(initPromise).rejects.toThrow()
 
     expect(api.validateAuthentication).toHaveBeenCalledTimes(1)
 
--- a/packages/cli/src/commands/destroy.ts
+++ b/packages/cli/src/commands/destroy.ts
@@ -31,8 +31,7 @@ export default class Destroy extends AuthCommand {
       command: 'destroy',
       description: 'Destroy all project resources',
       changes: [
-        `PERMANENTLY delete ALL resources associated with the project "${checklyConfig.projectName}"`,
-        `Account: "${account.name}"`,
+        `PERMANENTLY delete ALL resources associated with the project "${checklyConfig.projectName}" in account "${account.name}"`,
       ],
       flags,
       classification: {
@@ -48,6 +47,9 @@ export default class Destroy extends AuthCommand {
           type: 'text',
           message: `Type the project name "${checklyConfig.projectName}" to confirm:`,
         })
+        if (projectName === undefined) {
+          return false
+        }
         if (projectName !== checklyConfig.projectName) {
           this.log(`The entered project name "${projectName}" doesn't match the expected project name "${checklyConfig.projectName}".`)
           return false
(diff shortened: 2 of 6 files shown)


packages/cli/src/commands/__tests__/auth-command-init.spec.ts
```diff
@@ -0,0 +1,37 @@
...
+    const ctx = {}
+    const initPromise = AuthCommand.prototype.init.call(ctx as any)
+
+    await Promise.resolve()
+
+    expect(baseInitSpy).toHaveBeenCalledTimes(1)
+    expect(api.validateAuthentication).not.toHaveBeenCalled()
+
+    releaseBaseInit()
+    await expect(initPromise).rejects.toThrow('Cannot write private member #account')
```
sorccu: Seems like a pretty weird thing to test for. I assume you don't actually care about what the result of the initPromise is as long as it resolves somehow? If that's the case, I'd appreciate a comment making that clearer.

packages/cli/src/commands/destroy.ts
```diff
@@ -25,21 +22,40 @@ export default class Destroy extends AuthCommand {
...
+      },
+    }, {
+      force: flags.force,
+      interactiveConfirm: async () => {
+        const { projectName } = await prompts({
+          name: 'projectName',
+          type: 'text',
+          message: `Type the project name "${checklyConfig.projectName}" to confirm:`,
+        })
+        if (projectName !== checklyConfig.projectName) {
```
sorccu: I know this is the same as before but we should handle cancellation better. This is how it looks like if you cancel via ctrl+c  or esc:

```
This will:
  - PERMANENTLY delete ALL resources associated with the project "temp-project-0E6CA056-1778-4577-8B8E-BA38470E689C"
  - Account: "simo@checklyhq.com"

✖ Type the project name "temp-project-0E6CA056-1778-4577-8B8E-BA38470E689C" to confirm: …
The entered project name "undefined" doesn't match the expected project name "temp-project-0E6CA056-1778-4577-8B8E-BA38470E689C".
```

This should do it:

```suggestion
        if (projectName === undefined) {
          return false
        }
        if (projectName !== checklyConfig.projectName) {
```

PR #1286, opened 2026-04-10
docs(cli): trim SKILL.md and emphasize CLI-first lookup

stefanjudis:
## Affected Components
* [x] CLI
* [ ] Create CLI
* [ ] Test
* [x] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer
Trims the always-loaded `SKILL.md` so agents lean on `npx checkly skills <action>` for progressive disclosure instead of loading every sub-reference upfront.

- Removes the per-reference sub-command listings (`#### npx checkly skills <action> <ref>`) from `SKILL.md`.
- Rewrites the intro with a **Required:** directive telling agents to run `npx checkly skills` first, since the CLI is the source of truth and actions change between releases.
- Extends the skill `description` to mention account plan, entitlements, and feature limits.
- Drops the now-unused inner `references` loop from `generateSkillCommands()` in `prepare-ai-context.ts` and regenerates `skills/checkly/SKILL.md` to match.

Tradeoff worth flagging: agents reading only `SKILL.md` no longer see hints like `configure api-checks` directly — they have to run `npx checkly skills configure` to discover sub-references. Intentional, since we want to serve as much as possible from the CLI.

### DIFF
--- a/packages/cli/scripts/prepare-ai-context.ts
+++ b/packages/cli/scripts/prepare-ai-context.ts
@@ -80,15 +80,6 @@ function generateSkillCommands (): string {
       action.description,
     ]
 
-    if ('references' in action) {
-      for (const ref of action.references) {
-        const refId = ref.id.replace(`${action.id}-`, '')
-        lines.push('')
-        lines.push(`#### \`npx checkly skills ${action.id} ${refId}\``)
-        lines.push(ref.description)
-      }
-    }
-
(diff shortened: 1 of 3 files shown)


sorccu:
@stefanjudis If you resolve the conflicts and rebase the branch, tests should pass again. Following that we can merge.

sorccu (approved):

PR #1291, opened 2026-04-22
fix(cli): scope checks list summary bar to account-wide activated state

MichaelHogers:
## Problem

`checks list`'s summary bar combined **unfiltered** statuses with the **filtered** `paginated.total`, producing nonsense like `6 passing  1 failing  (1 total checks)` on `--status=failing`. Same bug in `stats.ts`. Secondary symptom: deactivated checks' stale statuses were silently counted as passing.

## Fix

Account-wide summary bar, never reacts to filters/pagination — matches the webapp's top cards.

```
Account wide:  ✓ 353 passing   ⚠ 9 degraded   ✖ 72 failing   ⊘ 11 inactive
```

- New pure helper `getActivatedStatuses(checks, statuses)` excludes stale statuses from deactivated checks.
- `formatSummaryBar(checks, statuses)` — new signature, drops the misleading `(N total)` label and the misused `activeCheckIds` param.
- `list.ts` and `stats.ts` add a parallel `checks.fetchAll()` to drive the bar independently of the filtered table.
- Explicit `⊘ N inactive` bucket because the CLI mixes deactivated into the table (webapp has separate tabs for them).
- Dim `Account wide:` prefix makes scope unambiguous when a filter is applied below.

## Tests

6 new unit tests using committed, sanitized fixtures under `rest/__tests__/__fixtures__/api/` (scenario helper + README documenting API quirks we verified on prod).

## Test plan

- [ ] `./bin/run checks list` — `Account wide:` bar with 4 buckets, no total
- [ ] `./bin/run checks list --status failing` — bar numbers identical to unfiltered
- [ ] `./bin/run checks list --output json` — shape unchanged
- [ ] `./bin/run checks stats` — same `Account wide:` bar
- [ ] `npm run test` passes


MichaelHogers:
@stefanjudis @thebiglabasky this should fix the awkwardness/confusion that the status bar caused, we now apply the same pattern/behavior as our web-app: always show the account wide status, regardless of filters applied

image

image

packages/cli/src/commands/checks/list.ts
```diff
@@ -71,6 +72,7 @@ export default class ChecksList extends AuthCommand {
           search: flags.search,
           status: flags.status,
         }),
+        api.checks.fetchAll(),
```
thebiglabasky, line 75: If we fetch all, why also fetching paginated? That looks like quite a waste. I get we're skipping the pagination logic that'd need to be replicated locally, but also 😩 
Idk, that feels a bit much.

MichaelHogers: Yeah, I need to dig a bit deeper tomorrow

Basically; the status bar requires data that is not efficiently returned via our public api, i'll check tomorrow if there's an easy workaround, extending the public api or simplifying the status bar  further

MichaelHogers:
@thebiglabasky on second consideration; i am looking at adding an aggregation endpoint in our public API `/v1/check-statuses/summary`, and always showing the entire account wide check status overview (not reactive to filters or paging, like the webapp)

we can also skip the account wide summary in the CLI `checkly checks list` command

let me know what you think/prefer

thebiglabasky:
> @thebiglabasky on second consideration; i am looking at adding an aggregation endpoint in our public API `/v1/check-statuses/summary`, and always showing the entire account wide check status overview (not reactive to filters or paging, like the webapp)
> 
> we can also skip the account wide summary in the CLI `checkly checks list` command
> 
> let me know what you think/prefer

I think the endpoint is a good way to go! I assume this would take filtering options too in case one wants a specific kind of summary then?
It could also likely (not sure about the exact details though) achieve the same as what the home dashboard's top bar has with the card-per-status with counts.

MichaelHogers:
> > @thebiglabasky on second consideration; i am looking at adding an aggregation endpoint in our public API `/v1/check-statuses/summary`, and always showing the entire account wide check status overview (not reactive to filters or paging, like the webapp)
> > we can also skip the account wide summary in the CLI `checkly checks list` command
> > let me know what you think/prefer
> 
> I think the endpoint is a good way to go! I assume this would take filtering options too in case one wants a specific kind of summary then? It could also likely (not sure about the exact details though) achieve the same as what the home dashboard's top bar has with the card-per-status with counts.

yes exactly

note that the webapp's home dashboard top bar's card-per-status with counts is completely non-reactive to filters (for a complexity reason i think)

we could make the CLI status bar responsive only to the use of `--tags`, `--(check)type` when using `checkly list`, but i wouldn't go much further than that after thoroughly checking the flags we have

PR #1292, opened 2026-04-28
feat(cli): cancel running checks on Ctrl+C

ferrandiaz:
## Affected Components
* [x] CLI
* [ ] Create CLI
* [x] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer

Pressing Ctrl-C during `checkly pw-test`, `checkly test`, or `checkly trigger` now cancels the in-flight test session server-side via a new `/v1/cancel` REST endpoint, instead of silently killing the CLI and leaving the remote session running.

### Motivation

Before this change, Ctrl-C terminated the CLI without signalling the backend — the remote session kept consuming cloud resources with no way to stop it.

### What changed

- **New `Cancel` REST client** (`rest/cancel.ts`) — `POST /v1/cancel` via `cancelTestSession` and `cancelCheckSession`; registered as `api.cancel` in `api.ts`
- **SIGINT handler in `AbstractCheckRunner`** — on first Ctrl+C: emits `Events.CANCEL` and the reporter shows a cancellation message with a `--detach` hint; on second Ctrl+C: force-exits with code 1. Pre-existing SIGINT listeners (from `when-exit`, a transitive dependency) are removed during the run and restored afterward to prevent them from terminating the process on Windows.
- **`--detach` flag** — when used, first Ctrl+C shows "Checks will continue running in the cloud." and exits cleanly with code 0 (checks keep running server-side)
- **Three CLI commands** (`pw-test`, `test`, `trigger`) wire `Events.CANCEL` and `Events.DETACH` to reporters and the cancel API
- **Reporters** — new `onCancel` / `onDetach` lifecycle on the `Reporter` interface; cancellation/detach messages rendered as part of the live summary panel; `CheckStatus.CANCELLED` with `⊘` symbol; cancelled-count in summary line

### Tests

Unit tests across cancel client (4), runner SIGINT lifecycle (9), reporter cancel/detach UI (4), util (24).

### Notes

- `cancelCheckSession` is wired but not called from any command in this PR — reserved for a future integration
- Any custom `Reporter` implementations outside this repo need stubs for `onCancel()` and `onDetach()`


packages/cli/e2e/__tests__/cancel-prompt.spec.ts
```diff
@@ -0,0 +1,91 @@
+import path from 'node:path'
```
sorccu: I don't quite get what this test is for. It doesn't use any code from the repo other than the fixture. Was this used for experimenting or what? If there's no particular reason, I think I will delete it.

sorccu (approved):

PR #1293, opened 2026-05-06
fix: adjustments on agentic check construct in prep for self-service billing rollout

thebiglabasky:
Removes the location and frequency enforcements, and adjust tests and fixtures accordingly.

It is acknowledged that people on the AGENTIC_CHECKS provisioned flag will be able to use the extended permissions of the adjusted construct (ie 1 region -> 3 regions allowed, 30 mins freq -> 5 mins freq): it is fine.


packages/cli/src/constructs/agentic-check.ts
```diff
@@ -273,47 +183,18 @@ export class AgenticCheck extends Check {
...
-      // Always emit `agentRuntime` so the backend has an explicit, complete
-      // picture of the runtime surface the user wants. The CLI is the source
-      // of truth: omitted skills/env vars mean "the agent should not have
-      // them", not "preserve whatever was there before".
+      // Always emit `agentRuntime` so the backend has an explicit picture of
+      // the skills the user wants. Omitted skills mean "the agent should not
+      // have them", not "preserve whatever was there before".
       agentRuntime: {
         skills: this.agentRuntime?.skills ?? [],
-        exposeEnvironmentVariables: this.agentRuntime?.exposeEnvironmentVariables ?? [],
```
MichaelHogers, line 316: breaking change, just a heads up, will need updating of our internal agentic check suite repo  soon

MichaelHogers (approved):

packages/cli/src/constructs/agentic-check.ts
thebiglabasky: Yes 😬 
I would prefer shipping a breaking change now while it's still alpha / beta so acceptable, rather than dragging debt behind

thebiglabasky: Good news is that it's versioned so we can pin on prev version while we upgrade and adopt at our own cadence

PR #1295, opened 2026-05-08
feat(cli): add account members command

thebiglabasky:
## Summary

- add `checkly account members` as a read-only command backed by `/v1/accounts/me/members`
- render account members and invites in table/markdown, with raw JSON output preserved
- add endpoint-backed filters: `--search`, `--type`, `--role`, `--status`, `--limit`, and `--next-id`
- normalize `--type`, `--role`, and `--status` values case-insensitively before sending canonical API values
- document the command in the embedded Checkly skill context and regenerate the public skill output

## Validation

- `npx vitest --run src/commands/__tests__/account-members.spec.ts src/formatters/__tests__/account-members.spec.ts src/commands/__tests__/command-metadata.spec.ts`
- `npm run lint`
- `npx tsc -p tsconfig.json --noEmit --skipLibCheck`
- `npm run prepare:ai-context --workspace packages/cli && npm run prepare:ai-context`

Note: full `npm run prepare:dist --workspace packages/cli` was attempted earlier and is currently blocked by existing `lru-cache` declaration errors under `node_modules`.


MichaelHogers:
@thebiglabasky high level thinking, but I think it is worth considering now rather than later:

considering https://github.com/checkly/monorepo/pull/1730 the rich openapi schema here, what about pulling in https://api.checklyhq.com/swagger.json and using https://heyapi.dev/ to create a SDK for our CLI? we can filter the openapi spec down to only the routes we use

thebiglabasky:
> @thebiglabasky high level thinking, but I think it is worth considering now rather than later:
> 
> considering [checkly/monorepo#1730](https://github.com/checkly/monorepo/pull/1730) the rich openapi schema here, what about pulling in https://api.checklyhq.com/swagger.json and using https://heyapi.dev/ to create a SDK for our CLI? we can filter the openapi spec down to only the routes we use

I'd like to tackle the SDK separately, mostly because of the public / private segmentation of our repos which would require publishing on npm and that felt like too much. But I agree that'd be something to do now-ish, not 6+ commands down the road.

Also noted yesterday that we have an issue with the zod API that doesn't generate proper OpenAPI spec and have a draft PR to fix that which I need to verify since I haven't looked enough yet into how it's currently done: https://github.com/checkly/monorepo/pull/1752

MichaelHogers:
makes sense, maybe there's a shortcut in using https://api.checklyhq.com/swagger.json directly 🤔 alright then

MichaelHogers (approved):
👍 looks good

PR #1300, opened 2026-05-15
ci: test self-hosted windows arm64 runner

danielpaulus:
## Summary
- Route the Windows test matrix entry to a self-hosted ARM64 runner (`[self-hosted, Windows, ARM64, checkly-cli]`) to validate the toolchain end-to-end.
- `lint` and `test-e2e` jobs untouched (still on GH-hosted runners).

## Test plan
- [ ] `test - ubuntu-latest` still passes on GH-hosted
- [ ] `test - windows-self-hosted-arm64` runs on the self-hosted runner and passes

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -24,9 +24,13 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, windows-latest-x64]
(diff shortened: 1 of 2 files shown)


sorccu:
I'll close this as unnecessary - Windows is in pretty decent shape now, so we don't need the extra maintenance load from our own actions runner.

PR #1303, opened 2026-05-19
Add RCA support for test session error groups

thebiglabasky:
## What changed

Adds `checkly rca run --test-session-error-group <id>` support alongside the existing `--error-group` flow.

- Adds a REST client for test session error groups.
- Calls `POST /v1/root-cause-analyses/test-session-error-groups/{testSessionErrorGroupId}` for the new source type.
- Keeps the existing error-group path intact.
- Uses a generic `normalizeFlagAliases()` helper for the multi-character `-te` alias instead of command-local argv rewriting.
- Updates pending RCA output for terminal, Markdown, and JSON, including aligned terminal labels for the longer test-session label.
- Adds formatter coverage plus flag-alias helper coverage.

## Release-transition import note

Simo mentioned that new local imports should use `.js` extensions before the next major release. New or modified local imports in this PR now use `.js` extensions, including `./test-session-error-groups.js`, `../../helpers/flags.js`, and the new helper test import.

## Validation

- `npx eslint packages/cli/src/commands/rca/run.ts packages/cli/src/formatters/rca.ts packages/cli/src/formatters/__tests__/rca.spec.ts packages/cli/src/rest/api.ts packages/cli/src/rest/rca.ts packages/cli/src/rest/test-session-error-groups.ts packages/cli/src/helpers/flags.ts packages/cli/src/helpers/__tests__/flags.spec.ts`
- `npm exec --workspace packages/cli -- vitest --run src/formatters/__tests__/rca.spec.ts src/helpers/__tests__/flags.spec.ts`
- `npm run prepack --workspace packages/cli`

`npm run prepare:dist --workspace packages/cli` currently fails on an unrelated existing issue in `src/auth/index.ts`: `jwt-decode` has no exported member `jwtDecode` in the installed dependency/type shape.

### DIFF
--- a/packages/cli/src/commands/rca/run.ts
+++ b/packages/cli/src/commands/rca/run.ts
@@ -1,7 +1,7 @@
 import { Flags } from '@oclif/core'
 import chalk from 'chalk'
 import { AuthCommand } from '../authCommand'
-import { outputFlag } from '../../helpers/flags'
+import { normalizeFlagAliases, outputFlag } from '../../helpers/flags.js'
 import * as api from '../../rest/api'
 import { NotFoundError, InadequateEntitlementsError } from '../../rest/errors'
 import { formatRcaPending, formatRcaCompleted } from '../../formatters/rca'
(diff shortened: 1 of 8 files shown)


packages/cli/src/commands/rca/run.ts
```diff
@@ -1,7 +1,7 @@
 import { Flags } from '@oclif/core'
 import chalk from 'chalk'
 import { AuthCommand } from '../authCommand'
-import { outputFlag } from '../../helpers/flags'
+import { normalizeFlagAliases, outputFlag } from '../../helpers/flags.js'
```
MichaelHogers, line 4: worth noting the js imports ? in an agents.md

MichaelHogers (approved):

packages/cli/src/commands/rca/run.ts
thebiglabasky: That's been an ask from @sorccu until we make the big major upgrade. That's how I understood it

thebiglabasky: So I don't think this will stay like this for long

MichaelHogers: 👍

PR #1304, opened 2026-05-20
feat: adds Slack App construct

tnolet:
## Affected Components
* [x] CLI
* [ ] Create CLI
* [ ] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer

Adds a new `SlackAppAlertChannel` standalone alert channel construct, following the same pattern as `SlackAlertChannel`.

- `SlackAppAlertChannel` construct with `slackChannels: string[]` config, synthesizing as type `SLACK_APP`
- `SlackAppAlertChannelCodegen` for code generation from imported resources
- Registered in `AlertChannelCodegen` type union and codegen map
- AI context updated with fixture, example config, and reference docs


thebiglabasky:
Just ran a review agent real quick, but didn't yet take the time to review myself, so posting more as an FYI than as a conviction. Usually these findings are relevant. I'll review this afternoon.

AI-generated comments below:

Findings

Critical: [packages/cli/src/constructs/slack-app-alert-channel-codegen.ts:12](https://github.com/checkly/checkly-cli/blob/c837fc371b51a35cbda0c641d14abac0096a561d/packages/cli/src/constructs/slack-app-alert-channel-codegen.ts#L12) uses SlackAlertChannel as the generated construct for SLACK_APP. The new class is SlackAppAlertChannel, but import/codegen instantiates the legacy Slack webhook construct while passing slackChannels. I confirmed the runtime effect: new SlackAlertChannel(..., { slackChannels: ['#ops'] }).synthesize() returns {"type":"SLACK","config":{}}, so JS imports silently become the wrong channel type, and TS imports fail because SlackAlertChannelProps requires url. This breaks import plan output for every Slack App alert channel. Change the codegen construct to SlackAppAlertChannel.

Important: [packages/cli/src/constructs/slack-app-alert-channel.ts:4](https://github.com/checkly/checkly-cli/blob/c837fc371b51a35cbda0c641d14abac0096a561d/packages/cli/src/constructs/slack-app-alert-channel.ts#L4) makes slackChannels optional, so new SlackAppAlertChannel('x', {}) type-checks and synthesizes {"type":"SLACK_APP","config":{}}. The backend alert-channel schemas require at least one #channel or @handle, and the delivery path expects config.slackChannels.length. The construct should make slackChannels required at minimum, and ideally validate non-empty entries with the same #/@ prefix rule before deploy.

Verification

I checked the PR diff against the GitHub base/head SHAs and inspected the surrounding alert-channel/codegen patterns. I ran npm ci, npm run prepare:dist --workspace packages/cli, and npm exec --workspace packages/cli -- vitest --run src/constructs/__tests__/alert-channel.spec.ts; the build and targeted existing tests passed. I did not post any GitHub review comments.

tnolet:
> Just ran a review agent real quick, but didn't yet take the time to review myself, so posting more as an FYI than as a conviction. Usually these findings are relevant. I'll review this afternoon.
> 
> AI-generated comments below:
> 
> Findings
> 
> Critical: [packages/cli/src/constructs/slack-app-alert-channel-codegen.ts:12](https://github.com/checkly/checkly-cli/blob/c837fc371b51a35cbda0c641d14abac0096a561d/packages/cli/src/constructs/slack-app-alert-channel-codegen.ts#L12) uses SlackAlertChannel as the generated construct for SLACK_APP. The new class is SlackAppAlertChannel, but import/codegen instantiates the legacy Slack webhook construct while passing slackChannels. I confirmed the runtime effect: new SlackAlertChannel(..., { slackChannels: ['#ops'] }).synthesize() returns {"type":"SLACK","config":{}}, so JS imports silently become the wrong channel type, and TS imports fail because SlackAlertChannelProps requires url. This breaks import plan output for every Slack App alert channel. Change the codegen construct to SlackAppAlertChannel.
> 
> Important: [packages/cli/src/constructs/slack-app-alert-channel.ts:4](https://github.com/checkly/checkly-cli/blob/c837fc371b51a35cbda0c641d14abac0096a561d/packages/cli/src/constructs/slack-app-alert-channel.ts#L4) makes slackChannels optional, so new SlackAppAlertChannel('x', {}) type-checks and synthesizes {"type":"SLACK_APP","config":{}}. The backend alert-channel schemas require at least one #channel or @handle, and the delivery path expects config.slackChannels.length. The construct should make slackChannels required at minimum, and ideally validate non-empty entries with the same #/@ prefix rule before deploy.
> 
> Verification
> 
> I checked the PR diff against the GitHub base/head SHAs and inspected the surrounding alert-channel/codegen patterns. I ran npm ci, npm run prepare:dist --workspace packages/cli, and npm exec --workspace packages/cli -- vitest --run src/constructs/**tests**/alert-channel.spec.ts; the build and targeted existing tests passed. I did not post any GitHub review comments.

@thebiglabasky why are you reviewing this?

tnolet:
Fixed both AI comments.

thebiglabasky:
> @thebiglabasky why are you reviewing this?

I've worked a fair bit on CLI and recently did the constructs and codegen for agentic check, so thought the more the merrier. Happy to drop and do something else though 🤷🏻

tnolet:
> > @thebiglabasky why are you reviewing this?
> 
> I've worked a fair bit on CLI and recently did the constructs and codegen for agentic check, so thought the more the merrier. Happy to drop and do something else though 🤷🏻

ah, ok. Thank for the offer but there is no need to distract you with this.

sorccu:
## Rebased onto main (post-v8.0.0)

Rebased the 3 commits onto current main, which includes the v8.0.0 ESM migration. Changes made during rebase:

- Resolved conflicts in `alert-channel-codegen.ts` and `index.ts` — took main's ESM imports, added Slack App entries with `.js` suffixes
- Updated imports in both new files (`slack-app-alert-channel.ts`, `slack-app-alert-channel-codegen.ts`) to ESM style: `.js` suffixes on all local imports, `index.js` for barrel re-exports from `./internal/codegen` and `../sourcegen`

TypeScript compiles cleanly after the rebase.

### Observations

Two minor things worth being aware of in the codegen:

1. **Fallback name when `slackChannels` is undefined:** In `prepare()`, if `slackChannels` is undefined the fallback name becomes `"slack-app-undefined"`. Unlikely in practice since the API should always provide channels for a `SLACK_APP` type, but worth noting.

2. **`slackChannels?.concat().slice(-8)` operates on array elements**, not string characters — it takes the last 8 channel names from the array. This differs from the existing Slack codegen's `url.slice(-4)` which takes the last 4 characters of a string. The resulting fallback name may be quite long if the channel names are long.

packages/cli/src/constructs/slack-app-alert-channel-codegen.ts
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    return 'Slack App Alert Channel'
+  }
+
+  prepare (logicalId: string, resource: SlackAppAlertChannelResource, context: Context): void {
+    const { slackChannels } = resource.config
+
+    const last8Chars = slackChannels?.concat().slice(-8)
+    const fallbackName = `slack-app-${last8Chars}`
```
sorccu: I think you may have intended to use `[].join()` rather than `[].concat()` here. The latter has barely any effect - at best it will produce a copy of  the original `slackChannels` array which has no purpose in this context.

That being said we probably don't need to do this dance here. The original slack alert channel only does it if it only has access to the url without a channel. Otherwise it uses the full channel name. Also, it seems to me like `slackChannels` should not be optional at all, but I'll make a separate note of that.

packages/cli/src/constructs/slack-app-alert-channel-codegen.ts
```diff
@@ -0,0 +1,67 @@
+import { Codegen, Context } from './internal/codegen/index.js'
+import { decl, expr, ident } from '../sourcegen/index.js'
+import { buildAlertChannelProps, AlertChannelResource } from './alert-channel-codegen.js'
+
+export interface SlackAppAlertChannelResource extends AlertChannelResource {
+  type: 'SLACK_APP'
+  config: {
+    slackChannels?: string[]
```
sorccu: Is there a valid situation where `slackChannels` of an existing `SLACK_APP` alert channel would be undefined? The construct itself has it as a required property. If we can make it required, it would potentially simplify the codegen a bit.

packages/cli/src/constructs/slack-app-alert-channel-codegen.ts
```diff
@@ -0,0 +1,67 @@
...
+        builder.new(builder => {
+          builder.string(logicalId)
+          builder.object(builder => {
+            if (config.slackChannels?.length) {
+              builder.array('slackChannels', arrayBuilder => {
+                for (const channel of config.slackChannels!) {
+                  arrayBuilder.string(channel)
+                }
+              })
+            }
```
sorccu: Given how `slackChannels` is a required property in the construct, I think we have to at least produce an empty array here. So I'd probably do this instead:


```suggestion
            builder.array('slackChannels', arrayBuilder => {
              for (const channel of config.slackChannels ?? []) {
                arrayBuilder.string(channel)
              }
            })
```

(And `?? []` can be removed if we make `slackChannels` required)

packages/cli/src/constructs/slack-app-alert-channel-codegen.ts
tnolet: it should indeed be required. Empty value should not be accepted.

packages/cli/src/constructs/slack-app-alert-channel-codegen.ts
tnolet: lgtm

packages/cli/src/constructs/slack-app-alert-channel-codegen.ts
tnolet: correct, but the channel here has no name, so I'm not sure how to make the `slack-app-xx` unique? But maybe I misunderstand what this does

sorccu (approved):

PR #1305, opened 2026-05-20
feat(cli): guide users to docs and skills on parse errors

stefanjudis:
## Affected Components
* [x] CLI

## Notes for the Reviewer

Improves error output across the CLI so users discover `checkly skills`, the docs, and `checkly help` from any misuse — not just from a small subset of failures. Two changes on this branch:

1. **Unknown commands** — replaced `@oclif/plugin-not-found` with a local `command_not_found` hook that points to the docs and `checkly skills`.
2. **Oclif parse errors** — new `finally` hook that appends the same hint to unknown-flag, invalid-flag-value, and missing-required-arg errors.

🔔 This PR also removes the previously used `@oclif/plugin-not-found` because it wasn't possible to tweak its output nicely. The code change was minimal. The only thing that got lost is the previous `Did you mean XXX?` prompt on unknown commands. I don't think it's a big deal, though.

### How the parse-error hook works

- New `finally` hook at `packages/cli/src/hooks/parse-error.ts` registered in `package.json`.
- All seven oclif parser error subclasses extend `CLIParseError`, which carries a `parse` field. The hook sniffs for that to detect parse errors without needing to import the unexported class.
- It mutates `err.message` *before* oclif's top-level `handle()` prints, so the hint shows up inside the boxed error block, right after oclif's `See more help with --help` line. No second writer to stderr.
- App-level errors (auth, network, validation we throw ourselves) don't carry `parse`, so they remain untouched.

#### Unknown command — `npx checkly does-not-exist`

Before:
```
 ›   Warning: does-not-exist is not a checkly command.
 ›   Error: Run checkly help for a list of available commands.
```

After:
```
 ›   Error: does-not-exist is not a checkly command.

Looking for help?

CLI help:         run checkly help
Agent workflows:  run checkly skills
Documentation:    https://www.checklyhq.com/docs/cli/
```

#### Unknown flag — `npx checkly checks list --does-not-exist`

Before:
```
 ›   Error: Nonexistent flag: --does-not-exist
 ›   See more help with --help

USAGE
  $ checkly checks list ...
```

After:
```
 ›   Error: Nonexistent flag: --does-not-exist
 ›   See more help with --help
 ›   Need agent workflows or docs? Run checkly skills.

USAGE
  $ checkly checks list ...
```

#### Invalid flag value — `npx checkly checks list --output bogus`

Before:
```
 ›   Error: Expected --output=bogus to be one of: table, json, md
 ›   See more help with --help
```

After:
```
 ›   Error: Expected --output=bogus to be one of: table, json, md
 ›   See more help with --help
 ›   Need agent workflows or docs? Run checkly skills.
```

#### Missing required arg — `npx checkly checks get`

Before:
```
 ›   Error: Missing 1 required arg:
 ›   id  The ID of the check to retrieve.
 ›   See more help with --help
```

After:
```
 ›   Error: Missing 1 required arg:
 ›   id  The ID of the check to retrieve.
 ›   See more help with --help
 ›   Need agent workflows or docs? Run checkly skills.
```

### Tests

- New e2e spec `packages/cli/e2e/__tests__/cli-parse-error.spec.ts` covering the unknown-flag and invalid-flag-value paths.
- Existing `packages/cli/e2e/__tests__/command-not-found.spec.ts` still passes.

### DIFF
--- a/packages/cli/e2e/__tests__/command-not-found.spec.ts
+++ b/packages/cli/e2e/__tests__/command-not-found.spec.ts
@@ -1,0 +1,17 @@
+import { describe, it, expect } from 'vitest'
+
+import { runChecklyCli } from '../run-checkly'
+
+describe('command-not-found', () => {
+  it('points users to the docs and skills command for unknown commands', async () => {
+    const { status, stderr } = await runChecklyCli({
+      args: ['does-not-exist'],
+    })
+
+    expect(status).toBe(127)
+    expect(stderr).toContain('does-not-exist is not a checkly command')
+    expect(stderr).toContain('https://www.checklyhq.com/docs/cli/')
+    expect(stderr).toContain('checkly skills')
+    expect(stderr).toContain('checkly help')
+  })
+})
--- a/package-lock.json
+++ b/package-lock.json
@@ -1983,6 +1983,7 @@
       "version": "1.0.2",
       "resolved": "https://registry.npmjs.org/@inquirer/ansi/-/ansi-1.0.2.tgz",
       "integrity": "sha512-S8qNSZiYzFd0wAcyG5AXCvUHC5Sr7xpZ9wZ2py9XR88jUz8wooStVx5M6dRzczbBWjic9NP7+rY0Xi7qqK/aMQ==",
+      "dev": true,
       "license": "MIT",
       "engines": {
         "node": ">=18"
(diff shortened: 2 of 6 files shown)


packages/cli/e2e/__tests__/command-not-found.spec.ts
thebiglabasky, line 1: Nit: shouldn't we merge this with the cli-parse-error somehow? Yes they're different but also so related that having different suites feels a bit too much?

thebiglabasky (approved):
A nit that's far from blocking, I think this is good 🙌🏻

PR #1320, opened 2026-05-25
feat(cli): add `checkly api` generic API pass-through command [ship]

martzoukos:
## Why

The Checkly CLI has dedicated commands for a handful of API endpoints, but the Checkly API has ~150 endpoints total. Today, users and AI agents that need to reach an uncovered endpoint have to fall back to `curl` with manually constructed auth headers — which is error-prone and breaks agent workflows.

This is especially important for AI agents: when an agent needs data from an endpoint that doesn't have a CLI command yet, it's stuck. This command gives agents (and power users) an immediate escape hatch — authenticated `curl` for the entire Checkly API, similar to how `gh api` works for GitHub.

**Linear:** AI-132

### New command: `checkly api <endpoint>`

A generic pass-through that makes authenticated HTTP requests to any Checkly API endpoint. The CLI injects auth headers and the correct base URL automatically.

```bash
checkly api /v1/checks
checkly api /v1/dashboards -X GET --jq '.[].name'
checkly api /v1/checks -X POST -F name=MyCheck -F activated:=true
```

**Flags:**
- `-X` / `--method` — HTTP method (GET, POST, PUT, PATCH, DELETE). Defaults to GET, or POST when fields are present
- `-f` — String field parameter (`key=value`)
- `-F` — Typed field parameter (`key=value` for strings, `key:=value` for JSON like booleans/numbers/arrays)
- `-H` — Custom HTTP header
- `--jq` — Filter JSON output using system `jq` (clear error message if `jq` isn't installed)
- `--input` — Request body from a file path or `-` for stdin
- `--paginate` — Auto-fetch all pages (supports both Content-Range and cursor-based pagination)
- `--include` / `-i` — Show response status and headers
- `--verbose` — Print full request/response headers to stderr

**Output:** Pretty-printed JSON in the terminal, compact JSON when piped (matches `gh api` behavior).

### AI context / skills update

Updated the Checkly skill file (`SKILL.md`) so AI agents know `checkly api` exists as a fallback. The section points to the [Checkly API reference](https://www.checklyhq.com/docs/api) for endpoint discovery — no static endpoint list baked in, so it won't go stale when the backend API changes.

### Security

- **URL validation** — Rejects absolute URLs (`https://evil.com/...`) and protocol-relative URLs (`//evil.com/...`) to prevent leaking auth headers to external hosts
- **No shell injection** — The `--jq` flag uses `execFile` (not `exec`), so the filter expression is passed as a direct argument to the `jq` binary with no shell interpretation
- **404 docs hint** — On a 404 response, stderr shows a link to the API docs to help users find the right endpoint

## New files

| File | Purpose |
|------|---------|
| `packages/cli/src/commands/api.ts` | The command |
| `packages/cli/src/helpers/api-fields.ts` | Parses `-f` / `-F` field flags |
| `packages/cli/src/helpers/api-paginate.ts` | Auto-pagination (Content-Range + cursor) |
| `packages/cli/src/commands/__tests__/api.spec.ts` | Command unit tests (19 tests) |
| `packages/cli/src/helpers/__tests__/api-fields.spec.ts` | Field parsing tests (11 tests) |
| `packages/cli/src/helpers/__tests__/api-paginate.spec.ts` | Pagination tests (5 tests) |

## Manual testing

Requires `CHECKLY_API_KEY` and `CHECKLY_ACCOUNT_ID` set, or a prior `checkly login`.

```bash
# Basic GET
./packages/cli/bin/run api /v1/checks

# Limit results
./packages/cli/bin/run api /v1/checks -X GET -f limit=2

# Filter with jq
./packages/cli/bin/run api /v1/checks --jq '.[].name'

# Verbose headers
./packages/cli/bin/run api /v1/checks --verbose

# Paginate all checks
./packages/cli/bin/run api /v1/checks --paginate

# Cursor-based pagination (status pages)
./packages/cli/bin/run api /v1/status-pages --paginate

# Include response headers
./packages/cli/bin/run api /v1/checks -i

# Compact JSON when piped
./packages/cli/bin/run api /v1/checks | jq length

# 404 with docs hint
./packages/cli/bin/run api /v1/nonexistent

# Security: rejects absolute URLs
./packages/cli/bin/run api https://evil.com/steal
./packages/cli/bin/run api //evil.com/steal
```

## Test plan

- [x] Unit tests pass (141 total — 19 command + 11 field parsing + 5 pagination + 106 existing metadata)
- [x] ESLint passes
- [ ] Manual smoke test against live API with real credentials
- [ ] Verify `checkly api --help` output
- [ ] Verify updated skill via `./packages/cli/bin/run skills install`


packages/cli/src/ai-context/skill.md
```diff
@@ -36,4 +36,19 @@ Write commands (e.g. `incidents create`, `deploy`, `destroy`) return exit code 2
 
 Run `npx checkly skills communicate` for the full protocol details.
 
+## API Pass-Through (fallback for any endpoint)
```
martzoukos, line 39: @stefanjudis I'd appreciate your 2c on this section regarding agent ergonomics.

packages/cli/src/commands/__tests__/api.spec.ts
```diff
@@ -0,0 +1,318 @@
...
+      expect(api.request).toHaveBeenCalledWith(expect.objectContaining({ url: '/v1/checks' }))
+    })
+
+    it('prepends / if missing', async () => {
+      const cmd = createCommand('v1/checks')
+      await cmd.run()
+      expect(api.request).toHaveBeenCalledWith(expect.objectContaining({ url: '/v1/checks' }))
+    })
+
+    it('rejects protocol-relative URLs', async () => {
```
thebiglabasky, line 89: Nice

packages/cli/src/ai-context/skill.md
```diff
@@ -36,4 +36,19 @@ Write commands (e.g. `incidents create`, `deploy`, `destroy`) return exit code 2
...
+```bash
+npx checkly api /v1/checks
+npx checkly api /v1/dashboards -X GET --jq '.[].name'
+npx checkly api /v1/checks -X POST -F name=MyCheck -F activated:=true
+npx checkly api /v1/checks -X GET -f limit=5 --paginate
+```
+
+Key flags: `-X` (method), `-f` (string field), `-F` (typed/JSON field), `-H` (header), `--jq` (filter with jq), `--input` (body from file/stdin), `--paginate`, `--verbose`.
+
+See the [Checkly API reference](https://www.checklyhq.com/docs/api) for available endpoints.
```
thebiglabasky: I'm wondering if we shouldn't point agents at https://api.checklyhq.com/openapi.json instead of docs. Docs may be ever so slightly out of date, and longer to parse. The OpenAPI spec is big so consumes quite some tokens, but takes less back and forth. I guess it's a tradeoff, but might be worth giving the agent the option

packages/cli/src/commands/__tests__/api.spec.ts
```diff
@@ -0,0 +1,318 @@
...
+      const cmd = createCommand('/v1/checks', '--paginate')
+      await cmd.run()
+      const loggedJson = vi.mocked(cmd.log).mock.calls[0][0]
+      const parsed = JSON.parse(loggedJson)
+      expect(parsed).toHaveLength(4)
+    })
+  })
+
+  describe('--jq', () => {
+    it('pipes output through system jq', async () => {
```
thebiglabasky, line 254: That won't work on Windows though?

packages/cli/src/commands/api.ts
```diff
@@ -0,0 +1,241 @@
...
+import { api } from '../rest/api.js'
+import { parseFields } from '../helpers/api-fields.js'
+import { paginateAll } from '../helpers/api-paginate.js'
+
+const WRITE_METHODS = new Set(['POST', 'PUT', 'PATCH'])
+
+export default class Api extends AuthCommand {
+  static hidden = false
+  static readOnly = false
+  static destructive = false
```
thebiglabasky: Hm. Could be destructive if you PUT / PATCH stuff with jibberish, so might be worth flagging this as `true`
```suggestion
  static destructive = true
```

packages/cli/src/commands/api.ts
```diff
@@ -0,0 +1,241 @@
...
+  static idempotent = false
+  static description = 'Make an authenticated HTTP request to the Checkly API.\n'
+    + 'Pass-through for any endpoint — handles auth automatically.\n'
+    + 'See https://www.checklyhq.com/docs/api for available endpoints.'
```
thebiglabasky: Same note about openapi.json as a replacement or complement

packages/cli/src/commands/api.ts
```diff
@@ -0,0 +1,241 @@
...
+      default: [],
+    }),
+    'header': Flags.string({
+      char: 'H',
+      description: 'Custom HTTP header: "Key: Value".',
+      multiple: true,
+      default: [],
+    }),
+    'jq': Flags.string({
+      description: 'Filter JSON output with a jq expression (requires jq installed).',
```
thebiglabasky, line 55: Would this be worth considering adding `node-jq` as a dependency?
Or we start without and see how much of a problem that is?
What does `gh` do?

packages/cli/src/commands/api.ts
```diff
@@ -0,0 +1,241 @@
...
+    }),
+  }
+
+  async run (): Promise<void> {
+    const { args, flags } = await this.parse(Api)
+
+    const hasFields = flags.field.length > 0 || flags['raw-field'].length > 0
+    const method = (flags.method ?? (hasFields || flags.input ? 'POST' : 'GET')).toUpperCase()
+
+    if (args.endpoint.startsWith('//') || /^[a-z][a-z\d+\-.]*:/i.test(args.endpoint)) {
```
thebiglabasky, line 77: Any other way hackers could try eg using some encoded chars or so?

packages/cli/src/commands/api.ts
```diff
@@ -0,0 +1,241 @@
...
+      description: 'Filter JSON output with a jq expression (requires jq installed).',
+    }),
+    'input': Flags.string({
+      description: 'Request body from file path, or "-" for stdin.',
+    }),
+    'paginate': Flags.boolean({
+      description: 'Fetch all pages of results automatically.',
+      default: false,
+    }),
+    'include': Flags.boolean({
```
thebiglabasky, line 60: I'm a bit confused by that name. If that's how `gh` does it, fine, let's keep that, otherwise it feels a bit overlapping with `--verbose` and I'd consider only keeping --verbose.

packages/cli/src/helpers/api-paginate.ts
```diff
@@ -0,0 +1,110 @@
...
+  warn (message: string): void
+}
+
+export function paginateAll (
+  apiClient: AxiosInstance,
+  requestConfig: AxiosRequestConfig,
+  firstResponse: AxiosResponse,
+  logger?: PaginateLogger,
+): Promise<unknown> {
+  const contentRange = firstResponse.headers['content-range']
```
thebiglabasky: Is this a standard header all of page/limit paginated APIs have? I wasn't aware. Just to be sure we wouldn't have one of them missing the header and fall through the rest of the implementation, and then fail.

packages/cli/src/helpers/api-paginate.ts
```diff
@@ -0,0 +1,110 @@
...
+  if (contentRange) {
+    return paginateContentRange(apiClient, requestConfig, firstResponse, contentRange)
+  }
+
+  if (
+    firstResponse.data
+    && typeof firstResponse.data === 'object'
+    && 'nextId' in firstResponse.data
+    && 'entries' in firstResponse.data
+    && Array.isArray(firstResponse.data.entries)
```
thebiglabasky: Shouldn't we condition on the presence of `nextId` instead? data.entries could theoretically work equally well with a page/limit approach?

packages/cli/src/helpers/content-range.ts
thebiglabasky, line 1: I was about to propose to move this into api-paginate, but see it's used by the REST client, I guess it's fine. Looks like a tiny file now :)

thebiglabasky (commented):
I'd love to discuss a few points before approving. Nothing really blocking, but also, not yet confident we want to go exactly as is

packages/cli/src/ai-context/skill.md
martzoukos: Yeah, I was looking for something like that, I'll adopt this url.

packages/cli/src/commands/api.ts
martzoukos: `gh api` ships without it, so let's start without it for now.

packages/cli/src/commands/__tests__/api.spec.ts
martzoukos: Correct. But we return a useful message if it's not found:
> --jq requires jq to be installed.
> Install it from https://jqlang.github.io/jq/

packages/cli/src/commands/api.ts
martzoukos: There's surprisingly not a big risk here since it's dumb string concatenation so an evil url would be converted to: `https://api.checklyhq.com/%2F%2Fevil.com` which is unintentionally safe 😅 

So, yeah, I'll do some proper url resolution and detection:
```js
  const baseUrl = api.defaults.baseURL // e.g. https://api.checklyhq.com
  const resolved = new URL(endpoint, baseUrl)
  if (resolved.origin !== new URL(baseUrl).origin) {
    this.error('Endpoint must be a relative path (e.g. /v1/checks), not a URL.')
  }
```

packages/cli/src/commands/api.ts
martzoukos: It is a behavior picked up from `gh`.

packages/cli/src/helpers/api-paginate.ts
martzoukos: Yeah, good catch!

packages/cli/src/helpers/api-paginate.ts
martzoukos: It is in Hapi, just verified it with Claude.

martzoukos:
Great feedback @thebiglabasky !
I've addressed the following:
- **`destructive = true`** — Flipped the flag since the command can PUT/PATCH/DELETE resources
- **Hardened URL validation** — Added hostname verification after URL resolution (not just regex on the raw input). Even if someone encodes characters or uses backslashes, the resolved URL's origin must match the API base URL. Added a test for the backslash bypass case
- **Fixed cursor pagination detection** — Now keys on `nextId` alone instead of requiring both `nextId` and `entries`. Dynamically finds whichever array field is in the response (`entries`, `items`, etc.). This fixes a real bug where `--paginate` silently failed for `/v1/accounts/{accountId}/members` which returns `items` instead of `entries`
- **Added OpenAPI spec URL** — Added `api.checklyhq.com/openapi.json` alongside the docs link in both the command description and the AI skill file, so agents can use the machine-readable spec for endpoint discovery

sorccu:
Can you show me how you'd do something more complex such as setting a retry strategy for a check or group?

martzoukos:
> Can you show me how you'd do something more complex such as setting a retry strategy for a check or group?

By doing something like:
```bash
checkly api /v1/checks/YOUR_CHECK_ID -X PUT --input - <<'EOF'
{
"retryStrategy": {
    "type": "LINEAR",
    "baseBackoffSeconds": 10,
    "maxRetries": 3,
    "maxDurationSeconds": 600,
    "sameRegion": true
}
}
EOF
```
Are you asking about multi-line parameters in general or about something specific to retries that I'm missing?

thebiglabasky (dismissed):
The fact this mimics `gh` makes me pretty confident in the design. I didn't test it locally, but trust that you did. Fine work!

martzoukos:
@sorccu anything regarding your question above that needs addressing or shall I go ahead and merge it?

sorccu:
Can we delay merging till at least tomorrow? I'd like to think about this a bit. To be clear I'm for the idea, but I'm not sure if our API is ready to be exposed like this

martzoukos:
Yeah, that works @sorccu . 👍

MegaMaddin (dismissed):
Please check internal discussion in #cli.

packages/cli/src/commands/api.ts
```diff
@@ -0,0 +1,250 @@
...
+      description: 'Add a field: key=value (string) or key:=value (parsed as JSON).',
+      multiple: true,
+      default: [],
+    }),
+    'raw-field': Flags.string({
+      char: 'f',
+      description: 'String parameter: key=value.',
+      multiple: true,
+      default: [],
+    }),
```
sorccu: Why do we need `-f` if `-F` is capable of the exact same thing with the `=` operator?

packages/cli/src/commands/api.ts
```diff
@@ -0,0 +1,250 @@
...
+    'jq': Flags.string({
+      description: 'Filter JSON output with a jq expression (requires jq installed).',
+    }),
+    'input': Flags.string({
+      description: 'Request body from file path, or "-" for stdin.',
+    }),
+    'paginate': Flags.boolean({
+      description: 'Fetch all pages of results automatically.',
+      default: false,
+    }),
```
sorccu: This description confuses me. So you are saying that if we do:

```
pnpm checkly api /v1/checks --paginate
```

... then that will in NOT paginate, but return all results?

It seems that this matches the gh cli, but I just find it very strange.

sorccu:
I added a couple comments. I'd at least like your thoughts on those, but they don't necessarily have to be changed if they're relatively standard as-is.

I also rebased the PR to latest main. You should take care of the failing CI.

Other than that this looks alright to me - it doesn't auto generate anything so it remains very flexible.

packages/cli/src/commands/api.ts
martzoukos: Hm, that's actually worse than bad UX. An agent might casually pass this flag thinking it'll get one page and accidentally load all dozens of checks in the context window. I'll deviate from `gh api` here and make things consistent with the rest of the CLI.

martzoukos: (so, I'll drop the --paginate flag)

packages/cli/src/commands/api.ts
martzoukos: We can also deviate from gh here. Let's drop it.

thebiglabasky (dismissed):

thebiglabasky (approved):

PR #1321, opened 2026-05-26
feat(cli): add test sessions get command

thebiglabasky:
## Summary

- Add `checkly test-sessions get <test-session-id>` as a read-only detail command.
- Add public REST client support for `GET /v1/test-sessions/{testSessionId}`.
- Render detail and Markdown session summaries, result tables, session links, and RCA hints from session/result `errorGroupIds`.
- Keep terminal result rows readable by putting human fields first and long IDs at the right edge.
- Add a capped `ERROR GROUP IDS` section with `--error-groups-limit`, plus singular hydrated drilldown via `--error-group <id>`.
- Keep default JSON output as the raw session API response.

Linear: https://linear.app/checklyhq/issue/AI-342/cli-add-checkly-test-sessions-get

## Validation

- `pnpm --filter checkly exec vitest --run src/commands/__tests__/test-sessions-get.spec.ts src/formatters/__tests__/test-sessions.spec.ts src/rest/__tests__/test-sessions.spec.ts src/commands/__tests__/command-metadata.spec.ts`
- `pnpm exec eslint packages/cli/src/commands/test-sessions/get.ts packages/cli/src/formatters/test-sessions.ts packages/cli/src/rest/test-sessions.ts packages/cli/src/commands/__tests__/test-sessions-get.spec.ts packages/cli/src/formatters/__tests__/test-sessions.spec.ts packages/cli/src/rest/__tests__/test-sessions.spec.ts packages/cli/src/commands/__tests__/command-metadata.spec.ts`
- `pnpm --filter checkly run prepare:dist`
- `pnpm --filter checkly run prepack`
- `packages/cli/bin/run test-sessions get --help`

### DIFF
--- a/packages/cli/src/commands/test-sessions/get.ts
+++ b/packages/cli/src/commands/test-sessions/get.ts
@@ -44,10 +44,10 @@ export default class TestSessionsGet extends AuthCommand {
       const { data: testSession } = await api.testSessions.get(args.id)
 
       if (flags['error-group']) {
-        const errorGroupIds = [
+        const errorGroupIds = [...new Set([
           ...(testSession.errorGroupIds ?? []),
           ...(testSession.results ?? []).flatMap(result => result.errorGroupIds ?? []),
-        ]
+        ])]
 
         if (!errorGroupIds.includes(flags['error-group'])) {
           const listCommand = `checkly test-sessions get ${args.id} `
--- a/packages/cli/src/commands/test-sessions/get.ts
+++ b/packages/cli/src/commands/test-sessions/get.ts
@@ -1,0 +1,91 @@
+import { Args, Flags } from '@oclif/core'
+import { AuthCommand } from '../authCommand.js'
+import { outputFlag } from '../../helpers/flags.js'
+import * as api from '../../rest/api.js'
+import {
+  formatTestSessionDetail,
+  formatTestSessionErrorGroupDetail,
+} from '../../formatters/test-sessions.js'
+import { type OutputFormat } from '../../formatters/render.js'
+
+export default class TestSessionsGet extends AuthCommand {
+  static hidden = false
+  static readOnly = true
+  static idempotent = true
+  static description = 'Get details of a recorded test session, including result error groups for RCA.'
+
+  static args = {
+    id: Args.string({
+      description: 'The ID of the test session to retrieve.',
+      required: true,
+    }),
+  }
+
+  static flags = {
+    'error-group': Flags.string({
+      description: 'Show details for a test session error group ID from this session.',
+    }),
+    'error-groups-limit': Flags.integer({
+      description: 'Number of error group IDs to show in the session summary.',
+      default: 5,
+    }),
+    'full-error': Flags.boolean({
+      description: 'Print the complete raw error when showing a test session error group.',
+      default: false,
+    }),
+    'output': outputFlag({ default: 'detail' }),
+  }
+
+  async run (): Promise<void> {
+    const { args, flags } = await this.parse(TestSessionsGet)
+    this.style.outputFormat = flags.output
+
+    try {
+      const { data: testSession } = await api.testSessions.get(args.id)
+
+      if (flags['error-group']) {
+        const errorGroupIds = [
+          ...(testSession.errorGroupIds ?? []),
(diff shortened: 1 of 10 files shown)


packages/cli/src/commands/test-sessions/get.ts
```diff
@@ -0,0 +1,91 @@
...
+    const { args, flags } = await this.parse(TestSessionsGet)
+    this.style.outputFormat = flags.output
+
+    try {
+      const { data: testSession } = await api.testSessions.get(args.id)
+
+      if (flags['error-group']) {
+        const errorGroupIds = [
+          ...(testSession.errorGroupIds ?? []),
+          ...(testSession.results ?? []).flatMap(result => result.errorGroupIds ?? []),
```
miliberlin: You are potentially adding duplicate errorGroupIds here.

packages/cli/src/formatters/test-sessions.ts
```diff
@@ -0,0 +1,333 @@
...
+      { header: 'Status', value: (result, fmt) => formatStatus(result.status, fmt) },
+      { header: 'Result Type', value: result => result.resultType ?? '-' },
+      { header: 'Run Location', value: result => result.runLocation ?? '-' },
+      { header: 'Error Groups', value: (result, fmt) => formatList(result.errorGroupIds, fmt) },
+    ]
+  }
+
+  const termWidth = process.stdout.columns || 120
+  const showCheckId = termWidth >= 160 && results.some(result => result.checkId)
+  const typeWidth = 14
```
miliberlin, line 130: I already see the hardcoded widths being a bit problematic on smaller screens. We might need to invest some more time (later) to find out how to output scalable tables nicely.

packages/cli/src/formatters/test-sessions.ts
```diff
@@ -0,0 +1,333 @@
...
+  fullError?: boolean
+  errorLines?: number
+}
+
+interface ErrorGroupReference {
+  id: string
+  sources: string[]
+}
+
+function formatStatus (status: string, format: OutputFormat): string {
```
miliberlin: Can you double-check the available statuses? I think half of these don't exist... We only have failed, cancelled, running, passed for test sessions.

miliberlin (changes requested):

packages/cli/src/formatters/test-sessions.ts
thebiglabasky: Yeah these formatting pieces have been like that across commands and I agree this would deserve a dedicated PR to adopt a systematic approach to move away from that 👍🏻

thebiglabasky: I'm throwing a session at this topic in parallel, let's see what I find. That'll be a separate PR though

thebiglabasky:
@miliberlin Addressed the comments, thank you 🙏🏻 
All yours to revisit

packages/cli/src/formatters/test-sessions.ts
```diff
@@ -0,0 +1,339 @@
...
+}
+
+function formatStatus (status: TestSessionStatus, format: OutputFormat): string {
+  const normalized = status.toLowerCase()
+  if (format === 'md') return normalized
+
+  switch (status) {
+    case 'FAILED':
+    case 'CANCELLED':
+      return chalk.red(normalized)
```
miliberlin, line 39: Can we make cancelled not red? Maybe grey or standard color.

packages/cli/src/commands/test-sessions/get.ts
```diff
@@ -0,0 +1,91 @@
...
+
+  async run (): Promise<void> {
+    const { args, flags } = await this.parse(TestSessionsGet)
+    this.style.outputFormat = flags.output
+
+    try {
+      const { data: testSession } = await api.testSessions.get(args.id)
+
+      if (flags['error-group']) {
+        const errorGroupIds = [...new Set([
```
miliberlin: Small nit: can we re-use `uniqueErrorGroupIds` here? Seems to do the same.

miliberlin (changes requested):

packages/cli/src/formatters/test-sessions.ts
thebiglabasky: Yeah, fair

miliberlin (approved):

PR #1323, opened 2026-05-27
feat(cli): add --check-id flag to trigger command [AI-218]

miliberlin:
## Summary

- Adds a `--check-id` flag to `checkly trigger` so users can target specific checks by ID. Accepts a comma-separated list and can be passed multiple times. AND-combines with `--tags` when both are provided.
- Plumbs `checkId` through `TriggerRunner` into the `POST /next/test-sessions/trigger` payload.
- Omits `checkId` from the payload when the array is empty, so existing trigger calls (no `--check-id` passed) keep matching all checks via the BE's `filterById` undefined bail-out instead of `whereIn('id', [])` matching zero rows.

Pairs with the backend change in checkly/monorepo#2005, which declares `checkId` on the `/next/test-sessions/trigger` Joi schema and wires it through `Check.modifiers.filterById`. Without that PR, the CLI request is rejected with `"checkId" is not allowed`.

## Test plan

- [ ] BE PR (checkly/monorepo#2005) merged and deployed before this lands.
- [ ] `checkly trigger --check-id <uuid>` runs only that check.
- [ ] `checkly trigger --check-id <uuid>,<uuid>` runs both checks.
- [ ] `checkly trigger --check-id <uuid> --check-id <uuid>` (repeatable form) runs both.
- [ ] `checkly trigger --check-id <uuid> --tags production` AND-combines (intersection only).
- [ ] `checkly trigger` (no `--check-id`) still matches all checks — regression check that omitting the field from the payload works.
- [ ] `checkly trigger --check-id <invalid-uuid>` returns a useful error from the BE.


thebiglabasky (approved):
I was originally thrown off by --check-id taking an array but that's fine given the comma-separated approach
🚀

miliberlin:
@thebiglabasky it's similar to how tags are handled `--tags production,webapp`

PR #1326, opened 2026-05-28
Add adaptive table formatter

thebiglabasky:
## Summary

- add `renderAdaptiveTable` so formatter columns can declare layout intent instead of computing terminal widths locally
- migrate `checks list` to the adaptive renderer as the first command example
- document formatter guidance to avoid reintroducing hard-coded table width math

## Validation

- `pnpm --filter checkly exec tsc --noEmit --pretty false`
- `pnpm --filter checkly exec eslint src/formatters/render.ts src/formatters/checks.ts src/formatters/__tests__/render.spec.ts src/formatters/__tests__/checks.spec.ts`
- `pnpm --filter checkly exec vitest --run ./src/formatters/__tests__/render.spec.ts ./src/formatters/__tests__/checks.spec.ts`

Note: formatter tests require the local `checkly-0.0.1-dev.tgz` fixture package, created with `npm pack --ignore-scripts --silent` and removed after the test run.

### DIFF
--- a/packages/cli/src/formatters/README.md
+++ b/packages/cli/src/formatters/README.md
@@ -1,0 +1,16 @@
+# CLI Formatters
+
+Prefer shared formatter helpers over command-local terminal layout math.
+
+For list-style terminal output, use `renderAdaptiveTable` when columns need to fit changing terminal widths or changing response content. Command formatters should describe column intent with `minWidth`, `maxWidth`, `width`, and `align`; the renderer should handle measuring, truncating, and padding.
+
+Avoid new formatter code that:
+
+- Reads `process.stdout.columns` directly.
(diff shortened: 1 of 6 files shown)


martzoukos (approved):
Claude mentioned that there is a bug in the implementation where in narrow widths the last column doesn't get trimmed and it flows in the second line. This, for example:

Image

To be honest, I don't agree with Claude here, the ID probably needs to be visible but maybe this indeed breaks the promise of this PR, so I'll let you decide if you also want to handle this case. 

Otherwise, lgtm.

thebiglabasky:
Yes I noticed it while testing but there's not much we can do past a point. I also believe the current behavior isn't technically better so I'll merge and we'll see about improving this further as we spot more cases

PR #1328, opened 2026-05-28
feat(cli): add test sessions list command

thebiglabasky:
## Merge condition

**Blocked by:** https://github.com/checkly/monorepo/pull/2043

This CLI PR **must not be merged or released** until the backend public `GET /v1/test-sessions` list endpoint from `checkly/monorepo#2043` is merged and ready for the CLI to call. The command added here depends on that public endpoint being available.

## Summary

- Add `checkly test-sessions list` backed by `GET /v1/test-sessions`.
- Support cursor pagination, ISO/Unix time filters, statuses, branches, users, providers, `--no-users`, text search, and test-session error group filtering.
- Add table, JSON, and Markdown output plus next-page and inspect-session hints.

## Validation

- `pnpm exec eslint packages/cli/src/commands/test-sessions/list.ts packages/cli/src/commands/__tests__/test-sessions-list.spec.ts packages/cli/src/formatters/test-sessions.ts packages/cli/src/rest/test-sessions.ts packages/cli/src/rest/__tests__/test-sessions.spec.ts packages/cli/src/commands/__tests__/command-metadata.spec.ts`
- Focused Vitest run with temporary config skipping the tarball global setup: `121 passed`
- `pnpm -C packages/cli run prepare:dist`
- `pnpm -C packages/cli run prepack`
- `packages/cli/bin/run test-sessions list --help`

## Notes

Full `pnpm pack` is still blocked locally by the existing `prepare:ai-context` issue where `./bin/run import plan` reports `import:plan is not a checkly command`.

### DIFF
--- a/packages/cli/src/commands/test-sessions/list.ts
+++ b/packages/cli/src/commands/test-sessions/list.ts
@@ -1,0 +1,219 @@
+import { Flags } from '@oclif/core'
+import { AuthCommand } from '../authCommand.js'
+import { outputFlag } from '../../helpers/flags.js'
+import * as api from '../../rest/api.js'
+import type { ListTestSessionsParams, TestSessionProvider, TestSessionStatus } from '../../rest/test-sessions.js'
+import type { OutputFormat } from '../../formatters/render.js'
+import {
+  formatTestSessionsList,
+  formatTestSessionsListNavigationHints,
+  formatTestSessionsListPaginationInfo,
+} from '../../formatters/test-sessions.js'
+
+const statusOptions = ['running', 'failed', 'passed', 'cancelled'] as const
+const providerOptions = ['github', 'vercel', 'api', 'trigger', 'pw_reporter'] as const
+
+function quoteArg (value: string): string {
+  if (/^[A-Za-z0-9._/@:+=,-]+$/.test(value)) return value
+  return `'${value.replace(/'/g, `'\\''`)}'`
+}
+
+function parseUnixTimestamp (value: string | undefined, label: string): number | undefined {
+  if (!value) return undefined
+
+  const trimmed = value.trim()
+  if (/^\d+$/.test(trimmed)) {
+    return Number(trimmed)
+  }
+
+  const timestamp = Date.parse(trimmed)
+  if (Number.isNaN(timestamp)) {
+    throw new Error(`Invalid --${label} "${value}". Use an ISO date or Unix timestamp in seconds.`)
+  }
+
+  return Math.floor(timestamp / 1000)
+}
+
+function normalizeStatus (value: string): TestSessionStatus | undefined {
+  const normalized = value.trim().toUpperCase()
+  return ['RUNNING', 'FAILED', 'PASSED', 'CANCELLED'].includes(normalized)
+    ? normalized as TestSessionStatus
+    : undefined
+}
+
+function normalizeProvider (value: string): TestSessionProvider | undefined {
+  const normalized = value.trim().replace(/-/g, '_').toUpperCase()
+  return ['GITHUB', 'VERCEL', 'API', 'TRIGGER', 'PW_REPORTER'].includes(normalized)
+    ? normalized as TestSessionProvider
+    : undefined
+}
+
+function buildListCommand (flags: Record<string, any>): string {
+  const parts = ['checkly test-sessions list', '--limit', String(flags.limit)]
+
+  const appendMany = (flag: string, values: string[] | undefined) => {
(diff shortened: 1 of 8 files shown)


packages/cli/src/commands/test-sessions/list.ts
```diff
@@ -0,0 +1,213 @@
...
+      noUsers: flags['no-users'] || undefined,
+      nextId: flags.cursor,
+      textSearch: flags.search,
+      errorGroupId: flags['error-group'],
+    }
+
+    try {
+      const { data } = await api.testSessions.list(params)
+
+      if (flags.output === 'json') {
```
miliberlin, line 177: There seems to be quite some inconsistency on response shapes from list commands. Let's make an intentional choice here. Here's the analysis and suggestion from Claude. I don't mind what format we pick, but we should be sticking to one going forward.


### `list` command consistency

| Command | Pagination | Cursor flag | `--limit` default | JSON envelope |
| --- | --- | --- | --- | --- |
| `checks list` | offset / page | `--page` | 25 | `{ data, pagination: { page, limit, total, totalPages } }` |
| `status-pages list` | cursor | `--cursor` | 25 | `{ data, pagination: { nextId, length } }` |
| `incidents list` | none exposed | — | 25 | `{ data, count, nextId? }` |
| `account members` | cursor (opt-in) | `--next-id` | none | raw → `{ members, length, nextId }` |
| **`test-sessions list`** | cursor | `--cursor` | 20 | raw → `{ length, entries, nextId }` |

`test-sessions list` mirrors `status-pages list` everywhere except the JSON branch, where it emits the raw API body instead of the reshaped `{ data, pagination }` form (used by 3 of 5 commands). `jq '.data'` returns `null` here. Suggest aligning to its sibling:

```ts
// test-sessions/list.ts
if (flags.output === 'json') {
  this.log(JSON.stringify({
    data: data.entries,
    pagination: { nextId: data.nextId, length: data.length },
  }, null, 2))
  return
}
```

Ripple: the `'returns raw API response for json output'` test asserts `toEqual(listResponse)` and needs the new envelope.

thebiglabasky: Yes, the pagination issue is real, but unfortunately something we inherited from a switch in approaches. My understanding is that we now use cursor-based pagination and we can't introduce a breaking change to endpoints easily. We could decide to harmonize on cursor-based and introduce a new version of the endpoints which still use limit / offset though.

Now, there is still an inconsistency in cursor-based response shapes, and that should at least be aligned in this one 💯 
I'll go with the status-pages list shape which is the canonical example for the cursor-based pagination endpoints (one of the more recent)

Account members could probably be updated eventually. I'm not sure how hard we want to stick to non-breaking changes to the CLI JSON response that's largely meant for agents to use, but on paper, that's breaking so we probably can't do that lightly.

TL;DR: Good catch, I'll fix this one up!

thebiglabasky: Fixed + added notes about this in AGENTS.md to help future ones to remain consistent

miliberlin:
@thebiglabasky just a small note: you always seem to reset your commits and force-push. That makes it quite hard to follow updates as reviewer. I see why you'd want to keep the commit history clean, but we squash anyways. Having separate commits for reviewing would be a bit easier to follow changes.

packages/cli/src/rest/test-sessions.ts
```diff
@@ -86,6 +88,57 @@ export type TestSessionDetail = {
...
+export type TestSessionListEntry = {
+  id: string
+  accountId: string
+  projectId?: string | null
+  name: string
+  provider: string
+  running?: string[] | null
+  passed?: string[] | null
+  failed?: string[] | null
+  cancelled?: string[]
```
miliberlin:
```suggestion
  cancelled?: string[] | null
```

miliberlin (dismissed):
Left a last small nit, otherwise looks good to me. Not sure if @sorccu has any comments before you merge?

thebiglabasky:
> @thebiglabasky just a small note: you always seem to reset your commits and force-push. That makes it quite hard to follow updates as reviewer. I see why you'd want to keep the commit history clean, but we squash anyways. Having separate commits for reviewing would be a bit easier to follow changes.

Hm weird, I ask my agent to rebase, but this seems to go sideways indeed. Thanks for the callout, I'll see if there's something fishy and I'll tilt the agent to avoid doing that 👍🏻

miliberlin (dismissed):

martzoukos (approved):

PR #1332, opened 2026-05-29
feat(cli): add --user-context to rca run [AI-343]

miliberlin:
## What

Adds a `--user-context <text>` flag to `checkly rca run`, letting callers pass extra context into RCA generation:

```
checkly rca run --error-group <id> --user-context "..."
checkly rca run --test-session-error-group <id> --user-context "..."
```

## Changes

- **`commands/rca/run.ts`** — new `--user-context` flag (and updated `usage`); passed through to both trigger calls.
- **`rest/rca.ts`** — `trigger()` and `triggerTestSessionErrorGroup()` accept an optional `userContext` and send it as the POST body `{ userContext }` (or `undefined` when omitted, preserving the existing no-body behavior).
- **`rest/__tests__/rca.spec.ts`** — new unit tests covering both endpoints, with and without user context.

The `detail|json|md` output shape is unchanged — the flag only affects the request payload. Both public endpoints already accept the optional `{ userContext }` body on the backend.

### DIFF
--- a/packages/cli/src/rest/__tests__/rca.spec.ts
+++ b/packages/cli/src/rest/__tests__/rca.spec.ts
@@ -1,0 +1,60 @@
+import { describe, expect, it, vi } from 'vitest'
+import Rca from '../rca.js'
+
+describe('Rca REST client', () => {
+  it('triggers an error group RCA without a body when no user context is given', async () => {
+    const api = {
+      post: vi.fn().mockResolvedValue({ data: { id: 'rca-id' } }),
+    }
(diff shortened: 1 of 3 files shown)


packages/cli/src/rest/__tests__/rca.spec.ts
thebiglabasky, line 1: Surprising we didn't have test for this before. Good we do now! 🥇

thebiglabasky (approved):

PR #1343, opened 2026-06-05
ci: shard checkly unit and e2e tests into 2 vitest shards

sorccu:
## What

Windows test runs (unit and e2e) had grown slow again. This splits the **checkly** unit and e2e suites into two vitest shards each, run in parallel across the OS matrix. **create-checkly** stays unsharded.

The `test` job is now matrixed by package via a `target` dimension, so one definition covers both packages (sharded checkly + unsharded create-checkly).

## Note for reviewer

The `--shard` flag is forwarded **without** a `--` separator. pnpm inserts a `--` when forwarding args after one, which makes vitest treat `--shard` as a test-name filter and silently run the full suite on every shard — so the no-`--` form is load-bearing, not stylistic.

## ⚠️ Branch protection

Check names changed (e.g. `test - ubuntu-latest` → `test - checkly - ubuntu-latest - 1/2`, `e2e - checkly - … - 1/2`). If the old names are required status checks on `main`, the required-checks list needs updating or PRs will hang. No aggregation jobs were added.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -38,7 +38,14 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, windows-latest-x64]
-    name: test - ${{ matrix.os }}
+        # checkly is sharded into two; create-checkly runs unsharded (empty
+        # shard). `id` is a filesystem-safe label since the shard string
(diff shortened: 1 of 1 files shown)


sorccu:
Superseded by #1357.

PR #1344, opened 2026-06-09
Add assets command group for result manifests

thebiglabasky:
## Summary
- Add a new `checkly assets` command group for check-result and test-session result asset manifests
- Support listing and downloading assets with shared source validation and unified REST models
- Improve terminal UX with source recap, adaptive tables, and an optional tree view for nested asset paths

## Demo
https://www.loom.com/share/b63dd3dcb8e446bc96e1637117d6037a

## Testing
- Added unit coverage for REST endpoint selection, list rendering, asset selection, overwrite behavior, and download summaries
- Updated help coverage for the new top-level topic

### DIFF
--- a/packages/cli/src/commands/assets/list.ts
+++ b/packages/cli/src/commands/assets/list.ts
@@ -9,6 +9,8 @@ import {
   formatAssetManifestTree,
 } from '../../formatters/assets.js'
 import {
+  assetManifestFiltersFromSelection,
+  assetTypeSelectionFromFlag,
   assetTypes,
   fetchAssetManifest,
   filterAssetsBySelector,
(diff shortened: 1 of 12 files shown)


packages/cli/src/commands/assets/list.ts
```diff
@@ -0,0 +1,117 @@
...
+      } else if (flags.view === 'tree') {
+        output.push(formatAssetManifestTree(assets, fmt))
+        output.push('')
+        output.push(chalk.dim('Tip: use --view table to copy exact Asset values for download.'))
+      } else {
+        output.push(formatAssetManifestEntries(assets, fmt))
+      }
+
+      if (manifest.truncated) {
+        const returned = manifest.entriesReturned ?? assets.length
```
hwartig: Should this be the number of assets returned by the API? currently it gives the length of the potentially filtered asset list.
```suggestion
        const returned = manifest.entriesReturned ?? manifest.assets.length
```

hwartig (dismissed):

packages/cli/src/commands/assets/list.ts
thebiglabasky: Nice catch! Fixed, thanks.
PTAL or stamp 🙏🏻

hwartig (approved):

PR #1346, opened 2026-06-10
test: remove flaky badssl.com e2e check

thebiglabasky:
## What changed

- Removed the `Skip SSL Check` fixture API check, which depended on `https://self-signed.badssl.com/`.
- Added a parse/synthesize-only ApiCheck fixture and unit test that assert `skipSSL: true` is synthesized correctly, without performing a network request.
- Narrowed the GitHub reporter E2E to `secret.check.ts` so reporter assertions are not coupled to unrelated fixture checks.

## Why

The E2E suite depended on `https://self-signed.badssl.com/`, which resets connections from Checkly cloud runners. That made the broad CLI smoke test and the GitHub reporter test fail for reasons unrelated to the behavior they are intended to cover. `skipSSL` synthesis is now covered locally by a fixture that never reaches the network.

## Validation

- `pnpm exec vitest --run ./src/constructs/__tests__/api-check.spec.ts`
- `pnpm exec eslint e2e/__tests__/test.spec.ts src/constructs/__tests__/api-check.spec.ts`
- Commit hook ran `pnpm run lint` on staged files

I did not run the remote E2E suite locally because this shell does not have `CHECKLY_API_KEY` / `CHECKLY_ACCOUNT_ID` configured.

### DIFF
--- a/packages/cli/e2e/__tests__/fixtures/test-project/src/services/api/api.check.ts
+++ b/packages/cli/e2e/__tests__/fixtures/test-project/src/services/api/api.check.ts
@@ -21,20 +21,3 @@ const apiCheck = new ApiCheck('homepage-api-check-1', {
     ],
   },
 })
-
-const skipSslApiCheck = new ApiCheck('ssl-api-check-1', {
-  name: 'Skip SSL Check',
-  activated: false,
-  alertChannels: [slackChannel, webhookChannel],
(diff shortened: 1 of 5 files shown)


sorccu:
Let's just remove the BadSSL check entirely. I don't think it's valuable enough to keep given the trouble it causes.

sorccu (approved):

PR #1345, opened 2026-06-10
feat(cli): show per-attempt retry detail in checks get [AI-340]

miliberlin:
## What

Adds `checkly checks get <check-id> --result <result-id> --include-attempts`, which renders the full retry sequence for a result instead of just the final outcome.

A single check run produces one `FINAL` result plus one `ATTEMPT` result for each earlier failed try, all sharing a `sequenceId`. By default only the `FINAL` result is shown; this flag surfaces the attempts.

```
$ checkly checks get a1b2c3d4-0000-4000-8000-000000000001 \
    --result 7f3e9a10-1111-4000-8000-0000000000f0 --include-attempts

Login flow · failing · eu-west-1 · 4.2s
...
ATTEMPTS
  #          Status    Location    Duration  Error                      Result ID
  1          failing   eu-west-1   3.9s      Timeout 30000ms exceeded   7f3e9a10-1111-4000-8000-0000000000a1
  2          failing   eu-west-1   4.0s      Timeout 30000ms exceeded   7f3e9a10-1111-4000-8000-0000000000b2
  3 (final)  passing   eu-west-1   4.2s      —                          7f3e9a10-1111-4000-8000-0000000000f0

  View attempt: checkly checks get a1b2c3d4-0000-4000-8000-000000000001 --result 7f3e9a10-1111-4000-8000-0000000000a1
```

## How

- The list endpoint has **no server-side `sequenceId` filter**, so attempts are fetched via `resultType=ALL` in a bounded time window anchored at the result and grouped client-side by `sequenceId` (ordered oldest-first, so index = run number).
- The extra call is skipped entirely when the result has no retries (`attempts === 0`).
- Renders attempt number, status, location, duration, and a best-effort short error summary, with the `FINAL` row marked `(final)`. Works in `detail`, `md`, and `json` (`json` emits `{ result, attempts }`).

## Backend dependencies (both merged)

- AI-223 — `resultType=ALL` now returns the full FINAL+ATTEMPT sequence (was returning empty).
- AI-219 — test-session result detail endpoint (related milestone work).

## Changes

- `rest/check-results.ts` — `resultType` accepts `ALL`; documented `from`/`to` as UNIX seconds; added `errorGroupIds`.
- `formatters/check-result-detail.ts` — `groupAttemptsBySequence`, `extractResultErrorSummary`, `formatAttemptsSection`; optional `extraSections` on `formatResultDetailWithNavigation`.
[...]

PR #1349, opened 2026-06-15
Update investigation skill asset guidance

thebiglabasky:
## Summary

- Update the Checkly investigation skill references so agents use `checkly assets list` and `checkly assets download` for check-result and test-session-result assets.
- Document the `checks get --result --include-attempts --output json` envelope so agents can inspect retry attempts programmatically.
- Refresh the public `skills/checkly/SKILL.md` investigation summary and add AI-context regression tests for these guidance paths.

## Why

The asset command group and per-attempt retry detail have landed, but the investigation skill still directed agents toward manual JSON URL extraction for test-session assets. That made the new CLI surface harder for agents to discover and use correctly.

## Dependency / release note

Builds on the already-merged asset and attempt CLI changes (#1344 and #1345). This PR is not blocked on another backend/API PR.

## Validation

- `pnpm install --frozen-lockfile`
- `pnpm run sync:skills`
- `pnpm --filter checkly exec eslint src/ai-context/__tests__/references.spec.ts src/ai-context/context.ts`
- `./packages/cli/bin/run skills investigate test-sessions`
- `./packages/cli/bin/run skills investigate checks`
- `pnpm --filter checkly test -- src/ai-context/__tests__/references.spec.ts` (script ran full package suite: 93 files, 1267 passed, 2 skipped)

### DIFF
--- a/packages/cli/src/ai-context/references/investigate-checks.md
+++ b/packages/cli/src/ai-context/references/investigate-checks.md
@@ -48,6 +48,7 @@ Flags:
 
 ```bash
 npx checkly checks get <check-id> --result <result-id>
+npx checkly checks get <check-id> --result <result-id> --output json
 ```
 
 ### View retry attempts for a result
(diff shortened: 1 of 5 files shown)


packages/cli/src/ai-context/references/investigate-checks.md
```diff
@@ -60,8 +61,56 @@ error summary:
...
+
+Result detail output summarizes available screenshots, traces, and videos when
+present. Use the dedicated asset manifest commands to inspect exact asset names
+and download files:
+
+```bash
+npx checkly assets list --check-id <check-id> --result-id <result-id>
+npx checkly assets list --check-id <check-id> --result-id <result-id> --type trace --view tree
+npx checkly assets list --check-id <check-id> --result-id <result-id> --output json
+npx checkly assets download --check-id <check-id> --result-id <result-id> --asset "<Asset>"
```
stefanjudis, line 89: @thebiglabasky 

Where do the agents know the name from?

thebiglabasky: `list` shows the entry names. We might put a word about it here though 👍🏻

packages/cli/src/ai-context/references/investigate-test-sessions.md
```diff
@@ -116,13 +116,47 @@ If RCA is unavailable because of plan or entitlement limits, run `npx checkly ac
...
+has a `checkId`, use:
+
+```bash
+npx checkly checks get <check-id> --result <test-session-result-id>
+```
 
-Do not invent asset names or assume every result has the same artifact set. Some results have screenshots only, some have traces or videos, and some have no downloadable assets.
+Do not invent asset names or assume every result has the same artifact set. Some
+results have screenshots only, some have traces or videos, and some have no
+downloadable assets.
```
stefanjudis, line 165: @thebiglabasky Maybe I'm misreading this. But as far as I can tell, checks and test sessions are more or less similar. 

Should these lines also be included in the check asset download section? I feel like, if we call it out here, we wanna do it there too?

thebiglabasky: That's true. I created this specific subskill since investigation largely was about running a check, investigate the test session and run again etc. Let me see if there's another place we should update with the same.

stefanjudis (approved):
:shipit:

PR #1354, opened 2026-06-18
Support agentic test session cancellation

thebiglabasky:
## Summary
- centralize test-session cancellation wiring for `checkly test`, `checkly trigger`, and `checkly pw-test`
- keep cancellation test-session scoped so the backend resolves cancellable result types instead of the CLI sending sequence IDs
- add coverage for local and triggered Agentic checks carrying the test session/sequence IDs needed by the shared cancel path

## Backend dependency
- Backend support landed in [checkly/monorepo#2423](https://github.com/checkly/monorepo/pull/2423), so this CLI PR is not blocked.

## Verification
- `pnpm install`
- `pnpm pack` from `packages/cli`
- `pnpm run lint -- packages/cli/src/services/test-session-cancel.ts packages/cli/src/services/__tests__/test-session-cancel.spec.ts packages/cli/src/services/__tests__/test-session-runners.spec.ts packages/cli/src/commands/test.ts packages/cli/src/commands/trigger.ts packages/cli/src/commands/pw-test.ts`
- `pnpm exec vitest run --silent=true ./src/services/__tests__/test-session-cancel.spec.ts ./src/services/__tests__/test-session-runners.spec.ts ./src/services/__tests__/abstract-check-runner.spec.ts ./src/rest/__tests__/cancel.spec.ts` from `packages/cli`
- `pnpm run prepare:dist` from `packages/cli`

### DIFF
--- a/packages/cli/src/commands/test.ts
+++ b/packages/cli/src/commands/test.ts
@@ -25,6 +25,7 @@ import { BrowserCheckBundle } from '../constructs/browser-check-bundle.js'
 import { prepareReportersTypes, prepareRunLocation, validateDetachReporterTypes } from '../helpers/test-helper.js'
 import { Runtime } from '../runtimes/index.js'
 import { Bundler } from '../services/check-parser/bundler.js'
+import { registerTestSessionCancelHandler } from '../services/test-session-cancel.js'
 
 const MAX_RETRIES = 3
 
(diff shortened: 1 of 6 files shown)


packages/cli/src/commands/test.ts
```diff
@@ -379,11 +380,7 @@ export default class Test extends AuthCommand {
       detach,
     )
 
-    runner.on(Events.CANCEL, async testSessionId => {
-      reporters.forEach(r => r.onCancel())
-      if (!testSessionId) return
-      await api.cancel.cancelTestSession({ testSessionId })
-    })
+    registerTestSessionCancelHandler(runner, reporters)
```
hwartig, line 383: Nice that reads much better now 👏

hwartig (approved):

PR #1360, opened 2026-06-24
feat: add HTTPS (RFC 9460) DNS record type

sbezludny:
## What

Adds the **HTTPS** (RFC 9460, DNS type 65 / SVCB family) record type to the DNS check construct's `DnsRecordType` union, so monitoring-as-code can query a domain's HTTPS service-binding record — e.g. verify HTTP/3 advertisement (`alpn`), IP hints (`ipv4hint`/`ipv6hint`), or ECH.

## Changes

- `constructs/dns-request.ts` — add `'HTTPS'` to `DnsRecordType`, with a doc note on the `recordType` field.
- `constructs/dns-assertion.ts` — update the `DnsAssertionBuilder` `@example` to the recommended forms:
  - `DnsAssertionBuilder.textAnswer('alpn="([^"]*)"').contains('h3')` (capture group scopes the match to the alpn value)
  - `DnsAssertionBuilder.jsonAnswer('$.Answer[0].data').contains('alpn="h3')`

## Context

Pairs with the runner + API support and the webapp record-type option in the main monorepo (runner/API and webapp PRs). The runner renders the HTTPS record's rdata into the answer (`1 . alpn="h3,h2" ipv4hint="…" …`), which these assertions target.

### DIFF
--- a/packages/cli/src/constructs/dns-request.ts
+++ b/packages/cli/src/constructs/dns-request.ts
@@ -20,9 +20,7 @@ export type DnsProtocol =
  */
 export interface DnsRequest {
   /**
-   * The DNS record type to query for. Use 'HTTPS' (RFC 9460) to monitor a
-   * domain's HTTPS service binding record — e.g. whether it advertises HTTP/3
-   * via its `alpn` parameter, plus `ipv4hint`/`ipv6hint`/`ech`.
+   * The DNS record type to query for.
    *
    * @example "A"
    * @example "AAAA"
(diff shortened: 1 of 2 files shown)


packages/cli/src/constructs/dns-request.ts
```diff
@@ -19,11 +20,14 @@ export type DnsProtocol =
  */
 export interface DnsRequest {
   /**
-   * The DNS record type to query for.
+   * The DNS record type to query for. Use 'HTTPS' (RFC 9460) to monitor a
+   * domain's HTTPS service binding record — e.g. whether it advertises HTTP/3
+   * via its `alpn` parameter, plus `ipv4hint`/`ipv6hint`/`ech`.
```
sorccu: I guess that's fine but the way it is the whole `recordType` description seems to become only about this single `HTTPS` type. I'd at least cut a new paragraph for it.

sorccu (dismissed):

sorccu (approved):

PR #1361, opened 2026-06-25
feat(cli): deprecate rules command in favor of skills

stefanjudis:
## Affected Components
* [x] CLI
* [ ] Create CLI
* [ ] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Notes for the Reviewer

Deprecates the `checkly rules` command and removes the AI-context artifact it depended on.

**Command** (`src/commands/rules.ts`) — previously read the bundled
`dist/ai-context/checkly.rules.md` and wrote it into AI IDE config folders
(`.cursor/rules`, `.windsurf/rules`, `.github/instructions`, etc.). That
capability now lives in `checkly skills`, so the command is reduced to a
deprecation notice:

```
Rules were deprecated. Use `npx checkly skills`.
```

- Keeps its `readOnly`/`idempotent` metadata and is marked
  `static state = 'deprecated'`, so oclif also surfaces a deprecation warning.
- Added a unit test for the message (the command had none before).

**AI-context pipeline** (`scripts/prepare-ai-context.ts`) — stops generating
`checkly.rules.md` and drops the now-dead heading/frontmatter helpers. Renamed
the misleading `RULES_OUTPUT_DIR` constant (it only feeds onboarding output now).

**Release workflow** (`release.yml`) — removes the step that attached
`checkly.rules.md` to GitHub Releases; the `finalize-release` job still marks
the release as latest. Updated `CONTRIBUTING.md` accordingly.

**Left in place / for reviewer's call:** the `release` job's
`Save LLM rules as an artifact` step (uploads the whole `dist/ai-context/*`
bundle as `llm-rules-release`) is now orphaned — its only consumer was the
finalize download. I kept it since it archives the full AI-context bundle, not
just rules, and its prerelease twin was already unconsumed. Happy to drop it if
we'd rather not keep dead archival.

## New Dependency Submission

None.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -171,22 +171,13 @@ jobs:
           SLACK_TITLE: ':white_check_mark: NPM publish succeeded'
           SLACK_MESSAGE: by ${{ github.actor }}
           SLACK_FOOTER: ''
-  # Attach checkly.rules.md to the GitHub Release and mark it as latest
+  # Mark the GitHub Release as latest if it's the highest version
   finalize-release:
     runs-on: ubuntu-latest
     needs: release
     permissions:
       contents: write
     steps:
-      - name: Download LLM rules artifact
-        uses: actions/download-artifact@v4
-        with:
-          name: llm-rules-release
-          path: llm-rules-release
-      - name: Upload checkly.rules.md to GitHub Release
-        env:
-          GH_TOKEN: ${{ github.token }}
-        run: gh release upload ${{ github.event.release.tag_name }} llm-rules-release/checkly.rules.md --repo ${{ github.repository }}
       - name: Mark release as latest if version is highest
         env:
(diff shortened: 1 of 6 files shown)


.github/workflows/release.yml
```diff
@@ -171,22 +171,13 @@ jobs:
...
           SLACK_FOOTER: ''
-  # Attach checkly.rules.md to the GitHub Release and mark it as latest
+  # Mark the GitHub Release as latest if it's the highest version
   finalize-release:
     runs-on: ubuntu-latest
     needs: release
     permissions:
       contents: write
     steps:
-      - name: Download LLM rules artifact
```
stefanjudis, line 181: This part is here to make the rules accessible on the docs via a marketing site redirect. I'll clean that all up.

thebiglabasky (approved):

PR #1362, opened 2026-06-26
feat(cli): gRPC/SSL/Traceroute typed result rendering + IaC constructs

danielpaulus:
## What this does
CLI support for the new uptime monitor types — **gRPC, SSL, Traceroute** — in three parts.

**1. Typed result rendering.** `checkly test` and `checks get` render per-type result output for gRPC/SSL/traceroute (typed diagnostics blocks, not generic JSON) with per-type formatters + tests.

**2. IaC constructs** — new `checkly/constructs` classes so users can author + deploy the monitors and backend export templates compile:
- **`GrpcMonitor`** — `checkType: 'GRPC'`, top-level `degradedResponseTime`/`maxResponseTime`; request mirrors the public API (`grpcConfig{…}`); `GrpcAssertionBuilder` (responseTime/statusCode/healthCheckStatus/responseMessage/responseMetadata).
- **`SslMonitor`** — `checkType: 'SSL'`; response-time limits inside `sslConfig`; models serverName/skipChainValidation/handshakeTimeoutMs/alertDaysBeforeExpiry/securityBaseline/clientCertificate; `SslAssertionBuilder` (expiry/trust/hostname/TLS version/cipher/key size/…).
- **`TracerouteMonitor`** — `checkType: 'TRACEROUTE'`; request (protocol/port/maxHops/…); `port` dropped for ICMP (matches the backend strip); `TracerouteAssertionBuilder` (responseTime/hopCount/packetLoss).
- Wired into `constructs/index.ts`, `check-codegen.ts`, and `constants.CheckTypes`. Synthesized payloads are byte-identical to the validated public-API create bodies.

**3. Deploy fix** — uptime-only projects (no Playwright files) skip the bundle/upload step (`Bundler.isEmpty`).

## Testing
- Unit: construct specs (grpc/ssl/traceroute-monitor), codegen + export-snippet specs, formatter `check-result-detail` (+ snapshots), reporter `json-builder`/`util` specs. `tsc` clean.
- **E2E**: `deploy.spec.ts` deploys a project with grpc/ssl/traceroute check fixtures.
- CI green on **Ubuntu + Windows** (CI gate + full e2e suite).

<details>
<summary>🧪 Manually tested (live CLI runs)</summary>

Exercised end-to-end against a **local devenv** and a **prod account**, for SSL / gRPC / traceroute, pass + fail:

- **`checkly test`** — all three types run and render typed result blocks; failing assertions show ✔/✖ with `Received:` (prod run: 3 pass / 3 fail exactly as configured; gRPC `HEALTH` mode carried through; traceroute produced a real 11-hop trace).
- **`checkly deploy`** — a 35-check exhaustive set (14 SSL / 12 gRPC / 9 traceroute) covering every valid property, config enum, and assertion source × comparator deploys clean.
- **`checkly checks get <id>`** — overview (per-type STATS, RECENT RESULTS, ERROR GROUPS) and `--result` detail: SSL certificate + per-rule security baseline, gRPC status/health/timing/metadata, traceroute hop table + assertions; also `--stats-range` and `--group-by location`.
- **`checkly checks list`** — table, `--output md`, and `--tag` / `--type` / `--status` filters.
- **`checkly checks stats`**, **`checks delete`**, **`trigger`** — all work for the three types on devenv and prod.
- **`--output json`** — emits the full result body (for agent consumption).

- **`checkly import`** (reverse codegen) — generates valid `SslMonitor` / `GrpcMonitor` / `TracerouteMonitor` constructs from existing checks. This surfaced a backend gap where the import plan dropped the SSL/gRPC per-type config (both threw `reading 'mode'`/`'hostname'` before) — fixed in monorepo #2864 — then verified **end-to-end on both a devenv and a prod account**: the SSL/gRPC/traceroute constructs regenerate cleanly with full `request.sslConfig`/`grpcConfig` and assertions.

</details>


sorccu:
## New-monitor checklist audit

I compared this PR against the `/checkly-cli-new-monitor` command, which documents the required functionality for a new monitor construct. The PR adds three monitors (GRPC, SSL, TRACEROUTE); findings below were independently re-verified against the diff.

### ✅ Covered
- **Phase 1 — Core constructs:** `{name}-monitor.ts` / `-assertion.ts` / `-request.ts` for all three. Constructors call `registerConstruct` + `addSubscriptions` + `addPrivateLocationCheckAssignments`; `describe()`, `validate()` and `synthesize()` (spreads `super.synthesize()` + sets `checkType`) all implemented.
- **Phase 2 — Codegen:** `-monitor-codegen.ts` / `-assertion-codegen.ts` / `-request-codegen.ts` for all three, following the `IcmpMonitorCodegen` pattern.
- **Phase 3 — Registry:** `index.ts` exports, `check-codegen.ts` (import/property/init/`describe`+`gencode` cases), `constants.ts` `CheckTypes`.
- **Phase 4 (partial):** `reporters/util.ts` per-type blocks; `batch-stats.ts` types added to `TIMING_TYPES`.
- **Phase 6 (partial):** construct unit specs for all three, plus bonus codegen coverage in `uptime-monitor-codegen.spec.ts` / `uptime-monitor-export-snippets.spec.ts`.

### ❌ Missing
- **Phase 4 — `src/rest/analytics.ts`:** no `checkTypeToPath` / `defaultMetrics` entries for the new types.
- **Phase 5 — AI context:** no `ai-context/references/configure-{name}-monitors.md`, and `ai-context/context.ts` (`REFERENCES` + `EXAMPLE_CONFIGS`) untouched. The three monitors are undocumented in AI context / public skills. (No `sync:skills` CI failure results, since the generated skill output doesn't change — but the docs gap remains.)
- **Phase 6 — E2E:** no `e2e/.../deploy-project/{name}.check.ts` fixtures and no `e2e/deploy.spec.ts` assertions.
- **Phase 7 — Examples:** no TS examples in `examples/advanced-project/.../uptime/` and no JS equivalents in `examples/advanced-project-js/`.

### ⚠️ Worth a look
- **Connection-error subsection (in code the PR already touches):** the ICMP/DNS templates in `reporters/util.ts` render a `Connection Error` subsection (`checkRunData.connectionError` → `formatConnectionError`). The new GRPC/SSL/TRACEROUTE blocks handle `requestError` + `response` + `assertions` but omit it.
- **`batch-stats.ts` per-type columns (optional):** the checklist only asks for metric columns when a type has non-response-time metrics. GRPC/SSL are fine on standard timing; **TRACEROUTE** is the candidate worth considering (packet-loss / hop-count, analogous to ICMP, which has bespoke columns).

**Net:** the runtime + codegen + result-rendering path is solid. The gaps are analytics integration, AI context/docs, e2e coverage, and user-facing examples — plus the connection-error subsection inside the new reporter blocks.

danielpaulus:
Thanks for the checklist audit @sorccu — addressed all of it. Pushed 3 commits (also rebased on `main`):

### ❌ Missing — now done
- **Phase 4 · `src/rest/analytics.ts`** — added `checkTypeToPath` (`grpc-checks` / `ssl` / `traceroute`) and `defaultMetrics` for all three, matched to the backend analytics routes + metric registries: gRPC `total_*`, SSL `handshakeTimeMs_*` + `daysUntilExpiry_avg`, TRACEROUTE `finalHopLatencyAvg_*` + `totalHops_avg`. (`feat(cli): wire … analytics paths …`)
- **Phase 5 · AI context** — added `ai-context/references/configure-{grpc,ssl,traceroute}-monitors.md` + `REFERENCES` / `EXAMPLE_CONFIGS` entries in `context.ts`. `sync:skills` now emits the three new commands. (`docs(cli): AI context references + examples …`)
- **Phase 6 · E2E** — added `e2e/.../deploy-project/{grpc,ssl,traceroute}.check.ts` (`activated: false`) and the three logical IDs in the `deploy.spec.ts` Create assertions. (`test(cli): e2e deploy fixtures …`)
- **Phase 7 · Examples** — added TS examples under `examples/advanced-project/.../uptime/` and JS equivalents under `examples/advanced-project-js/...`, with group + assertions + doc-link comments.

### ⚠️ Worth a look
- **Connection-error subsection** — added to the GRPC/SSL/TRACEROUTE reporter blocks, mirroring ICMP exactly (`checkRunData?.response?.error` → `formatConnectionError`, after the response subsection, before assertions).
- **`batch-stats.ts` TRACEROUTE columns** — intentionally **deferred** for now. It's the one item you flagged as optional, and it adds bespoke ICMP-style columns (packet-loss / hop-count). Happy to add it in this PR if you'd prefer it land together — just say the word.

### Verification
- `tsc --build` clean; `eslint --fix` clean; `sync:skills` regenerates with the three new commands; the 3 new construct specs pass (17 tests).
- Full unit run: 1312 pass; the 40 failures are confined to `project-parser.spec.ts` / `playwright-check.spec.ts` and are a local corepack/pnpm supply-chain policy block on `undici-types@6.21.0` (the `FixtureSandbox` `pnpm install` step) — unrelated to this change and not in any file this PR touches. CI on the PR is the source of truth there.

danielpaulus:
Correction on **Phase 6 (e2e)**: I initially added the three fixtures to `e2e/.../deploy-project/` but **reverted that** — that suite deploys against **production `api.checklyhq.com`**, where GRPC/SSL/TRACEROUTE aren't GA yet, so the deployed checks make the shared-account `GET /v1/checks` readback 500 and cascade into the other deploy tests. The `test` workflow went red for exactly that reason; with the e2e fixtures dropped it should go back green.

Proposal: add the deploy-project e2e fixtures once the three types are GA on production (or gate them behind a skip until then). The other checklist items (analytics, AI context, examples, connection-error) are unaffected and remain. End-to-end deploy/create/read/destroy for all three is already covered against the local devenv in the spec's P5 UAT.

danielpaulus:
Re-instated the Phase 6 e2e fixtures + `deploy.spec.ts` assertions for gRPC/SSL/Traceroute — now that the backend + check types are merged, the prod-hitting deploy e2e should exercise them for real (which is what we want). Watching CI.

packages/cli/src/constructs/ssl-monitor.ts
```diff
@@ -0,0 +1,82 @@
...
+import { SslRequest } from './ssl-request.js'
+import { InvalidPropertyValueDiagnostic, RequiredPropertyDiagnostic } from './construct-diagnostics.js'
+
+export interface SslMonitorProps extends MonitorProps {
+  /**
+   * Determines the request that the monitor is going to run.
+   *
+   * Unlike most monitors, the response-time limits for an SSL monitor live
+   * inside the request as `sslConfig.degradedResponseTimeMs` and
+   * `sslConfig.maxResponseTimeMs`.
```
sorccu: But why? Why can't we have it the same way we have it for every other monitor and check?

packages/cli/src/constructs/ssl-request.ts
```diff
@@ -0,0 +1,185 @@
...
+
+  /**
+   * The number of milliseconds to wait for the TLS handshake to complete before
+   * timing out.
+   *
+   * @minimum 1000
+   * @maximum 30000
+   * @default 10000
+   */
+  handshakeTimeoutMs?: number
```
sorccu: We don't traditionally include an `Ms` suffix for properties even if they are timeouts and even if the value is milliseconds. I would remove it.

packages/cli/src/constructs/ssl-request.ts
```diff
@@ -0,0 +1,185 @@
...
+
+  /**
+   * The handshake time in milliseconds above which the monitor is considered
+   * failing. Must be greater than or equal to `degradedResponseTimeMs`.
+   *
+   * @minimum 0
+   * @maximum 30000
+   * @default 10000
+   */
+  maxResponseTimeMs?: number
```
sorccu: These don't belong here. In every other monitor/check, they are part of the check's properties. In the check properties, you have a comment saying exactly that. However, the reason is not explained and is therefore completely unclear to me.

packages/cli/src/constructs/ssl-request.ts
```diff
@@ -0,0 +1,185 @@
...
+  /**
+   * The SSL-specific configuration for the connection.
+   */
+  sslConfig: SslConfig
+
+  /**
+   * The ID of the stored client certificate to present. Required when
+   * `sslConfig.clientCertificateMode` is `explicit`.
+   */
+  sslClientCertificateId?: string
```
sorccu: Why is this at the top level when everything else is inside `sslConfig`? If it's for model reasons - the request doesn't have to match the model exactly. You can remap things when you synthesize the bundle.

However, I feel like a few more properties that are directly related to forming the connection to the server should actually be OUTSIDE `sslConfig`. And if those are moved, then maybe this one should stay at the top level.

sorccu:
I have a bit of an issue with the way SslMonitor is structured. Here's what it would roughly look like almost populated:

```typescript
new SslMonitor('example-com-ssl', {
  name: 'example.com SSL Certificate',
  activated: true,
  frequency: 60,
  locations: ['eu-west-1'],
  request: {
    sslClientCertificateId: 'clientcert_1234', // required when mode is 'explicit'
    sslConfig: {
      hostname: 'example.com', // no scheme, no port
      port: 443,
      ipFamily: 'IPv6',
      serverName: 'example.com', // defaults to hostname
      alertDaysBeforeExpiry: 30,
      degradedResponseTimeMs: 3000,
      maxResponseTimeMs: 10000,
      skipChainValidation: false,
      // Override the account-default security baseline.
      securityBaseline: {
        enabled: true,
        minTLSVersion: { value: 'TLS1.2', severity: 'fail' },
        minKeySizeBits: { value: 2048, severity: 'fail' },
        weakCipherSuite: { severity: 'warn' },
      },
    },
    assertions: [
      SslAssertionBuilder.certExpiresInDays().greaterThan(30),
      SslAssertionBuilder.certNotExpired().equals(true),
      SslAssertionBuilder.chainTrusted().equals(true),
      SslAssertionBuilder.hostnameVerified().equals(true),
      SslAssertionBuilder.tlsVersion().equals('TLS1.3'),
      SslAssertionBuilder.keySizeBits().greaterThan(2048),
      SslAssertionBuilder.issuerCn().contains("Let's Encrypt"),
    ],
  },
})
```

In my opinion here is a closer approximation of what it should look like:
```typescript
new SslMonitor('example-com-ssl', {
  name: 'example.com SSL Certificate',
  activated: true,
  frequency: 60,
  locations: ['eu-west-1'],
  request: {
    hostname: 'example.com', // no scheme, no port
    port: 443,
    ipFamily: 'IPv6',
    sslConfig: {
      serverName: 'example.com', // defaults to hostname
      sslClientCertificateId: 'clientcert_1234', // required when mode is 'explicit'
      alertDaysBeforeExpiry: 30,
      skipChainValidation: false,
      // Override the account-default security baseline.
      securityBaseline: {
        enabled: true,
        minTLSVersion: { value: 'TLS1.2', severity: 'fail' },
        minKeySizeBits: { value: 2048, severity: 'fail' },
        weakCipherSuite: { severity: 'warn' },
      },
    },
    assertions: [
      SslAssertionBuilder.certExpiresInDays().greaterThan(30),
      SslAssertionBuilder.certNotExpired().equals(true),
      SslAssertionBuilder.chainTrusted().equals(true),
      SslAssertionBuilder.hostnameVerified().equals(true),
      SslAssertionBuilder.tlsVersion().equals('TLS1.3'),
      SslAssertionBuilder.keySizeBits().greaterThan(2048),
      SslAssertionBuilder.issuerCn().contains("Let's Encrypt"),
    ],
  },
  degradedResponseTime: 3000,
  maxResponseTime: 10000,
})
```

This moves `degradedResponseTime` and `maxResponseTime ` to the top level to match all other checks, and it moves properties that are directly related to forming a connection (request) from `sslConfig` one level higher, and `sslClientCertificateId` actually goes inside `sslConfig`. I think this should work reasonably well with our existing models as well. Is there any strong reason why we cannot have it this way?

sorccu (changes requested):
See comments

danielpaulus:
Done — restructured `SslMonitor` to your shape (commit `4c9ad1f`). No strong reason not to; it also lines the construct up more closely with the Terraform provider (which already keeps `hostname`/`port` at the request level).

- `degradedResponseTime` / `maxResponseTime` → **construct top level** (was `sslConfig.degradedResponseTimeMs`/`maxResponseTimeMs`)
- `hostname` / `port` / `ipFamily` → **`request` top level** (were in `sslConfig`)
- `sslClientCertificateId` → **into `sslConfig`** (was on `request`)

`synthesize()` remaps back to the exact same API wire shape (`request.sslConfig.*` + `request.sslClientCertificateId`), so the payload the backend receives is unchanged — the change is purely construct ergonomics. Updated codegen (wire→construct), unit + codegen + export-snippet tests, examples (ts+js), the e2e fixture, and the AI-context reference to your canonical example. `tsc` + the SSL/codegen specs are green.

Also folded in monorepo #2730: added `GrpcAssertionBuilder.textBody()` for the new `TEXT_BODY` gRPC assertion source (commit `51bf278`).

Follow-up I'll handle separately: the backend `cli/ssl-check.ejs` export template needs the matching shape so "export as code" round-trips — small monorepo PR.

danielpaulus:
Also dropped the `Ms` suffix on `handshakeTimeoutMs` → `handshakeTimeout` (commit `1bd67bb`) — matches your "no `Ms` suffix even for ms timeouts" note; `synthesize()` still emits the API's `handshakeTimeoutMs` on the wire. The backend CLI export template is updated to the new shape in checkly/monorepo#2765 so "export as code" round-trips.

danielpaulus:
On the optional **`batch-stats.ts` TRACEROUTE columns** — agreed, TRACEROUTE is the ICMP-like case (bespoke `hopCount` / `packetLoss` columns rather than the generic `responseTime` timing columns). I'd like to do it, but as a **follow-up PR** to keep this one focused on the new-monitor rollout. Tracked as **[SIM-278](https://linear.app/checklyhq/issue/SIM-278/cli-traceroute-specific-columns-in-checkly-checks-stats-hop-count)**.

Everything else from your review is addressed in this PR:
- **SslMonitor restructure** (your 4 inline notes): `degradedResponseTime`/`maxResponseTime` → construct top level; `hostname`/`port`/`ipFamily` → `request` top level; `sslClientCertificateId` → inside `sslConfig`; dropped the `Ms` suffix (`handshakeTimeout`). `synthesize()` remaps back to the identical wire shape, so the API payload is unchanged.
- **Checklist gaps:** `src/rest/analytics.ts` paths + default metrics; AI-context references + `context.ts`; e2e deploy fixtures + assertions; TS/JS examples; the connection-error subsection in the GRPC/SSL/TRACEROUTE reporter blocks.
- **gRPC `TEXT_BODY`** assertion (monorepo #2730).

CI is green now — the earlier e2e `deploy.spec` failure was unrelated to this PR: a backend bug where CLI-deployed SSL/TRACEROUTE checks defaulted to `frequency: 0` → `frequencyOffset: 0`, which the read schema rejects, 500ing `GET /v1/checks`. Fixed in checkly/monorepo#2778 (now in prod). Ready for another look when you have a moment 🙏

sorccu (approved):
I'm not sure why we need the `sslConfig`/`grpcConfig` level under `request` given that the Terraform provider PR has excluded them from the start, but alright.

PR #1387, opened 2026-07-06
fix(constructs): make SSL/traceroute assertion builders emit backend-valid payloads [SIM-287]

danielpaulus:
## SIM-287

Fixes CLI assertion builders that generated payloads the backend rejects, or couldn't express supported checks. Targets `feat/cli-pertype-result-rendering` (#1362).

- **Traceroute `responseTime()` was unusable** — it synthesized `property:''`, but the backend requires `property ∈ {avg,min,max,stdDev}` → 400. Now defaults to `avg` and adds `.avg()/.min()/.max()/.stdDev()` selectors; codegen preserves the property on round-trip (was silently collapsing to `avg`).
- **SSL baseline severity** `'warn'` → `'degrade'` to match the backend enum (`['fail','degrade','ignore']`).
- **Missing operators** — added `greaterThanOrEqual` (→ `GREATER_THAN_OR_EQUAL`, needed for `KEY_SIZE_BITS`/`TLS_VERSION`) and `matches` (→ `MATCHES`, for `CIPHER_SUITE`/`ISSUER_CN`); wired both into codegen.
- **Missing SSL sources** — added `OCSP_STAPLED`, `HANDSHAKE_TIME_MS`, `SAN_CONTAINS` (builders + codegen).
- **`degraded ≤ max` validation** — added once in the shared `validateResponseTimes`; also guards the case where `maxResponseTime` is omitted (compares against the per-type default).

Codegen (`checkly import`) was updated alongside so the new sources/operators/selectors round-trip instead of throwing or corrupting.

## Tests
New `ssl-assertion-codegen.spec.ts` and `traceroute-assertion-codegen.spec.ts` round-trip the new sources/operators and traceroute property preservation; `ssl-monitor`/`traceroute-monitor` specs cover the new severity, selectors, and the degraded>max diagnostic.

## Verified live
`TracerouteAssertionBuilder.responseTime().max().lessThan(…)` now round-trips as `max` and evaluates: *`response time property "max" is less than target … Received: 6.8`*.

## Note
The `degraded ≤ max` check now applies to all monitor types (was SSL-only inline) — a small, intentional broadening.

https://claude.ai/code/session_01NC9pESYE5wGqy5TKRzufyc

### DIFF
--- a/packages/cli/src/constructs/__tests__/ssl-assertion-codegen.spec.ts
+++ b/packages/cli/src/constructs/__tests__/ssl-assertion-codegen.spec.ts
@@ -1,0 +1,55 @@
+import { describe, it, expect } from 'vitest'
+
+import { valueForSslAssertion } from '../ssl-assertion-codegen.js'
+import { SslAssertion } from '../ssl-assertion.js'
+import { GeneratedFile, Output } from '../../sourcegen/index.js'
+
+function render (assertion: SslAssertion): string {
+  const output = new Output()
+  const file = new GeneratedFile('foo.ts')
+  const result = valueForSslAssertion(file, assertion)
+  result.render(output)
+  return output.finalize()
+}
+
+describe('SSL Assertion Codegen', () => {
+  it('generates the new SSL assertion sources', () => {
+    const cases: { input: SslAssertion, expected: string }[] = [
+      {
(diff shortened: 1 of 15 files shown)


danielpaulus:
Superseded by https://github.com/checkly/checkly-cli/pull/1391 (SIM-286 + SIM-287 combined into one PR).

PR #1390, opened 2026-07-07
feat(cli): add deploy --preserve-resources + destructive-delete guard (AI-426) [ship]

martzoukos:
## What & why
[AI-426](https://linear.app/checklyhq/issue/AI-426) (CLI half). Adds `checkly deploy --preserve-resources`: resources removed from code are **detached** (kept in the account, returned to UI management) instead of deleted. Also adds a guardrail so a default `deploy` can't silently destroy run history.

## Changes
- New `--preserve-resources` boolean flag; forwarded as `preserveResources` to the deploy endpoint (`rest/projects.ts`).
- Pre-deploy "This will:" summary advertises the flag.
- **Preflight destructive-delete guard**: for an interactive, non-forced, non-preserve run, a dry-run surfaces resources that would be permanently deleted and requires explicit confirmation (skipped by `--force`).
- Preview renders removed resources in a dedicated **"Detached (kept in account, now UI-managed)"** section, separate from the red "Delete:".
- **Explicit `DETACHED` diff action**: the preview reads a `DETACHED` action from the deploy diff and buckets those resources as detached, rather than inferring detachment purely from the client-side flag. The CLI is now the source of truth for the label based on what the backend actually did.
- Unit test for the query-param forwarding.

## Backwards compatibility of the `DETACHED` action
- **Older backends** that still return `DELETE` for preserved resources keep working: when `--preserve-resources` is set, `DELETE` items are folded into the detached bucket.
- **Older CLIs** receiving `DETACHED` from a newer backend won't break — `action` is an unvalidated `string` matched by an `if/else` chain, so an unknown value is simply skipped (no crash). It would be silently omitted from the preview, so the backend should only emit `DETACHED` when the request opted in via `preserveResources`.

## Behavioral notes
- Every interactive, non-forced deploy now makes one extra dry-run round-trip; a destructive one prompts twice (generic intent, then the deletion gate).
- Requires the backend `preserveResources` param (checkly/monorepo: https://github.com/checkly/monorepo/pull/2854). The `DETACHED` action is an additive, opt-in response value.

## Testing
`tsc` build ✅, eslint ✅, vitest ✅.


packages/cli/src/commands/deploy.ts
```diff
@@ -26,8 +26,28 @@ enum ResourceDeployStatus {
   UPDATE = 'UPDATE',
   CREATE = 'CREATE',
   DELETE = 'DELETE',
+  // Returned by newer backends for resources removed from code that are kept in
+  // the account (detached, now UI-managed) instead of deleted.
+  DETACHED = 'DETACHED',
 }
 
+const PRETTY_RESOURCE_TYPES: Record<string, string> = {
```
thebiglabasky, line 35: Isn't this missing StatusPages?

packages/cli/src/commands/deploy.ts
```diff
@@ -108,6 +133,9 @@ export default class Deploy extends AuthCommand {
           scheduleOnDeploy
             ? 'Schedule checks after deploy'
             : 'Checks will NOT be scheduled after deploy',
+          preserveResources
+            ? 'Detach any resources removed from code, keeping them and their run history for management in the UI'
```
thebiglabasky: \>`management in the UI`<br><br>`management in the Checkly Webapp`

packages/cli/src/commands/deploy.ts
```diff
@@ -108,6 +133,9 @@ export default class Deploy extends AuthCommand {
...
+            : 'Delete any resources removed from code, losing their run history — pass --preserve-resources to detach and keep them instead',
```
thebiglabasky: The em dash actually hurts readability because of the -- next to it. I'd use a "." or ";"

packages/cli/src/commands/deploy.ts
```diff
@@ -238,10 +266,55 @@ export default class Deploy extends AuthCommand {
...
+            destructive: Deploy.destructive,
+            idempotent: Deploy.idempotent,
+          },
+        }, { force })
+      }
+    }
+
     try {
-      const { data } = await api.projects.deploy({ ...projectPayload, repoInfo }, { dryRun: preview, scheduleOnDeploy })
+      const { data } = await api.projects.deploy(
```
thebiglabasky, line 323: Looks like we're unconditionally calling deploy again, so if we `!preview && !preserveResources && !force` then there's no return (unless `confirmOrAbort` effectively exits the method) so that looks like a miss?

packages/cli/src/commands/deploy.ts
```diff
@@ -238,10 +266,55 @@ export default class Deploy extends AuthCommand {
...
+        { ...projectPayload, repoInfo },
+        { dryRun: preview, scheduleOnDeploy, preserveResources },
+      )
       if (preview || output) {
-        this.log(this.formatPreview(data, project, verbose))
+        this.log(this.formatPreview(data, project, verbose, preserveResources))
       }
       if (!preview) {
         await setTimeout(500)
```
thebiglabasky, line 341: I know this is pre-existing but that looks quite fishy

packages/cli/src/commands/deploy.ts
```diff
@@ -289,9 +380,20 @@ export default class Deploy extends AuthCommand {
...
+        // Newer backends report detached resources explicitly. The construct
+        // isn't in the project since it was removed from code.
+        detaching.push({ resourceType: type, logicalId })
       }
     }
 
+    // Backwards compatibility: older backends don't emit DETACHED and instead
+    // return DELETE for resources removed from code even when preserveResources
+    // is set. When we asked to preserve, treat those deletions as detachments.
+    if (preserveResources) {
```
thebiglabasky: I'd kill "backwards compatibility" which is nonsensical here: this is the CLI, so this version can count on the new backend endpoint.

thebiglabasky (changes requested):
My bad for the commit that's reverted, I used Linear and asked a question to the agent which just went ahead and changed stuff.<br><br>This is a human review, I think there are legit questions I'd consider before approving. I'll have a bot review done separately to consider my remarks too

packages/cli/src/commands/deploy.ts
martzoukos: Yes, good catch. Status pages and Status page services.

packages/cli/src/commands/deploy.ts
martzoukos: Oh damn, yeah that hurts.

packages/cli/src/commands/deploy.ts
martzoukos: `confirmOrAbort` indeed exits earlier. The whole `(!preview && !preserveResources && !force)` block is there just for the first dry run that needs to happen so we get a list of deleted resources:
- Don't run it if the user wants to preview (the final call does the same)
- Nor if they wanna force (so no need to surface any information)
- Neither if they choose to preserve resources (no need to check first)

packages/cli/src/commands/deploy.ts
martzoukos: Ah, the agent got overjealous here and applied backwards compat both in the backend and this code.
As we'll deploy the backend first, we indeed don't need this block – I'll remove it.

thebiglabasky (dismissed):

PR #1391, opened 2026-07-07
fix(cli): SSL/gRPC/traceroute — valid assertion builders, honest reporter, rich checks get [SIM-286][SIM-287][SIM-291]

danielpaulus:
Completes SSL/gRPC/traceroute CLI support on top of #1362: assertion constructs emit backend-valid payloads, `checkly test` output is honest, and `checkly checks get` shows rich per-type detail. Builds on the per-type result rendering + IaC constructs from **#1362 (now merged to `main`)**.

## What changed

**Assertion constructs** (SIM-287)
- **Per-source assertion builders.** Each SSL / gRPC / traceroute assertion source now has its own builder exposing only the operators (and value type) the backend accepts, so an invalid combination is a compile error instead of a deploy-time 400. E.g. `keySizeBits()`/`tlsVersion()` are `equals`-only, SSL boolean sources take `equals(true|false)`, `cipherSuite()`/`issuerCn()` take `equals`/`notEquals`/`matches`, and `hopCount()`/`packetLoss()` drop `notEquals` (which the backend rejects for them). The previously-documented `keySizeBits().greaterThan(2048)` — which the backend 400s on — no longer compiles.
- **Deploy-time assertion validation.** `checkly deploy`/`validate` now flag an assertion whose source, per-source operator, or boolean target the backend would reject as a fatal diagnostic — covering hand-written object literals that bypass the typed builders — for SSL, gRPC and traceroute.
- Traceroute `responseTime(property)` takes a statistical property (`avg` default / `min` / `max` / `stdDev`); it previously emitted an empty property the backend rejects → 400.
- `matches()` (regex) is exposed for the SSL string sources (`CIPHER_SUITE`/`ISSUER_CN`/`SIGNATURE_ALGORITHM`); boolean SSL sources serialize a boolean literal.
- Add the `OCSP_STAPLED` SSL assertion source, round-tripped through codegen. SSL baseline severity `'warn'` → `'degrade'` to match the backend enum.
- Shared `degradedResponseTime ≤ maxResponseTime` cross-field validation; gRPC response-time cap raised to 180000 ms.

> **Note:** `greaterThanOrEqual` stays dropped from the CLI, so `KEY_SIZE_BITS`/`TLS_VERSION` are `equals`-only and a "minimum key size / minimum TLS version" assertion can't yet be expressed — tracked as a follow-up. (`matches` and boolean-literal handling are now supported; `HANDSHAKE_TIME_MS`/`SAN_CONTAINS` stay excluded pending a backend source rename.)

**`checkly test` reporter** (SIM-286)
- No longer renders assertions as green ✔ under a request error (nothing was evaluated).
- Humanized gRPC labels + response-time line; SSL response-time reason + security-baseline verdict.

**`checkly checks get --result`** (SIM-291)
- Adds the missing ASSERTIONS section for SSL/gRPC/traceroute via a shared renderer used by both `checkly test` and `checks get` (can't drift).
- Rich per-type detail: SSL CERTIFICATE (issuer/subject/validity/key/signature/SHA-256/SANs/serial/self-signed·CA/OCSP) + per-rule SECURITY BASELINE; gRPC TIMING (dns/connect/total); traceroute probe protocol + DNS timing.
- `--output json` unchanged (full result body).

## Testing
- Unit: formatter + reporter tests; ssl-assertion-codegen, traceroute-assertion-codegen, grpc/ssl/traceroute-monitor specs; `tsc` + lint clean. Extracting the shared assertion renderer is not purely mechanical: it also fixes two rendering edge cases that apply to all check types (including the already-released API/DNS/TCP/URL ones) — a `Received:` value now shows for present-but-falsy actuals (`0`/`false`/`''`), and long non-string values now truncate (the cap is on the stringified length). Existing reporter snapshots are unchanged only because no fixture exercises those cases, so they don't by themselves prove neutrality.
- **Live (local devenv + a prod account), all three types, pass + fail**: `deploy` + `checkly test` render typed blocks and ✔/✖ assertions with `Received:` (no green-on-error); `checks get --result` shows per-type CERTIFICATE/BASELINE/TIMING/hop-table + the new ASSERTIONS section; `checks list/stats/delete/trigger` work for the new types; `--output json` emits the full body. Prod run = 3 pass / 3 fail as configured, real 11-hop trace.

<details>
<summary>🧪 Manually tested (live CLI runs)</summary>

Exercised end-to-end against a **local devenv** and a **prod account**, for SSL / gRPC / traceroute, pass + fail:

- **`checkly test`** — typed result blocks + ✔/✖ assertions with `Received:`, no green-on-error (prod: 3 pass / 3 fail as configured; gRPC `HEALTH` carried; traceroute real 11-hop trace).
- **`checkly checks get --result`** — SSL `CERTIFICATE` + per-rule `SECURITY BASELINE`, gRPC status/health/`TIMING`, traceroute hop table, all with the new `ASSERTIONS` section; also `--stats-range` / `--group-by location`.
- **`checkly checks list`** (table, `--output md`, filters), **`checks stats`**, **`checks delete`**, **`trigger`** — all work for the three types.
- **`checkly deploy`** — an exhaustive multi-check set covering properties, enums, and assertion sources deploys clean.
- **`--output json`** — full result body for agents.
- **`checkly import`** (reverse codegen) — generates valid `SslMonitor` / `GrpcMonitor` / `TracerouteMonitor` constructs from existing checks. Surfaced a backend gap where the import plan dropped the SSL/gRPC per-type config (both threw `reading 'mode'`/`'hostname'`) — fixed on the backend — then verified end-to-end on both a devenv and a prod account.

</details>


packages/cli/src/constructs/traceroute-assertion.ts
```diff
@@ -7,15 +7,56 @@ type TracerouteAssertionSource =
...
+
+  /** Asserts against the maximum response time. */
+  max () {
+    return new NumericAssertionBuilder<TracerouteAssertionSource, TracerouteResponseTimeProperty>('RESPONSE_TIME', 'max')
+  }
+
+  /** Asserts against the standard deviation of the response time. */
+  stdDev () {
+    return new NumericAssertionBuilder<TracerouteAssertionSource, TracerouteResponseTimeProperty>('RESPONSE_TIME', 'stdDev')
+  }
```
sorccu: Why do we want these methods vs just:
```
TracerouteAssertionBuilder.responseTime('avg').lessThan(200)
TracerouteAssertionBuilder.responseTime('max').lessThan(200)
// etc
```

This is the way we have it elsewhere. The string argument can be a typed enumeration, so from a type safety POV there isn't much difference between the two (though with functions, a non-existing function would throw an error obviously).

Can we redo this or is there a strong argument to keep it like you already have it?

sorccu (changes requested):
See comment

sorccu (approved):

PR #1392, opened 2026-07-08
feat(ssl-assertion): export TlsVersion, CipherSuite, SignatureAlgorithm typed constants

danielpaulus:
## What this does

Exports typed constant objects for SSL assertions from `ssl-assertion.ts` so callers get IDE autocomplete and avoid typo-driven silent assertion failures on the finite value sets. `GeneralAssertionBuilder` gains an optional `TargetType` parameter that narrows the value-comparison methods to these sets.

| Export | Type alias | Values |
|--------|-----------|--------|
| `TlsVersion` | `TlsVersionValue` | `TLS1_0`, `TLS1_1`, `TLS1_2`, `TLS1_3` |
| `SignatureAlgorithm` | `SignatureAlgorithmValue` | Go `x509` names — `SHA256_RSA`, `ECDSA_SHA256`, … |
| `CipherSuite` | `CipherSuiteValue` | TLS 1.3 suites + common TLS 1.2 ECDHE/RSA suites |

```typescript
import { SslAssertionBuilder, TlsVersion, CipherSuite, SignatureAlgorithm } from 'checkly/constructs'

SslAssertionBuilder.tlsVersion().equals(TlsVersion.TLS1_3)
SslAssertionBuilder.cipherSuite().equals(CipherSuite.TLS_AES_256_GCM_SHA384)
SslAssertionBuilder.signatureAlgorithm().equals(SignatureAlgorithm.SHA256_RSA)

// Existing string-literal callers still compile:
SslAssertionBuilder.tlsVersion().equals('TLS1.3')
```

## Operators

Only `equals` / `notEquals` are exposed for these SSL sources. The `greaterThanOrEqual` (`GREATER_THAN_OR_EQUAL`) and `matches` (`MATCHES`) operators were intentionally **not** exposed for now — we're not yet sure we want to keep those assertions. The backend (Joi §9 + go-runner) still supports both, so they can be re-added later without a backend change.

## Testing

- `ssl-assertion.spec.ts` — typed-constant acceptance, string-literal acceptance, and `@ts-expect-error` rejection of invalid values (strict `tlsVersion` union; `signatureAlgorithm` and open-ended `cipherSuite`).
- `tsc` clean.

## Notes

- Base is `main` (the parent `feat/cli-pertype-result-rendering` branch was squash-merged as #1362).

Closes SIM-295


packages/cli/src/constructs/ssl-assertion.ts
```diff
@@ -117,9 +213,38 @@ export class SslAssertionBuilder {
...
+   */
+  static ocspStapled () {
+    return new GeneralAssertionBuilder<SslAssertionSource>('OCSP_STAPLED')
+  }
+
+  /**
+   * Creates an assertion builder for the TLS handshake time in milliseconds.
+   * @returns A numeric assertion builder for the handshake time.
+   */
+  static handshakeTimeMs () {
```
sorccu: I would drop the `Ms` suffix here. We don't have it anywhere else, including `responseTime()` which is perhaps the closest match.

```suggestion
  static handshakeTime () {
```

packages/cli/src/constructs/ssl-assertion.ts
```diff
@@ -117,9 +213,38 @@ export class SslAssertionBuilder {
...
+    return new NumericAssertionBuilder<SslAssertionSource>('HANDSHAKE_TIME_MS')
```
sorccu: I would also drop the `_MS` suffix from here (again, look at `RESPONSE_TIME`), but that would involve some backend changes too. Now would be the time to do it though.

packages/cli/src/constructs/ssl-assertion.ts
```diff
@@ -117,9 +213,38 @@ export class SslAssertionBuilder {
...
+
+  /**
+   * Creates an assertion builder for the certificate subject alternative names
+   * (SANs).
+   * @returns A general assertion builder for the SAN entries.
    */
-  static signatureAlgorithm () {
-    return new GeneralAssertionBuilder<SslAssertionSource>('SIGNATURE_ALGORITHM')
+  static sanContains () {
+    return new GeneralAssertionBuilder<SslAssertionSource>('SAN_CONTAINS')
```
sorccu: This is very unlike anything else we have. Seemingly, the subject and the operator have been collapsed into one. Why could it not stay the way it was?

sorccu (changes requested):
See comments

packages/cli/src/constructs/ssl-assertion.ts
danielpaulus: Dropped the handshake-time assertion entirely (a022bc4) — builder + source removed, so there is no unit-suffix/naming concern left.

packages/cli/src/constructs/ssl-assertion.ts
danielpaulus: Dropped the whole assertion (builder + `HANDSHAKE_TIME_MS` source) in a022bc4, so no backend rename is needed. The SSL result still reports `handshakeTimeMs` as a response field (unaffected).

packages/cli/src/constructs/ssl-assertion.ts
danielpaulus: Removed `sanContains()` (the `SAN_CONTAINS` source) from the CLI in f7f6422 — un-collapsing subject/operator would require a backend source rename, so we dropped it for now rather than take that on here. Can reintroduce it as `san().contains()` when we do the backend change.

sorccu (approved):

sorccu:
This was actually still wrong - the assertion codegen was not updated. I'll make a PR for that.

PR #1408, opened 2026-07-15
fix(cli): avoid conflicting symlink entries in user code bundles

timohuovinen:
## Affected Components

* [x] CLI
* [ ] Create CLI
* [ ] Test
* [ ] Docs
* [ ] Examples
* [ ] Other

## Summary

Prevents Playwright user code bundles from containing the same path as both a symlink and a directory with child entries.

When selected files include a symlink and files reached through that symlink, the bundler now omits the conflicting symlink entry and archives the selected child files. This produces an archive that GNU tar can extract successfully while leaving symlinks without selected descendants unchanged.

## Test Plan

Run from `packages/cli`:

* [x] `pnpm test src/services/check-parser/__tests__/bundler.spec.ts`
* [x] `pnpm exec eslint src/services/check-parser/bundler.ts src/services/check-parser/__tests__/bundler.spec.ts --no-warn-ignored`
* [x] `pnpm test`

## Notes for the Reviewer

The regression test models a pnpm workspace link selected together with files below it and verifies that the generated tar contains the child files but not the conflicting symlink entry.

### DIFF
--- a/packages/cli/src/services/check-parser/__tests__/bundler.spec.ts
+++ b/packages/cli/src/services/check-parser/__tests__/bundler.spec.ts
@@ -1,0 +1,69 @@
+import { mkdir, mkdtemp, rm, symlink, writeFile } from 'node:fs/promises'
+import { tmpdir } from 'node:os'
+import path from 'node:path'
+
+import { list } from 'tar'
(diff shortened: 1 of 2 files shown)


sorccu:
Closing in favor of #1429.

PR #1419, opened 2026-07-23
Support Telegram forum topics [ship]

thebiglabasky:
## Summary

- add optional `messageThreadId` support to `TelegramAlertChannel`
- serialize it as Telegram's `message_thread_id` request parameter
- preserve the value when exporting existing Telegram alert channels to code
- retain the existing template shape when no topic is configured

## Why

This closes the CLI round-trip gap in [ENG-22](https://linear.app/checklyhq/issue/ENG-22/telegram-group-support). Without this change, exporting a topic-enabled channel and redeploying it would silently remove its topic routing.

## Validation

- focused construct and codegen tests — 3 tests passed
- ESLint on all touched files
- `pnpm --dir packages/cli pack`
- `git diff --check`

### DIFF
--- a/packages/cli/src/constructs/__tests__/telegram-alert-channel-codegen.spec.ts
+++ b/packages/cli/src/constructs/__tests__/telegram-alert-channel-codegen.spec.ts
@@ -1,0 +1,61 @@
+import { mkdtemp, readFile, rm } from 'node:fs/promises'
+import { tmpdir } from 'node:os'
+import path from 'node:path'
+
+import { afterEach, beforeEach, describe, expect, it } from 'vitest'
+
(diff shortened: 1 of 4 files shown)


thebiglabasky:
CI note: the Telegram-focused construct/codegen tests, lint, and package build pass. The remaining CLI failure is the pre-existing global `retry-project` e2e in `e2e/__tests__/test.spec.ts:312`: it expects four “Failing Check Result” lines but receives one. It failed identically on Ubuntu and Windows shard 3/4, and the isolated Ubuntu rerun reproduced it. None of this PR's four changed files touch that test or retry execution. I have left this PR scoped to ENG-22 rather than folding an unrelated retry regression into it.

PR #1420, opened 2026-07-24
feat(cli): run test sessions via the async v1 scheduling API [RED-680] [show]

sorccu:
Linear: RED-680

## What

`checkly test` and `checkly trigger --record` now run test sessions through the new asynchronous scheduling API instead of the legacy synchronous endpoint:

- `TestSessions.run()` calls `POST /v1/test-sessions/run`. The 202 arrives as soon as the session is validated and persisted — the backend dispatches the checks in the background — and additionally carries a `schedulingId` identifying the dispatch operation.
- New REST helpers cover the scheduling status endpoints: `getSchedulingCompletion` and `pollSchedulingUntilComplete` (30s long-poll on `GET /v1/test-sessions/scheduling/{schedulingId}/completion`, 408 → retry).

## Keeping dispatch failures visible

With the synchronous endpoint, a dispatch failure (bad snippet, missing private location, queue trouble) surfaced in the HTTP response itself. Under the async contract the 202 predates dispatch, so without anything else a background dispatch failure would mean: 202, then no results, until the per-check timeout expires with a generic message. To keep the old failure visibility:

- While results stream over MQTT, a `watchScheduling` promise polls the scheduling operation and rejects with `TestSessionSchedulingFailedError` (`code: SCHEDULING_ERROR | ABANDONED`) the moment dispatch fails — including when the scheduling worker dies and the backend reaper reclaims the operation. The run exits non-zero within seconds, with the real error.
- `--detach` awaits scheduling success before emitting DETACH, so detach mode cannot exit zero for a session that was never dispatched.
- Poll infrastructure errors are best-effort: they never fail an otherwise healthy run.

## Behavior notes (accepted)

- An ERROR can now surface mid-stream (previously only at start or timeout).
- `--detach` gains a few seconds of latency (one completion poll round-trip).
- The 408 retry loop is deliberately unbackoffed, matching the deploy completion polling convention.

## Backend dependency

Requires the new v1 test-session scheduling endpoints on the Checkly backend (internal change, tracked under RED-680; in the merge queue at the time of writing). **Blocked from release — not from merge** — until that backend change is live in production: the CLI calls v1 unconditionally, with no fallback to the legacy endpoint (deliberate decision). Merging is safe; releasing before the backend is deployed would break `checkly test` for everyone on the new version.

https://claude.ai/code/session_01HFcrNm7uzzkD1p87Nf9AhP

### DIFF
--- a/packages/cli/src/rest/__tests__/test-sessions.spec.ts
+++ b/packages/cli/src/rest/__tests__/test-sessions.spec.ts
@@ -41,7 +41,7 @@ describe('TestSessions REST client', () => {
       shouldRecord: true,
     } as any)
 
-    expect(api.post).toHaveBeenCalledWith('/next/test-sessions/run', expect.objectContaining({
+    expect(api.post).toHaveBeenCalledWith('/v1/test-sessions/run', expect.objectContaining({
       repoInfo: githubRepoInfo,
     }), expect.any(Object))
   })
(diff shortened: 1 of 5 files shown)


sorccu:
The failing `retry-project > Should execute retries` e2e is not caused by this PR: it fails identically on every branch (including ones using the legacy endpoint, e.g. #1419) since 2026-07-23, because of a regression in the production results pipeline that silently disables ad-hoc retries. Tracked internally as TREE-1094. The rest of the e2e suite passes on this branch.

PR #1479, opened 2026-09-11
feat(constructs): status page component configuration and supportLink

tnolet:
## Summary

Companion to the backend PR in checkly/monorepo (`feat/status-page-component-configuration-support-link`, #4164); deploys need that backend change to accept the new fields.

### Component settings on `StatusPageV3Component`
- The backend's per-type `configuration` is surfaced as top-level props, typed by `type` through a discriminated union:
  - `type: 'SERVICE'` (or omitted): `showHistoricalData?: boolean`, backend default `true`
  - `type: 'GROUP'`: `expandedByDefault?: boolean` (default `false`) and `showHistoricalData?: boolean` (default `true`)
- `expandedByDefault` without `type: 'GROUP'` fails to compile (it is declared `never` on the SERVICE branch so the omitted-type case is caught as well). `validate()` repeats the check at runtime for JavaScript and loosely typed callers.
- `synthesize()` folds whichever settings are set into the API's `configuration`; omitted ones are left to the backend defaults, and nothing set sends no configuration.
- Import codegen generates a setting only when its value differs from the backend default, so an import of an untouched page stays as terse as before.
- The per-type defaults live in one internal table (`constructs/internal/status-page-v3-component-configuration.ts`) shared by validation, synthesis and codegen.

```ts
new StatusPageV3Component('regions', {
  statusPage,
  type: 'GROUP',
  name: 'Regions',
  displayOrder: 1,
  expandedByDefault: true,
})
```

### `StatusPageV3.supportLink`
- New optional footer link next to `privacyPolicyLink` and `termsOfServiceLink`, synthesized and generated on import.

### Docs
- AI-context reference for supporting constructs documents the settings and their defaults. The published skill needed no changes.

## Test plan
- [x] `status-page-v3.spec.ts` and `status-page-v3-codegen.spec.ts` (17 tests)
- [x] Compile-time probe: `expandedByDefault` rejected both without `type` and with `type: 'SERVICE'`; accepted with `type: 'GROUP'`
- [x] `tsc --noEmit` and ESLint clean
- [x] `pnpm --filter checkly run prepare` + `pnpm run sync:skills`: no skill diff

https://claude.ai/code/session_01XNTMz7yhDuvjdoK7ajzKFp

### DIFF
--- a/packages/cli/src/constructs/status-page-v3-component.ts
+++ b/packages/cli/src/constructs/status-page-v3-component.ts
@@ -2,23 +2,57 @@ import { Construct } from './construct.js'
 import { InvalidPropertyValueDiagnostic } from './construct-diagnostics.js'
 import { Diagnostics } from './diagnostics.js'
 import { validatePhysicalIdIsUuid } from './internal/common-diagnostics.js'
+import { defaultConfigurationByType } from './internal/status-page-v3-component-configuration.js'
 import { Ref } from './ref.js'
 import { Session } from './session.js'
 import { StatusPageV3, StatusPageV3Ref } from './status-page-v3.js'
 
 export type StatusPageV3ComponentType = 'SERVICE' | 'GROUP'
 
-export interface StatusPageV3ComponentProps {
+/**
+ * Type-specific settings of a `SERVICE` component.
+ *
+ * Mirrors the backend contract (`@checkly/shapes/status-pages-component` in
+ * the monorepo): an omitted key takes the backend default, an unknown key is
+ * rejected at deploy time.
+ */
+export interface StatusPageV3ServiceComponentConfiguration {
+  /**
+   * Show the historical status (the uptime bar) of the component on the
+   * status page. Defaults to true.
+   */
+  showHistoricalData?: boolean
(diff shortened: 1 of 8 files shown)


packages/cli/src/constructs/status-page-v3-component.ts
```diff
@@ -163,6 +208,43 @@ export class StatusPageV3Component extends Construct {
...
+  // check at runtime for JavaScript users and loosely typed objects, matching
+  // what the backend rejects.
+  private validateSettings (diagnostics: Diagnostics): void {
+    const allowedKeys = Object.keys(defaultConfigurationByType[this.componentType] ?? {})
+    for (const [key, value] of Object.entries(this.settings())) {
+      if (value === undefined) {
+        continue
+      }
+      if (!allowedKeys.includes(key)) {
+        diagnostics.add(new InvalidPropertyValueDiagnostic(
```
sorccu, line 235: You could have used `UnsupportedPropertyDiagnostic` here but it's not a strict rule.

sorccu (approved):

PR #1484, opened 2026-09-16
feat(cli): device-flow login for agents, remote shells and inline auth (prototype)

hlenke:
## Why change the login method

`checkly login` today runs an authorization-code flow that completes through a callback server on `localhost:4242`. That has two consequences:

1. **The browser must run on the same machine as the CLI.** Over SSH, in a devcontainer, in a cloud coding session or in an agent sandbox the callback never arrives, so login cannot complete at all.
2. **It cannot be driven without a terminal.** Without a TTY the command prints its first menu and exits 0 having logged nothing in (screenshot below). Every AI agent hits this, which is why the bundled skill currently tells agents that login "cannot be completed by an AI agent" and to hand off to the user.

The OAuth 2.0 Device Authorization Grant (RFC 8628) removes both constraints: the CLI shows a URL and a short code, the user approves on any device with a browser, and the CLI polls for the result. No local port, no redirect, and the output can be plain JSON for an agent to relay.

This is also what comparable CLIs have converged on, verified in their source:

- **Vercel CLI** replaced its email, GitHub, GitLab and Bitbucket login flags with a single device flow (`vercel login`, deprecations announced in their [changelog](https://vercel.com/changelog/new-vercel-cli-login-flow)). It also starts the login inline when a command has no credentials and, in agent mode, emits structured `action_required` output for the agent to hand to the user, with `VERCEL_TOKEN` for CI. This PR follows that model.
- **Render CLI** uses a device grant against its dashboard for `render login` (`/device-grant`, code plus URL, polling), with `RENDER_API_KEY` for automation.

Neither keeps a localhost-callback path for interactive use, and neither has a separate flow for humans and agents. Our equivalent of the CI token, `CHECKLY_API_KEY` + `CHECKLY_ACCOUNT_ID`, stays as it is. What we keep that they do not have is the exchange of the OAuth token for a long-lived Checkly API key, so nothing expires mid-session, and the CLI-side registration of brand-new users with a trial account, so sign-up from the CLI keeps working.

## What this does

Prototype of a login that an AI agent, a CI-less script, or a developer on a remote shell can actually complete. Built test-first in five iterations, each its own commit.

1. **OAuth device authorization flow** (`src/auth/device-flow.ts`). RFC 8628 against the Auth0 tenant: request a device code, show the user a URL and a short code, poll the token endpoint honouring `authorization_pending`, `slow_down`, `expired_token` and `access_denied`. No local callback server, so it works over SSH, in containers and from agent sandboxes. The token-to-API-key exchange (including first-time user registration) is extracted into `src/auth/api-key.ts` and shared with the existing PKCE flow.
2. **Mode-aware `checkly login`.**
   - *interactive*: prints `Visit auth.checklyhq.com/activate and enter the code ABCD-EFGH`, opens the browser, waits. The login/sign-up menu and the "open a browser?" question are gone; the hosted page offers both. Account prompt only when there are several.
   - *agent*: zero prompts. One JSON line `{"status":"action_required","reason":"login","userActionRequired":true,"verification_uri":…,"user_code":…,"expires_in":…}` for the agent to relay to the human, then a JSON `success` line with the selected account and the list of all accounts (or a JSON error, exit 1). `--account-id` selects an account. With several accounts and no flag the CLI does **not** guess: it stores the key (keys are user-scoped), prints `{"status":"action_required","reason":"select_account","userActionRequired":false,"accounts":[…],"next":[{"command":"npx checkly login --account-id <id>"}]}` and exits 1. Running that command resumes from the stored key without a second authentication.
   - *ci*: fails fast and points at `CHECKLY_API_KEY` / `CHECKLY_ACCOUNT_ID`.
3. **Inline login.** Any authenticated command run without credentials starts the login flow in place (interactive and agent modes) instead of erroring with "run `npx checkly login`". CI keeps the error.
4. **Fallback while Auth0 is not switched on.** The device_code grant is not yet enabled for the CLI client in Auth0 (`unauthorized_client` today). When that happens the command falls back to the current localhost-callback flow, so this branch is safe to merge before the Auth0 change: interactive users see exactly today's prompts, and agents at least get the `action_required` JSON with the authorize URL instead of a silent exit 0.
5. **Test hooks.** `CHECKLY_AUTH_URL` overrides the Auth0 host in the `local` environment (same pattern as `CHECKLY_API_URL`). `--no-browser` / `CHECKLY_NO_BROWSER=1` skip launching a browser, also useful on headless hosts.
6. **The agent skill.** The bundled skill said login "cannot be completed by an AI agent" and told agents to hand off. It now has a "Logging in as an agent" section: run `login` yourself, relay the URL and code from the `action_required` line, wait, handle `select_account` with `--account-id`, set `CHECKLY_NO_BROWSER=1` on headless hosts, and rely on authenticated commands starting the login inline. The initialize reference's login step is rewritten the same way, and `skills/` is re-synced for the `check-ai-context` job. The wording holds for both the device flow and today's fallback, since both print the same JSON shape.

## Where this came from

Feedback from a user without a repository: "let an agent click through my site and create a check". The loop exists, but the very first step is broken for agents: `checkly login` with no TTY prints a menu and exits 0 without logging in, and the localhost callback cannot complete from remote or sandboxed sessions. Vercel and Render both moved to device flow; Vercel additionally starts login inline and emits structured `action_required` output in agent mode. This PR brings those three behaviours to Checkly while keeping our long-lived API key model, which is simpler for agents than refresh tokens.

## Before and after

Captured from a pseudo-terminal and rendered; the "before" runs use the released checkly 9.5.0, the "after" runs this branch with the Auth0 grant enabled.

**Before, interactive:** menu, browser question, then a wait for the localhost:4242 callback.

before interactive

**Before, without a terminal (agent):** the menu is printed, nothing is logged in, exit 0.

before agent

**After, interactive:** URL and code, browser opens with the code prefilled, no prompts.

after interactive

**After, agent mode:** `action_required` with the code, approval in the browser, `select_account` with all accounts, then `login --account-id` completes.

after agent

Images live on the `assets/pr-1484-login-screenshots` branch, to be deleted after merge.

## Verified against the real tenant

The Device Code grant was enabled on the Auth0 `cli` application for about ten minutes on 2026-09-16 and rolled back again (grant list restored to `authorization_code, implicit, refresh_token`). During that window `npx checkly login` from this branch completed the whole flow against production Auth0 in agent mode: `action_required` JSON with the activation URL and code, approval in the browser, API-key exchange, account selection, JSON success, empty stderr. With the grant removed the same command falls back to the current localhost flow automatically, also verified.

Auth0 findings from the read-only inspection: the application is already `native`, OIDC conformant and a public client, so the switch is purely additive. Neither the Universal Login page template nor any Action reads the CLI's custom `allowSignUp` / `allowLogin` / `mode` parameters, so sign-up stays available on the activation page.

## What still needs a decision

- **Enable the grant for good** once this PR is agreed: `auth0 apps update <cli client id> --grants code,implicit,refresh-token,device-code`. Rollback is the same command without `device-code`; the CLI falls back on its own, no release needed.
- **Update `e2e/__tests__/login.spec.ts`** at that point: its "should show URL to login/signup" tests assert the fallback output and start failing as soon as the grant is live. The Windows shard 3 failure on this PR was exactly that, it ran during the verification window.
- Whether the removed login/sign-up menu is acceptable in interactive mode.
- Product review of the agent JSON shape (`action_required` with `reason` = `login` | `select_account`, then `success`).

## Verified end to end as an agent

With a local build combining this PR with #1483 and #1485, and the Auth0 grant enabled for the login step only: `init` in an empty folder (skill installed under the real Claude Code environment), `login` → `action_required` → human approval → `select_account` listing 7 accounts → `login --account-id` → `whoami` → a URL monitor written and deployed through the confirmation envelope → visible in `checks list` → `destroy` → gone. That run is what produced the `confirmCommand` finding fixed in #1483.

## Tests

- Unit: 48 new tests across `device-flow`, `api-key`, `login`, `authCommand` inline login and `config.getAuthUrl`. Full CLI suite green locally (3007 tests).
- e2e, real Auth0 (fallback path today): existing login scenarios unchanged, plus agent-mode `action_required` output.
- e2e, fake Auth0 + Checkly API in-process server with an isolated home (POSIX and Windows variables): full device flow in a real CLI process for agent mode, interactive mode, and an authenticated command (`whoami`) logging in inline. This is the automated coverage of the success path while the grant is off.

## Notes for the reviewer

- The existing `e2e/__tests__/login.spec.ts` runs `checkly logout` without isolating `HOME`, so running it locally clears the developer's real CLI login (it did for me). The new spec isolates `HOME`; the harness itself is unchanged in this PR.
- `AuthMode` gained `'any'` for the agent fallback so the authorize URL allows both login and sign-up.
- Interactive mode with a closed stdin still shows the fallback menu and exits silently; treating a non-TTY stdin as non-interactive is a separate decision in `detectCliMode`.

## Affected Components
* [x] CLI


hlenke:
Live verification note: the Device Code grant was enabled on the Auth0 `cli` application for ~10 minutes today and rolled back again (grant list restored to `authorization_code, implicit, refresh_token`).

While it was on, `npx checkly login` from this branch completed the full device flow against the real tenant in agent mode: `action_required` JSON with the activation URL and code, approval in the browser, API key exchange, account selection, JSON success. With the grant removed, the same command falls back to the current localhost flow automatically.

Consequence for the existing `e2e/__tests__/login.spec.ts`: its "should show URL to login/signup" tests assert the fallback output, so they start failing as soon as the grant is enabled permanently (that is what the Windows shard 3 failure on this PR was, it ran during the window). They should be updated to expect the device-flow output when the grant is switched on for good.

